PR #2, opened 2021-08-28
Describe dispute game k-section mechanism

andrewgordstewart:
This describes an alternative mechanism to Arbitrum's bisection mechanism, which would have the sequencer commit to the root of a binary Merkle tree encoding the entire execution trace of some computation.

Instead, the sequencer simply commits to equally spaced machine steps spanning a shrinking interval.

This seems to be a simpler and more economical way to quickly zoom into the first contested machine state transition, demonstrated by the prototype Typescript implementation.

---
Note: In the absence of an easier way to share prototype Typescript code, I'm sharing this via a PR. Perhaps python is a better language for this kind of code, but I didn't want to go through the effort of getting a python dev environment running. 

Thoughts?

### DIFF
--- a/components/dispute-game-contracts/README.md
+++ b/components/dispute-game-contracts/README.md
@@ -1,0 +1,276 @@
+# Bisection game design
+
+# Context
+
+We assume that the sequencer committed to some transition in the roll-up's state $T = (R, transactions) \to R'$, where R' is the state root (aka `chainState.root`) after applying `transactions` to R, in order.
+
+Suppose a challenger wants to challenge this state. Consider a `MachineState` type something like the following:
+
+```tsx
+type MachineState {
+  root: Bytes32; // hash of (chainState.root, memory.root, instructions)
+  chainState: ChainState;
+  memory: Memory;
+  instructions: Instruction;
+  // etc
+}
+```
+
+(MachineState is assumed to be a Merkle tree, thus the `root` property.)
+
+We force the sequencer to post `S_0.root`, where `S_0 = {chainState: R, memory: emptyMemory, instructions: dissassemble(transactions)}` . The challenger contract must validate that `R` is equal to `R` posted for normal operation rollup. In addition, the challenger contract must validate that `dissassemble(transaction)` is correct. These validations are currently out of scope. We also force the sequencer to post `S'.root`, where `S' = {chainState: R', memory: emptyMemory, instructions: emptyInstructions}` . We might force the sequencer to post additional intermediate states.
+
+We assume there is a function `progress: (state: MachineState) => MachineState` which takes in a machine state `s` and returns the next state. It works by popping the latest instruction, and modifying itself according to the instruction to produce the next state. When `instructions` is empty, `progress(state)` should indicate that it's reached a final state.
+
+The following description is restricted in scope to _correctness_ requirements, and ignores _liveness_ requirements, **assuming that the sequencer & challenger each act in a timely manner.** The specific incentives & timeouts are not considered in this document.
+
+# High level overview
+
+The `progress` function _defines_ what it means for a transition $(R, transactions) \to R'$ to be valid, in the following sense:
+
+1. From $(R, transactions)$, one can construct the "initial" machine state $M_0$.
+2. Define $M_{i+1} = progress(M_i)$
+3. Compute $M_1, M_2, ...$ until you arrive at a "terminal" state $M_t$
+4. Check that $M_t.chainState.root == R'$. If yes, $(R, transactions) \to R'$ is valid; else, it is invalid.
+
+In the event of a validator disputing a sequencer-posted state root, the "bisection game" plays out as follows.
+
+1. [Off-chain] The sequencer computes _its_ belief about what the correct sequence of machine states is, generating a sequence $S_0, ..., S_n$. (A correctly behaving sequencer would find $M_i = S_i$.)
+2. [Off-chain] The validator computes _their_ belief about what the correct sequence of machine states is, generating a sequence $V_0, ..., V_m$. (A correctly behaving validator would find $M_i = V_i$.)
+3. [On-chain] The sequencer commits to the root of the machine state at some steps $step_0, step_1, ..., step_k$, where initially, $step_k = \lfloor k/n \rfloor$.
+4. [On-chain] The validator compares $S_{step_i}.root$ to $V_{step_i}.root$ for each $i$, finding the first $i$ such that they differ. (This implies $S_{step_{i-1}} \to S_{step_i}$ is an invalid transition). They assert on-chain that $step_i$ is invalid, forcing the sequencer to split the transition from $step_{i-1}$ to $step_i$ into k smaller transitions.
+5. Repeat 4, narrowing down the values of $step_i$ until the validator can directly prove an invalid transition on-chain.
+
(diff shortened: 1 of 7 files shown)


components/dispute-game-contracts/README.md
```diff
@@ -0,0 +1,276 @@
...
+1. From $(R, transactions)$, one can construct the "initial" machine state $M_0$.
+2. Define $M_{i+1} = progress(M_i)$
+3. Compute $M_1, M_2, ...$ until you arrive at a "terminal" state $M_t$
+4. Check that $M_t.chainState.root == R'$. If yes, $(R, transactions) \to R'$ is valid; else, it is invalid.
+
+In the event of a validator disputing a sequencer-posted state root, the "bisection game" plays out as follows.
+
+1. [Off-chain] The sequencer computes _its_ belief about what the correct sequence of machine states is, generating a sequence $S_0, ..., S_n$. (A correctly behaving sequencer would find $M_i = S_i$.)
+2. [Off-chain] The validator computes _their_ belief about what the correct sequence of machine states is, generating a sequence $V_0, ..., V_m$. (A correctly behaving validator would find $M_i = V_i$.)
+3. [On-chain] The sequencer commits to the root of the machine state at some steps $step_0, step_1, ..., step_k$, where initially, $step_k = \lfloor k/n \rfloor$.
```
andrewgordstewart, line 40: Sadly, github doesn't render latex. I thought it might, since VSCode seems to without any extra configuration:
Screen Shot 2021-08-27 at 5 04 51 PM

components/dispute-game-contracts/README.md
```diff
@@ -0,0 +1,276 @@
...
+
+Assume that a `Challenge.sol` contract is used to adjudicate a single challenge. There may be a `ChallengeFactory.sol` contract which deploys new `Challenge` contracts — this is how Arbitrum does it, for instance.
+
+The challenger deploys a challenge contract instance to manage the challenge for the transition T. A commitment to R, R' and `transactions` should be recorded.
+
+This logic is not currently specified.
+
+## First commitment
+
+The sequencer constructs some arbitrary sequence $S_0, S_1, ..., S_n$ and commits to the $k+1$ steps $[(S_0.root, 0), (S_{\lfloor n/k\rfloor}.root, \lfloor n/k \rfloor),..., (S_n.root, n)]$ on-chain:
```
karlfloersch, line 58: And to be clear, this is the **no upfront execution trace approach** - right?

components/dispute-game-contracts/README.md
```diff
@@ -0,0 +1,276 @@
...
+# High level overview
+
+The `progress` function _defines_ what it means for a transition $(R, transactions) \to R'$ to be valid, in the following sense:
+
+1. From $(R, transactions)$, one can construct the "initial" machine state $M_0$.
+2. Define $M_{i+1} = progress(M_i)$
+3. Compute $M_1, M_2, ...$ until you arrive at a "terminal" state $M_t$
+4. Check that $M_t.chainState.root == R'$. If yes, $(R, transactions) \to R'$ is valid; else, it is invalid.
+
+In the event of a validator disputing a sequencer-posted state root, the "bisection game" plays out as follows.
```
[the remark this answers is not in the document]
andrewgordstewart: While it's less elegant, I'm not sure there's a good way to avoid 2-phase dissection.

In principle, the system could work like this:
- your initial machine state could start with an arbitrary list of transactions (they could be organized by block or not, this doesn't matter)
- your fraud proof generator would process each transaction in order, stopping when all transactions are processed
- this produces an array `[S_0, ..., S_n]` of machine states. 

This is elegant, as the dispute game can simply binary search for the first incorrect state in that list.

BUT, I would assume that producing a fraud proof for a single transaction has a non-trivial cost. The above scheme needs to know the total number of steps `n` to be safe, meaning you have to compute a fraud proof for every transaction.

With a 2-phase dispute game, where you:
1. narrow down to the first incorrectly processed transaction (if it exists)
2. construct a fraud proof for that transaction
3. narrow down to the first incorrect step in that fraud proof (if it exists)

then you only have to construct the fraud proof for a single transaction.

---

All that said, you can use the same rules for the two phases. The code here is simply solving the following problem:
- Alice has an array `a = [A_0, A_1, ..., A_n]`
- Bob has an array `b = [B_0, B_1, ..., B_n]`
- Bob forces Alice to reveal `a[i]` for certain values of `i`, comparing them to `b[i]`, allowing him to learn the first `i` so that `a[i] != b[i]`

In a 2-phase dispute game, the interactive part of both phases can be reduced to this problem.

components/dispute-game-contracts/README.md
andrewgordstewart: that's right!

---

I thought a little more about it, and it's worth pointing out that you can use the same approach with an **upfront commitment to the whole execution trace**:
- The sequencer computes the hashes of the machine states `S_0, ..., S_n`
- They place the hashes as leaves in a balanced, k-ary Merkle tree
- They commit to the first level of the tree (they could commit to the root, but this saves a step)

Then, the verifier points to one of the currently committed to nodes. The rules dictate that the sequencer must then reveal the children of that node. The verifier can find which node to point to by constructing the correct Merkle tree, and comparing the values posted vs the local values.

(If you want to save gas, you can avoid verifying Merkle proofs, since the verifier can verify the Merkle proofs locally. you just need to allow the verifier to point out "your Merkle tree is invalid" to win the game.)

So, whether the game is searching for a leaf in a Merkle tree, or searching for an element of an array, you can implement the game as:
- sequencer posts an array of hashes
- verifier points to one of those hashes

The interpretation of the hashes differs, and the exit logic differs accordingly.

---

In the "explore the Merkle tree of the execution trace" game, it's easier to see how to halve the number of rounds:
- on their turn, each person finds the leftmost node that they contest, and they expand it according to _their_ copy of the Merkle tree
- if Alice reveals the final row of leaves, then Bob proves a proof of an invalid transition to a state corresponding to one of those leaves. (A correct actor can guarantee that if _they_ reveal a row of leaves, that every machine state corresponding to those leaves is valid.)

andrewgordstewart: > I thought a little more about it, and it's worth pointing out that you can use the same approach with an **upfront commitment to the whole execution trace**:
> 
> * The sequencer computes the hashes of the machine states `S_0, ..., S_n`
> * They place the hashes as leaves in a balanced, k-ary Merkle tree
> * They commit to the first level of the tree (they could commit to the root, but this saves a step)
> 
> Then, the verifier points to one of the currently committed to nodes. The rules dictate that the sequencer must then reveal the children of that node. The verifier can find which node to point to by constructing the correct Merkle tree, and comparing the values posted vs the local values.
> 
> (If you want to save gas, you can avoid verifying Merkle proofs, since the verifier can verify the Merkle proofs locally. you just need to allow the verifier to point out "your Merkle tree is invalid" to win the game.)


FYI @kerzhner is going to code up this idea to see how it looks.

I just discussed it with him; it seems like it's not too bad, but you have to worry about more edge cases. Plus, you are forced to supply two Merkle proofs, costing `~k log_k(n)` hashes to provide witnesses for a k-ary Merkle tree commitment to `n` machine steps. You could use a binary Merkle tree, skipping k levels at a time, at the cost of additional complexity and edge cases.

As a comparison with the simpler design outlined in this PR:
- ❌ the implementation is more difficult complex
- ❌ single step fraud proofs are more expensive
- ✅ it supports being able to refer to _any_ state prior to the first disputed state, rather than just the step immediately prior to the first disputed state
- ✅ no Merkle proofs are needed in the interactive phase
- 🏳️ (neutral) Since both parties need to compute every machine state, they've already computed all of the hashes of the machine state. The "commit to a Merkle root" approach requires both parties to compute ~twice as many hashes to fill in the tree, but that's negligible compared to creating the execution trace.

PR #3, opened 2021-08-29
Expand overview, add diagram

ben-chain:
**Description**
Adds an architecture diagram, suggests some naming and description improvements, and adds stub for batch submitter.  Thanks to @karlfloersch for much of this.

**Additional context**
The SVG metadata should contain full [draw.io](draw.io) XML source so you can import and edit that file directly.

PR #4, opened 2021-08-31
Add design goals for L1 contracts

ben-chain:
**Description**
Based on a conversation with @andrewgordstewart today, adds a set of design goals for the L1 contracts.

**Additional context**
We identified these goals as explicit decisions to be made which would unblock the majority of work on the bisection game.

Fixes ENG-1309

### DIFF
--- a/layer-1-contracts/README.md
+++ b/layer-1-contracts/README.md
@@ -1,0 +1,19 @@
+# Layer 1 Contracts
+
+## Design Goals
+We want our ORU contracts to be built in compliance with the following goals:
+- A **pluggable state machine** which:
+    - allows separation of concerns and parallelizable work
+    - is fully stateless, i.e. all state transitions can be expressed in the form `bytes32 _inputHash, bytes _witness => bytes32 _outputhash` and are only dependent on the immediately preceding pre-state
+- A cryptoeconomic **State Oracle** which:
+    - Provides a cryptoeconomic mechanism to propose, assert, challenge, and finalize optimistic states of the L2 machine
+        - Does NOT require that the entire execution trace is provable, just the execution output
+            - i.e. an honest party may not necessarily ever post a hash of the full trace
+        - DOES require that there is a winning strategy for all, and only for all, possible state proposals
+    - Provides "useful" state hash proposals, i.e. block roots or state roots, which may not be equal to the `_inputHash`es above
+    - Is optimized to minimize the number of rounds of interaction, so as to be most resistant to hostile L1 censorship and griefing
+- A set of **data feed** contracts which:
+    - allow parties and contracts to provide inputs which are processed by the L2 machine
+    - allow verifiability of those inputs for disputes
+    - is highly gas-efficient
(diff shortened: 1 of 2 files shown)


layer-1-contracts/README.md
```diff
@@ -0,0 +1,19 @@
...
+
+## Design Goals
+We want our ORU contracts to be built in compliance with the following goals:
+- A **pluggable state machine** which:
+    - allows separation of concerns and parallelizable work
+    - is fully stateless, i.e. all state transitions can be expressed in the form `bytes32 _inputHash, bytes _witness => bytes32 _outputhash` and are only dependent on the immediately preceding pre-state
+- A cryptoeconomic **State Oracle** which:
+    - Provides a cryptoeconomic mechanism to propose, assert, challenge, and finalize optimistic states of the L2 machine
+        - Does NOT require that the entire execution trace is provable, just the execution output
+            - i.e. an honest party may not necessarily ever post a hash of the full trace
```
ben-chain, line 11: This callout is one of the assumptions allowing #3 [here](https://statechannels.notion.site/Various-bisection-games-8d5ed428f7114b52a1c6bbeb08fe96d5)

layer-1-contracts/README.md
```diff
@@ -0,0 +1,19 @@
+# Layer 1 Contracts
+
+## Design Goals
+We want our ORU contracts to be built in compliance with the following goals:
+- A **pluggable state machine** which:
+    - allows separation of concerns and parallelizable work
+    - is fully stateless, i.e. all state transitions can be expressed in the form `bytes32 _inputHash, bytes _witness => bytes32 _outputhash` and are only dependent on the immediately preceding pre-state
```
ben-chain, line 7: cc @protolambda , this is an explicit callout that steps are only dependent on the immediately preceding one

layer-1-contracts/README.md
andrewgordstewart: FYI `#3` links to pull request #3 in this repo, which is misleading. I think you mean "Option 3" in the linked notion doc, right?

layer-1-contracts/README.md
```diff
@@ -0,0 +1,19 @@
...
+- A cryptoeconomic **State Oracle** which:
+    - Provides a cryptoeconomic mechanism to propose, assert, challenge, and finalize optimistic states of the L2 machine
+        - Does NOT require that the entire execution trace is provable, just the execution output
+            - i.e. an honest party may not necessarily ever post a hash of the full trace
+        - DOES require that there is a winning strategy for all, and only for all, possible state proposals
```
andrewgordstewart, line 12: Should this say something like this?
```suggestion
        - DOES require that there is a winning strategy for correct state proposals
```

I'm not sure what guarantees you can make in case BOTH parties are making incorrect state proposals. But I don't know if you care about that case?

ben-chain: Used your suggested wording, and added a bullet below it with the (probably obvious but hey) thing I was trying to convey with "only for all"

PR #7, opened 2021-09-06
Initial rollup node spec

lightclient:
Adds an initial spec for the rollup node.

### DIFF
--- a/components/rollup_node.md
+++ b/components/rollup_node.md
@@ -1,1 +1,81 @@
 # Rollup Node
+
+The consensus module of Optimisc Ethereum.
+
+## Summary
+
+The Rollup Node is a consenus client that determines the latest state of the
+rollup. It reads state from the canon chain (L1) to compute the rollup state.
+
+## Components
+
+#### Feed Oracle
+
+The Feed Oracle is an adapter that monitors the canon chain for new events in
+the rollup contract. The rollup contract provides the *canonical* order for the
+rollup chain.
+
+The Feed Oracle can index the batch submissions and deposits locally into the
+database. The batch submissions should be converted into equivalent rollup
+blocks.
+
+#### Consensus
+
+Additional checks should be imposed on the execution engine. Roughly:
+
+```python
+def on_block(db, block):
+    ctx = None
+    for tx in block.transactions:
+        # verify context tx updates canon light client with correct data
+        if tx.is_ctx():
+            assert db.is_valid_ctx(tx)
+            ctx = tx
+
+        # verify deposits are valid in the canon chain and in the context of the ctx tx
+        if tx.is_deposit():
+            assert db.is_valid_deposit(ctx, tx)
+
+    # run the block in the executin engine, if it's valid it's the new head of the
+    # rollup chain
+    assert verify_block_in_execution_engine(block)
+```
+
+#### Execution Engine
+
+The execution engine implements the [execution specification][execution-spec].
+There already many teams who plan to transition their legacy clients into
+execution engines. The rollup client will communicate to the engine via a
+bidirectional JSON-RPC interface ([WIP][execution-engine-rpc]).
+
+One of the main goal of the rollup client is to use the exeuction engine
+without modification.
+
+### Miner
+
+The miner assembles the block via the Execution Engine. It injects context
+transactions that update the light client to the canon chain and it inject new
+depsit transactions. It submits the transaction batch to and waits until the
+transaction is mined (potentially bumping fee price if needed). 
+
+#### Database
+
+The database persists the rollup chain and indexed data from the canon chain.
+Pruning of canon chain data can be aggressive since indexed data is only relavent
+to consensus during the force-inclusion period of deposits.
+
+#### Networking
+
+The Rollup Node should provide proxied access to the Execution Engine's
+networking stack. This allows for two main things:
+
+* Context updates and deposits are injected into the execution engine via
+  transactions. This means that it's possible for the Execution Engine to
+  recieve signed system updates in the public mempool. This could cause block
+  producing rollup nodes to submit bundles with invalid contexts/deposits.
+* To provide fast-confirmations, a group of rollup block producers can define a
+  "unconfirmed" blocks that haven't yet been submitted to mainnet (but will
+  likely be in the near future). 
+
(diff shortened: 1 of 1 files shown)


components/rollup_node.md
```diff
@@ -0,0 +1,81 @@
+# Rollup Node
```
karlfloersch, line 1: Is `Rollup Node` more canonical than `client`? Both terms seem good to me, I just don't have the context for the evolution of eth2 terminology so I'm curious curious

components/rollup_node.md
```diff
@@ -0,0 +1,81 @@
...
+The database persists the rollup chain and indexed data from the canon chain.
+Pruning of canon chain data can be aggressive since indexed data is only relavent
+to consensus during the force-inclusion period of deposits.
+
+#### Networking
+
+The Rollup Node should provide proxied access to the Execution Engine's
+networking stack. This allows for two main things:
+
+* Context updates and deposits are injected into the execution engine via
```
protolambda, line 72: If we end up retrofitting the deposits into transactions to use the merge API with 0 modification, then there must be a better way than wrapping the whole network stack of the execution engine. One other way could be to deny-list blocks or transactions that contain the key, unless the tx/block is allow-listed by the rollup node.

After the merge, the networking entrypoints for consensus data are split as:
- No block sync enabled, only block-inserts via rollup/beacon node through engine api (insert block).
- State sync only to trusted point, as informed by rollup/beacon node through engine api (finalized info)
- Transaction pool with p2p propagation stays

The first two are easy to check within the rollup node: parse block, check unauthorized deposit transactions (check the sender addresses), and use the "finalized" (= undisputed for a week) block hash as target for the sync (after which block-insertions complete sync to the head).

The transaction pool seems to be the only concern when doing this hack to get deposits into the unmodified engine. If a single address could be blacklisted from the tx pool via configuration, except for insertion by the API, then we are safe from spam & the rollup node can do the rest.

components/rollup_node.md
```diff
@@ -0,0 +1,81 @@
...
+The consensus module of Optimisc Ethereum.
+
+## Summary
+
+The Rollup Node is a consenus client that determines the latest state of the
+rollup. It reads state from the canon chain (L1) to compute the rollup state.
+
+## Components
+
+#### Feed Oracle
```
karlfloersch, line 12: I like `feed indexer` myself to describe this but strong opinion very weakly held.

components/rollup_node.md
protolambda: A less hacky way would be if the Merge API just exposed something in the block-insertion to ignore singatures of the block transaction(s), trusting the inserter to validate it in their way (deposits in this case, signature can just be zeroed)

karlfloersch (commented):
Oooo this is exciting!

components/rollup_node.md
lightclient: I tend to use `node` and `client`mostly  interchangeably. Maybe the distinction to me would be a node is a "thing" while a client is the actual "software"?

components/rollup_node.md
lightclient: I generally agree that there are a few ways to do this that don't require fully wrapping the execution engines networking stack. I think a denylist could be a good solution. My main concern is the continued need to add little tweaks to the execution engine's networking to get what we want. Today it's just a denylist, but I wonder if there will be additional needs in the future that can't be obtained via little tweaks.

I think I will update this remove the concept of wrapping the execution engines p2p stack.

> A less hacky way would be if the Merge API just exposed something in the block-insertion to ignore singatures of the block transaction(s), trusting the inserter to validate it in their way (deposits in this case, signature can just be zeroed)

Not sure I follow this ?

components/rollup_node.md
lightclient: Yeah, I don't love `Feed Oracle`. I think `Feed Indexer` is a more appropriate name.

components/rollup_node.md
protolambda: Without system-signed transactions, another option is to expose some trusted root in the EVM, maybe with the new `BEACON_BLOCK_ROOT` opcode. Then we can do deposits like L1->beacon deposits work (beacon chain is aware of eth1 deposit tree root), and how beacon->L1 withdrawals work (eth1 uses new opcode to retrieve previous beacon block hash).

Example setup, without any execution engine changes, aside from the standard opcode (if it's introduced for withdrawals):
- Deploy a modified DepositContract.sol from eth2 on L1. You make it only usable by other Optimism contracts. And instead of deposit messages, you can just use it to append arbitrary byte strings (with some mixin-hash or something to indicate what the message is intended for)
- Rollup node watches this L1 deposit contract, and exposes the hash via the `BEACON_BLOCK_ROOT` opcode (maybe hashing in the previous rollup block hash too, to be more similar to the real beacon block commitment properties)
- A premined contract on the rollup (effectively infinite amount of eth, and contract code in rollup genesis) can expose a method, for anyone to call with a binary merkle proof, to process a message (just like we do in the beaconchain). Since it may be the first deposit of the user, some other account can run it, to pay for gas, and get a small processing fee refunded from the deposit

See https://github.com/zilm13/mergenet-tutorial/tree/feature/withdrawal for eth2 withdrawals research work.

protolambda: Update, two possible approaches:
1. Force rollup block proposers to sign & include transaction(s) with their own (secret) key to process L1 messages. Can be part of the fraud proof of a block as a whole, and otherwise no changes in execution engine. As opposed to everyone forming and processing these transactions locally (which has the key signing problem)
2. Force rollup block proposers to include a L1 commitment in the randomness-field (`DIFFICULTY`/`RANDOMNESS` opcode is already available, as opposed to `BEACON_BLOCK_ROOT`) of the execution payload (also part of the eth1 block hash). This is exposed in the EVM, which means any lazy non-sequential L1 message processing could be implemented

lightclient: Hmm won't we want `BEACON_BLOCK_ROOT` to be usable in OE -- in the spirit of 1:1 compatibility?

karlfloersch:
This doc has a lot of good information but the PR is pretty stale! I'm going to merge and we can do a second round of revisions with another PR.

PR #11, opened 2021-09-30
Update spec to use block oracle terminology

karlfloersch:
**Description**
This PR renames `state oracle` -> `block oracle`. The reason is because it seems advantageous to finalize block hashes as opposed to state root hashes considering it is not added complexity & will allow us to prove the validity of logs.

**Additional context**
One could argue that the term 'block' is less general than state & being super generalized is beneficial. However, doesn't seem obvious to me that the benefits of generality outweigh the nice mental model that finalizing blockhashes provides. This is **especially** true considering we can use an L1 Block Oracle for fraud proofs, and then use an L2 Block Oracle for withdrawals, and each can share the same APIs. Being able to reuse code there & concepts there seems very beautiful.


karlfloersch:
@maurelian and I discussed this & decided that `block oracle` > `block hash oracle`.

@ben-chain very curious if you have opinions on this.

ben-chain:
I think that L2 Block Oracle is good.  The only confusing thing to watch out for is that we also want an L1 blockhash oracle, which is has some chances for name collisions/confusions. Overall I think I still approve of this PR.

maurelian:
> One could argue that the term 'block' is less general than state

IMO the term 'state' is excessively general, but also overloaded with possible specific meaning, as it's common to use 'state' to refer to accounts and storage. For example, a solidity dev is likely to say "`view` functions do not change state". That is wrong in the sense that they cost gas and reduce the `balance` of `tx.origin`. 

'Block' seems like a better defined term, (and happens to contain a `stateRoot` field), so Iike that.

karlfloersch:
Alright I'll merge!

PR #16, opened 2021-10-02
Add block generation python sim & writeup

karlfloersch:
**Description**
Adds a description of the simple sequencer block generation algorithm. Additionally it adds a simulation for both a really simple rollup (every first L1 tx is considered a rollup block), and a simulation for that simple sequencer algorithm.

The simulation is still very much a WIP. Improvements include:
- Operate on actual block data structures by importing pyevm.
- Change the type of blocks generated in the `generate_rollup_epoch(....)` function to generate block inputs instead of blocks. Then we can break out the actual rollup state transition function from the generation of all the block inputs.


components/block_gen.md
```diff
@@ -0,0 +1,102 @@
+# Block Generation
+
+The logic which is used to generate the rollup chain from L1.
+
+## Glossary
+
+- **Block Inputs**: All information in an Ethereum block required to generate the full block contents.
```
ben-chain, line 7:
```suggestion
- **Block Inputs**: All information required to generate a full Ethereum bock, given the pre-state and previous block. For example, does NOT include a state root for that block's input.
```

components/block_gen.md
```diff
@@ -0,0 +1,102 @@
...
+- **Rollup Blockchain**: The rollup blocks generated by performing the rollup's state transition function over all block inputs.
```
ben-chain, line 8:
```suggestion
- **Rollup Blockchain**: The rollup blocks generated by performing the rollup's state transition function over all block inputs. For example, DOES include a state root for that block's input.
```

ben-chain: could be worth just specifying this explicitly, what is NOT in block input that IS in yellow paper section `4.3. The Block`

karlfloersch:
Note that there is still code in this branch which could be useful because it starts to implement the algorithm laid out in https://github.com/ethereum-optimism/optimistic-specs/pull/29 ; however, there are bugs in the code / not enough tests which means that I think it's worth keeping this branch around but should not be considered a real PR. We need to update this and actually implement the rollup driver

PR #28, opened 2021-11-01
Update roadmap diagram

karlfloersch:
**Description**
Updated the roadmap:
- Add infrastructure stability after v0.5 regenesis
- Simplify the separation of concerns for v1.0
- Add some "future optimizations" to reflect a more accurate list of what we might want to do to push the chain to the next level.
- Update the terminology in the fraud proof track.
- Moved the boxes around so they are more chronologically ordered


karlfloersch:
@snario I'ma merge to keep things moving and then we can go over this together & then add a new PR with edits.

PR #29, opened 2021-11-02
Add block generation writeup

karlfloersch:
**Description**
Adds the block generation write up. This logic defines much of the logic we use for transforming the L1 chain into the rollup chain. 

Some key highlights:
- I would love feedback a change that I am considering. Specifically the change is to change `SequencerBlock` to `SequencerBlockInput` or even `SequencerInput` in the writeup (and the same for `DepositBlock`).
- I made the executive decision to add the timestamp logic into the spec. It is relevant for block generation and I am convinced that it is critical in simplifying EIP1559.
- I updated the broken links to the block gen MD

This PR replaces https://github.com/ethereum-optimism/optimistic-specs/pull/16 and only contains the writeup, not the simulation.

---

Todo: Change `Deposit Block` and `Sequencer Block` to `Deposit Feed Element` and `Sequencer Feed Element`


components/rollup_node/block_gen.md
```diff
@@ -0,0 +1,147 @@
+# Block Generation
+
+The logic which is used to generate the rollup chain from L1.
+
+## Glossary
+
+- **Block Inputs**: All information required to generate a full Ethereum bock, given the pre-state and previous block. For example, does NOT include a state root for that block's input.
```
tynes, line 7: typo `bock` -> `block`

components/rollup_node/block_gen.md
```diff
@@ -0,0 +1,147 @@
...
+- **Rollup Blockchain**: The rollup blocks generated by performing the rollup's state transition function over all block inputs. For example, DOES include a state root for that block's input.
+
+## Summary
+
+The rollup chain can be deterministically generated given an L1 Ethereum chain. The fact that the entire rollup chain can be generated based on L1 blocks is _what makes OE a rollup_. This process can be represented as:
+
+```python
```
tynes, line 14: Perhaps remove the python label here since it isn't python?

components/rollup_node/block_gen.md
karlfloersch: 😄 whoops fixed!

components/rollup_node/block_gen.md
```diff
@@ -0,0 +1,147 @@
...
+- **Block Inputs**: All information required to generate a full Ethereum block, given the pre-state and previous block. For example, does NOT include a state root for that block's input.
+- **Rollup Blockchain**: The rollup blocks generated by performing the rollup's state transition function over all block inputs. For example, DOES include a state root for that block's input.
+
+## Summary
+
+The rollup chain can be deterministically generated given an L1 Ethereum chain. The fact that the entire rollup chain can be generated based on L1 blocks is _what makes OE a rollup_. This process can be represented as:
+
+```python
+f(l1_blockchain) -> rollup_blockchain
+```
```
tynes, line 16: Could you define `f` here?

components/rollup_node/block_gen.md
```diff
@@ -0,0 +1,147 @@
...
+
+In this document we define a block generation function which is designed to be:
+
+1. Context independent given a constant sized subset of the L1 chain.
+2. Support sequencers and sequencer consensus.
+3. Resilient sequencer censorship.
+
+## Rollup Epochs
+
+The rollup chain is subdivided into epochs. There is a 1:1 correspondence between L1 block numbers and epoch numbers. For L1 block number `n`, there is a corresponding rollup epoch `n` which can only be generated after L1 block number `n` is added to the L1 chain. Epoch contain one to many rollup blocks.
```
tynes, line 26: `Epoch` -> `An epoch`

components/rollup_node/block_gen.md
```diff
@@ -0,0 +1,147 @@
...
+
+### Types of Blocks within Epochs
+
+Within these epochs, there are two block types:
+
+1. Deposit block
+2. Sequencer block
+
+Deposit blocks contain contextual information about L1 such as the block hash. They also contain transactions initiated on L1 by any user or contract.
```
tynes, line 35: `... on L1 by any user or contract that execute on L2` or something along those lines

components/rollup_node/block_gen.md
```diff
@@ -0,0 +1,147 @@
...
+
+Time can be manipulated by the sequencer; however, it must stay within reasonable bounds and must be monotonic (ie. can never go backwards). The specific conditions that must be met are:
+
+```python
+# last_l1_timestamp is the L1 timestamp set during the last ContextDeposit
+# last_timestamp is the timestamp from the previous rollup block
+# average_l1_block_time is the average time between two L1 blocks
+# sequencing_window_size is the number of L1 blocks the sequencer has to submit their blocks
+min_timestamp = max(last_l1_timestamp, last_timestamp)
+max_timestamp = last_l1_timestamp + sequencing_window_size * average_l1_block_time
```
tynes, line 120: We should be sure that this is safe when it comes to post-merge. If i understand correctly, the blocktime is basically deterministic and will effectively double if a slot is missed. I think it will be, but potentially good to note here eventually

components/rollup_node/block_gen.md
```diff
@@ -0,0 +1,147 @@
+# Block Generation
+
+The logic which is used to generate the rollup chain from L1.
```
tynes, line 3: What do you think about expanding this to clearly say that L1 is used as a data availability layer to generate the rollup chain and state?

tynes: This is syncing from L1 from the PoV of a verifier/fresh sequencer - guessing the block generation from the PoV of the sequencer would go in the batch submitter doc?

components/rollup_node/block_gen.md
```diff
@@ -0,0 +1,147 @@
...
+assert sequencer_suggested_timestamp >= min_timestamp
+assert sequencer_suggested_timestamp <= max_timestamp
+```
+
+The timestamp exposed on the rollup chain is:
+
+```python
+# last_sequencer_suggested_timestamp is the most recent valid sequencer_suggested_timestamp
+def timestamp():
+    max(last_l1_timestamp, last_sequencer_suggested_timestamp)
```
ben-chain, line 131: I think that setting the value in this way contradicts above where we say it's a validity condition/assertion. FWIW I'm more of a fan of setting the timestamp for the sequencer if they screw up, seems like a better UX to still apply the transactions albeit "late."

karlfloersch: Ah so I intended for this function to be used to ensure the `timestamp` increases even when there haven't been sequencer feed elements in a very long time. So the validity condition only relates to sequencer feed elements, but this function is used to determine the block timestamp at all times.

components/rollup_node/block_gen.md
karlfloersch: Great callout. I think it would be worth explicitly calling out the data availability layer

components/rollup_node/block_gen.md
```diff
@@ -0,0 +1,147 @@
...
+#### **Footnote**: The DepositFeed ensures that no matter the size of the event it is possible to prove the deposit data to the fraud proof. See the DepositFeed spec for details.
+
+## Sequencer Blocks in Depth
+
+Sequencer blocks contain a majority of the transactions submitted to the rollup chain. Sequencer blocks can also influence the current timestamp of L2 which is exposed to contracts & is used in the EIP1559 fee pricing calculations.
+
+Sequencer blocks are of the following form:
+
+```python
+class SequencerBlock:
```
ben-chain, line 96: Currently this type does not allow for/delineate multiple L2 blocks. Presumably we could just add an array of `num_transactions, time_increase` which would say how many of the `transactions` to pull into the next L2 block, and how much to add to the previous L2 block's timestamp.

karlfloersch: Ah so I intentionally left out the batch submitter encoding scheme because it was going to bloat the scope a bit. My intention is that after you decode the batches that it results in an array of `SequencerBlock`s (aka `SequencerFeedElements`) and those eventually become blocks. Then we can work on optimizing the encoding scheme in another PR. That said, I do think that there is a very simple encoding scheme which should work & I'll write up my strawman soon

components/rollup_node/block_gen.md
karlfloersch: I don't think it will be an issue because the actual 'max timestamp' is independent of the L1 block time. I just used the L1 block time as the config because otherwise I would have to introduce another "window", this time a "suggested timestamp window". So the updated equation would be:

```
max_timestamp = last_l1_timestamp + suggested_timestamp_window_size
```

So the way that I wrote it is a little misleading but writing it this way introduces another concept which scared me a bit.

I'm super interested in feedback on:
1. Which approach you like better
2. If there's a better variable name for this than `suggested_timestamp_window_size` 😅😭

components/rollup_node/block_gen.md
ben-chain: Ah, got it. I would be in favor of making the `timestamp()` function purely on the *current* epoch sequencer suggestion (if present) and the last l2 timestamp. If I'm understanding your proposal right, then `last_sequencer_suggested_timestamp` would otherwise have to be a new thing persisted in state, correct?

karlfloersch: Hmm so I'm having trouble figuring out how to get around persisting `last_sequencer_suggested_timestamp` in state. I think you need either `last_sequencer_suggested_timestamp` or you need to persist `last_timestamp`. This covers the case where:

1. Deposit block comes in setting `last_l1_timestamp=1`
1. Sequencer sets timestamp to `last_l1_timestamp + 10`
1. Sequencer goes offline
1. Deposit block is processed which updates increases the `last_l1_timestamp` to `5`

What is `timestamp` in this case? Presumably the timestamp would not increase past the sequencer suggested timestamp until the `last_l1_timestamp` increases naturally to greater than `10`.

I figured it would be a little cleaner to store the `last_sequencer_timestamp` vs storing `last_l2_timestamp` because it gives you a better sense of where the timestamp number is coming from. However, I'm not super opinionated.

components/rollup_node/block_gen.md
karlfloersch: Hmm @tynes we're going to add another doc which introduces block gen & feeds & whatnot. I think that will be the best place to add extra detail about availability layers. For now I just added the word "availability layer" but you're totally right that we should try to define what we need / mean by an availability layer because then it will become clearer to readers what we need to do to support eth2 shards

components/rollup_node/block_gen.md
karlfloersch: Seems `last_l2_timestamp` is better because we don't have to store an extra value and can just look back into the previous block. That said, it's an implementation detail

karlfloersch: Maybe `last_l2_timestamp + 1` so that two blocks never have the same timestamp

karlfloersch:
Just had conversation with @ben-chain . We've determined that `block input` being contextual is not as useful as a `block input feed`. This is because determining the context requires executing the previous block, and at that point you might as well just execute the next block and store that. 

Instead we will call it the `block input feed` and that feed will preserve the context independent property.

components/rollup_node/block_gen.md
karlfloersch: Right now ETH deposits are a special kind of "enqueued" transaction. We may consider in the future making ETH deposits special (not require paying L2 gas) and then have another type of more generic deposit which actually has to pay gas on L2. But right now everything goes through the "enqueue" pathway.

Also, in this spec we're trying to move away from the "enqueue" term and towards "deposit" but you'll still notice enqueue terminology is used often in our current docs & convos.

components/rollup_node/block_gen.md
karlfloersch: Ah so we might actually have too many terms at this point which makes it confusing.

A "batch" is just a single transaction from the sequencer to the sequencer feed with any number of blocks that it can fit. What makes this "batch" term a bit more confusing now is that we can actually submit transactions to the sequencer feed that include partial blocks. So you don't need to submit a batch that contains a specific number of blocks but instead submit a batch with as much info as possible and then start the next batch off continuing where you left off.

The fact that a "batch" is just the sequencer appending to the sequencer feed makes me a little less into the term. Hmm... one of the major edits I think we'll have to make once we've sketched out the design is do a mega bikeshed to determine what terms are useful and what terms are overkill.

components/rollup_node/block_gen.md
```diff
@@ -0,0 +1,147 @@
...
+
+The rollup chain can be deterministically generated given an L1 Ethereum chain. The fact that the entire rollup chain can be generated based on L1 blocks is _what makes OE a rollup_. This process can be represented as:
+
+```python
+f(l1_blockchain) -> rollup_blockchain
+```
+
+In this document we define a block generation function which is designed to be:
+
+1. Context independent given a constant sized subset of the L1 chain.
```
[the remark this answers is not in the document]
karlfloersch: > Is this intended to mean that the generation function is a pure function of the L1 chain subset?

Yes that's right. However, I will modify this definition to instead be something along the lines of:

1. Require no state other than what is easily accessible in an L1 EE or L2 EE.

The key property that I really wanted was that the rollup driver doesn't impose **additional** state on top of L1 / L2. This way we don't need to "sync" the rollup driver and maintain a special database for it.

components/rollup_node/block_gen.md
```diff
@@ -0,0 +1,147 @@
...
+2. Sequencer block
+
+Deposit blocks contain contextual information about L1 such as the block hash. They also contain transactions initiated on L1 by any user or contract.
+
+Sequencer blocks are submitted by the sequencer and target _future_ epochs which satisfy the following two conditions:
+
+1. Target epoch is larger than the current L1 block number.
+2. Target epoch is less than the current L1 block number PLUS `sequencing_window_size`.
+
+The ability for the sequencer to append sequencer blocks to future epochs allows the sequencer to predict and influence the state of the epoch before the L1 chain has mined it. This property is what enables _fast transaction confirmations_ via the sequencer replicator.
```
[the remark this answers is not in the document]
karlfloersch: > Is it necessary for the sequencer to pick an epoch number in order to promptly confirm the the transaction?

If by "promptly confirm" you mean both confirm that the transaction will be included **and** _compute the result of the transaction_ then yes. Of these two, the more important property is the latter property of predicting the future state. It is critical that the sequencer can predict future state because 1) the sequencer can tell users what the chain will look like in the future with low latency, 2) the sequencer can be confident the user has a high enough balance to pay their fee.

The intuition for why this is possible is because the rollup chain is made up of deposit blocks & sequencer blocks. The sequencer knows the deposit block for a future epoch `sequencer_window_size` blocks in advance. And then the sequencer is able to themselves determine what sequencer blocks to submit. So even before they've submitted their blocks on chain, they know that assuming they are able to submit them in time they can pre-compute the state of L2 even before L1 reflects it.

karlfloersch:
Addressed the feedback and talked with @ben-chain who gave me the green light to merge and then we can start iterating on this. As proto says it's easier to iterate with additional PRs vs only code review.

components/rollup_node/block_gen.md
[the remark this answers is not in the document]
norswap: > Require no state other than what is easily accessible in an L1 EE or L2 EE.

What kind of "or" is it:
1. We can require state available on L1, or state available on L2 (but not (yet?) on L1), or
2. We can require state that is available on L1 and on L2?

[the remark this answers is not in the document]
ben-chain: Calculating the block for epoch `N` should be possible with only:
1) The data from the recent L1 blocks `[N - sequencing_window, N]`.
2) The data or state from L2 after epoch `N-1`.

In other words, "easily accessible" basically means readable.  So if there are multiple L2 blocks inputs for epoch, you should not have to execute any of them to know all of them.

>  state available on L2 (but not (yet?) on L1)

If by this you mean does the data need to be finalized (passed challenge window) on L1? If so, the answer is no.  This data is accessible in the fraud proof will build on a previous assertion, meaning that it can accept the L2 state from the last epoch as an input.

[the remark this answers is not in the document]
norswap: I meant state not yet posted on L1, but it seems to me this is purely a L1 thing, I was just tripped up by the reformulation to "Require no state other than what is easily accessible in an L1 EE or L2 EE."

PR #31, opened 2021-11-08
Add fraud proof VM & program writeup

karlfloersch:
**Description**
Adds an updated description of the fraud proof VM and the program which is run inside of the fraud proof VM during a fraud proof.

Credit to @ben-chain for the sub-process concept - it does make conceptualizing the fraud proof much cleaner.

### DIFF
--- a/components/fraud-proof-vm/fraud_proof_program.md
+++ b/components/fraud-proof-vm/fraud_proof_program.md
@@ -1,0 +1,43 @@
+# Fraud Proof Program
+
+The fraud proof program is the specific program which is run on a [fraud proof VM](./fraud_proof_vm.md) during a dispute. This program is designed to evaluate any subset of the Optimistic Ethereum chain.
+
+## Initialization
+
+The OE fraud proof program is used to weed out invalid block hash assertions in the `Block Hash Oracle`. The fraud proof program is initialized using values which are recorded at the time of block hash assertions. This is all handled within the `AssertionManager` contract.
+
+Every time an L2 block hash is asserted using the `AssertionManager`, the following information is hashed together and recorded:
+
+- Latest L1 block hash at the time of assertion.
+- Asserted next L2 block hash.
+- The previous assertion this assertion builds upon.
+- Epoch number of the L2 block hash.
+
+The fraud proof program takes in as input the hash of all these values.
+
+## Execution
+
+The fraud proof program is designed to execute the full state transition between any two epochs. The epochs used for a dispute are the previous assertion epoch & the new assertion epoch number in the `AssertionManager`. The fraud proof program will execute the state transition between these epochs and determine the final epoch block hash (ie. the one valid assertion). If the asserted L2 block hash is deemed incorrect over the course of a dispute game, the `AssertionManager` will delete the invalid assertion and forfeit the asserter's bond.
+
+In practice, the fraud proof program is broken up into two separate processes:
+
+1. Block generation
+2. Block processing
+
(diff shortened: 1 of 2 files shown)


norswap:
test

norswap:
test2

components/fraud-proof-vm/fraud_proof_vm.md
```diff
@@ -0,0 +1,50 @@
...
+1. Program execution can be split up into deterministic state transitions.
+2. Each state transition can be evaluated on-chain with less than or equal to a constant size gas allocation (eg. 10 million L1 gas).
+3. The FPVM must be executable off-chain in such a way that state checkpoints (ie. merklized commitments to the VM state) can be generated.
+
+## Usage
+
+### Initialization
+
+The fraud proof VM is initialized with:
+1. an executable binary; and
```
norswap, line 17: This is the fraud proof program, link to fraud_proof_program.md

components/fraud-proof-vm/fraud_proof_vm.md
```diff
@@ -0,0 +1,50 @@
...
+2. an initialization hash for the preimage oracle.
```
norswap, line 18: What exactly is this hash? I suspect it is the last L1 block hash against which all proofs must be proven?

components/fraud-proof-vm/fraud_proof_vm.md
```diff
@@ -0,0 +1,50 @@
...
+
+```python
+def get_preimage(hash: bytes) -> bytes:
+    # The preimage is looked up from a local database held by the fraud prover
+    preimage = fetch_preimage(hash)
+    assert keccak256(preimage) == hash
+    return preimage
+```
+
+The preimage oracle must be populated with all `hash->preimage` mappings before they are queried by the FPVM. These preimages are committed to in the initialization hash which is supplied as one of the two inputs to the FPVM. During the course of execution, the FPVM must **never** request a preimage which is unavailable as that would result in the FPVM being indeterminate. In this way, a valid executable binary must account for the structure of the initialization hash and only query available preimages.
```
norswap, line 33: > must never request a preimage that is unavailable

Nevertheless, this could happen due to user actions. What happens then? --> This probably depends on the context in which the FPVM is run!

norswap: Also, I think it's good to be a bit more precise for "are committed to in the initialization hash" (btw,extra "in". Could be good to say that all hash->preimage mapping must be proven against the initialization hash using a sequence of Merkle proofs. Probably good to link to a document that explains this, preferrably with some examples.

components/fraud-proof-vm/fraud_proof_program.md
```diff
@@ -0,0 +1,43 @@
...
+
+The fraud proof program is designed to execute the full state transition between any two epochs. The epochs used for a dispute are the previous assertion epoch & the new assertion epoch number in the `AssertionManager`. The fraud proof program will execute the state transition between these epochs and determine the final epoch block hash (ie. the one valid assertion). If the asserted L2 block hash is deemed incorrect over the course of a dispute game, the `AssertionManager` will delete the invalid assertion and forfeit the asserter's bond.
+
+In practice, the fraud proof program is broken up into two separate processes:
+
+1. Block generation
+2. Block processing
+
+We split the program into these two distinct parts to isolate the complexity of block processing. This way if there is an error which causes a single block to fail (see [this discussion](https://github.com/ethereum-optimism/optimistic-specs/discussions/22) for details), it does not break the entire assertion game. While the block processing may fail, it is critically important that the block generation process never fails.
+
```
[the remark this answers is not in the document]
karlfloersch: Ah so if calculating the block fails, we throw away the block. However even if we throw away an individual block, the rest of the blocks still remain

norswap:

PR #33, opened 2021-11-10
Update rollup node readme, replicator -> p2p spec

protolambda:
Reorganize rollup-node specifications, start on p2p doc.

**Description**

- Update rollup node readme (overview of services)
- Replace sequencer replicator by p2p approach
- Network spec


components/rollup_node/p2p_interface.md
```diff
@@ -0,0 +1,206 @@
+# P2P Interface
+
+This peer-to-peer (P2P) network service in the rollup node is optional,
```
norswap, line 3: Isn't it also used by all nodes to forward L2 nodes acquired from L1 to the execution engine?

norswap: I think there should a paragraph explaining what the p2p interface does (e.g. discover peers (& who those are), propagate new L2 sequence blocks, L1 blocks, ...).

I'll also echo my other comment that "P2P Interface" is a sub-optimal name -- an interface is not a component and the organization of the doc is pretty much listing components at the moment. Maybe "P2P layer" or "networking layer"?

Also, this doesn't really describe an interface, but rather the internal organization of sub-components whose interface is specified in other external documents!

components/rollup_node/p2p_interface.md
```diff
@@ -0,0 +1,206 @@
...
+- `PeerID`: a LibP2P specific ID derived from the pubkey (through protobuf encoding, typing and hashing)
+- `NodeID`: a Discv5 specific ID derived from the pubkey (through hashing, used in the DHT)
+- `Multi-address`: an unsigned address, containing: IP, TCP port, PeerID
+- `ENR`: a signed record used for discovery, containing: IP, TCP port, UDP port, signature (pubkey can be derived) and L2 network identification. Generally encoded in base64.
+
+### Discv5
+
+#### Structure
+
+The Ethereum Node Record (ENR) for an Optimism rollup node must contain keyed values:
```
norswap, line 32: What do you mean by "keyed" here?

components/rollup_node/p2p_interface.md
```diff
@@ -0,0 +1,206 @@
...
+
+- An IPv4 address (`ip` field) and/or IPv6 address (`ip6` field).
+- A TCP port (`tcp` field) representing the local libp2p listening port.
+- A UDP port (`udp` field) representing the local discv5 listening port.
+- An Optimism (`optimism` field) L2 network identifier
+
+The `optimism` value is encoded as the concatenation of:
+- chain ID (`varint`)
+- fork ID (`varint`)
```
norswap, line 41: We probably want to explain / link to what those are.

It would also be interestig to know why Optimism needs those.

- chain ID: I don't really know how discv5 works, but I'm assuming these mesages will intermingle with discv5 messages sent by L1 nodes and this value signals L2 capabilities?
     - does the presence of the optimism identifier signals that the node should not receive L1 messages, or will it receive both L1 and L2 messages?
- fork ID: why does L2 need this but not L1?

components/rollup_node/p2p_interface.md
```diff
@@ -0,0 +1,206 @@
...
+
+Note that DiscV5 is a shared DHT: the L1 consensus and execution nodes, as well as testnet nodes,
```
norswap, line 43: introduce Distributed Hash Table acronym

components/rollup_node/p2p_interface.md
```diff
@@ -0,0 +1,206 @@
...
+and even external IOT nodes, all communicate records in this large common DHT.
```
norswap, line 44: IOT = internet of things? what are those in this context?

components/rollup_node/p2p_interface.md
```diff
@@ -0,0 +1,206 @@
...
+
+The discovery process in Optimism is a pipeline of node records:
```
norswap, line 46: Explicitly mention what should be read in order to be able to understand this explanation (the DiscV5 spec?)

components/rollup_node/p2p_interface.md
```diff
@@ -0,0 +1,206 @@
...
+#### Protocol negotiation
+
+[Multistream-select 1.0][multistream-select] (`/multistream/1.0.0`) is an interactive protocol
+used to negotiate sub-protocols supported in LibP2P peers. Multistream-select 2.0 may be used in the future.
+
+#### Identify
+
+LibP2P offers a minimal identification module to share client version and language.
+This is optional and can be disabled for enhanced privacy.
+It also includes the same protocol information, which can speed up initial connections.
```
norswap, line 85: Not sure what that last sentence means.

Are there other advantages in sharing client version & language? (btw language? programming language??)

components/rollup_node/p2p_interface.md
norswap: Actually, an API or programmatic interface to this component would be a really useful thing to have here.

norswap:
Some preliminary comments! I'll need to read the various networking specs to go more in-depth on this.

norswap (commented):
Some preliminary comments! I'll need to read the various networking specs to go more in-depth on this.

components/rollup_node/README.md
```diff
@@ -1,62 +1,109 @@
...
+## Services
 
-| Node Type \ Component | Execution Engine | Rollup Driver | Sequencer Replicator | Block Producer | Batch Submitter |
-|-----------------------|------------------|----------------------|----------------------|----------------|-----------------|
-| Verifier  | X                | X                    | optional                    |                |                 |
-| Sequencer             | X                | X                    |                      | X              | X               |
+### L2 Execution Engine
 
-## Components
+The [Execution Engine][exec-engine] (EE) is *separated from the rollup node*, following the separated consensus/execution layer design of L1.
```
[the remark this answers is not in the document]
protolambda: No, meant to say separate from the node. The client-bindings to the engine are part of the node, but the engine itself is very much a thing that runs on its own.

And it does explain what the engine does, but a bit implicit; it follows the Consensus-execution split of L1, i.e. it does all the same execution tasks as an engine would do for L1. The few differences are then explained below. Will add some clarification.

components/rollup_node/README.md
```diff
@@ -1,62 +1,109 @@
...
+The consensus-layer of Optimism.
+
+This design mirrors the separation of L1 consensus- and execution-layer to L2,
+for maximum compatibility and minimal complexity.
 
 ## Summary
 
-The Rollup Node is a consensus client that determines the latest state of the rollup. It reads state from L1, and possibly the sequencer, to compute the L2 state. The [block generation spec][block-gen] defines the rules by which L1 state is converted into L2 blocks.
+The Rollup Node is a consensus client that tracks the rollup:
+- Fork-choice: following L1
```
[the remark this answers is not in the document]
protolambda: This is the summary part, shorter is better. Will clarify it just a bit, but don't think this is the right place to explain forkchoice in-depth.

components/rollup_node/README.md
```diff
@@ -1,62 +1,109 @@
...
+- Input Data: deposits and sequencer data extracted from L1
+- Latest blocks: optimistically via L2 p2p, overrideable by L1
+
+The input data and forkchoice information collected from L1 (and optimistically via p2p)
```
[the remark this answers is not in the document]
protolambda: >  optimically because we don't wait for L1 confirmations?

Exactly

> "via p2p": as-is it's more confusing than nothing - should expand or remove

Again, this is the shortest summary on top of readme doc. This is not the place to go in-depth about p2p stack. And yes, that's exactly what it means: use the p2p interface to float around these blocks. I'll change it to "propagate optimistically via p2p interface", hope that helps.

components/rollup_node/README.md
```diff
@@ -1,62 +1,109 @@
...
 
-The sequencer replicator connects directly to the sequencer. It requests pending L2 blocks which have not yet been batch submitted, and inserts them into the tip of the local EE.
+### Block Producer
 
-These blocks are marked in the EE as not yet confirmed on L1. Replicating verifiers still run a CL, which will later mark them as L1-confirmed within the EE.
+The [Block Producer][block-producer] is an optional service to run as sequencer, *within the rollup node*,
+to progress the L2 state in advance of batch submission.
 
-### [Block Producer][block-producer]
+When tasked to produce a new rollup block, the rollup node adds `PayloadAttributesV1` to the `engine_forkchoiceUpdated`
```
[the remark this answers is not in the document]
protolambda: It's using the naming from the Engine API specification, I'll add a note to look there.

components/rollup_node/README.md
```diff
@@ -1,62 +1,109 @@
...
-These blocks are marked in the EE as not yet confirmed on L1. Replicating verifiers still run a CL, which will later mark them as L1-confirmed within the EE.
+The [Block Producer][block-producer] is an optional service to run as sequencer, *within the rollup node*,
+to progress the L2 state in advance of batch submission.
 
-### [Block Producer][block-producer]
+When tasked to produce a new rollup block, the rollup node adds `PayloadAttributesV1` to the `engine_forkchoiceUpdated` 
+call to instruct the engine to start sequencing a block of pending contents from the transaction pool.
 
-The block producer service is run by the sequencer to progress the L2 state in advance of batch submission. The block producer repeatedly calls `engine_preparePayload` to get a new L2 block and then uses `engine_executePayload` to insert the resulting block into the state. When new L1 blocks appear, it passes the corresponding deposit transactions to `engine_preparePayload`.
+This execution block is then retrieved with `engine_getPayload`, to then be pushed to L1 for availability,
```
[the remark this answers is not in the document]
protolambda: A L2 block is the same as an execution payload. It's up to the batch-submitter to combine the blocks efficiently in transactions to the data-layer. The packing is not the job of the producer, but the general idea that it will be published on both L1 (data avail) and L2 (p2p) is there. I'll try to clarify it.

components/rollup_node/README.md
```diff
@@ -1,62 +1,109 @@
...
+The [P2P Interface][p2p-interface] service enables verifiers to stay in sync with the tip of the rollup with lower latency, and sequencers to distribute the server work.
+
+Staying in sync with the tip requires rollup nodes to learn about data more directly, faster than the L1 can make them available.
+
+This is solved with an optimistic process:
+1. Sequencer publishes the L2 block through a lightweight L2 P2P network (single gossip topic)
+2. Verifier listens on the P2P network, verifies sequencer signature, and applies the block to their L2 EE.
+3. If the data cannot be confirmed on L1, or if conflicting data is confirmed, then the bad L2 block is reorganized away.
+
+The amount of L2 blocks that can be optimistically applied to the L2 EE is constrained by the supported reorg-depth by the execution engine:
```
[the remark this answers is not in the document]
protolambda: I will try to make it clearer, but I think specs should keep things brief where possible, this is just a small detail.

components/rollup_node/README.md
```diff
@@ -1,62 +1,109 @@
...
+The [Batch Submitter][batch-submitter] service takes a queue of locally produced payloads,
+and submits the corresponding transactions in batches to L1.
 
 The batch submitter encodes these transactions based on the rules of [block generation][block-gen], effectively as the reverse of the process carried out by the rollup driver. As a result:
 - Blocks previously inserted by the Block Producer into the Sequencer's EE will now appear in L1 Verifiers' EEs.
-- Blocks marked as sequencer-confirmed in both the Sequencer and in Replicating verifiers will now be marked as L1-confirmed.
+- Blocks marked as sequencer-confirmed in both the Sequencer and in Replicating verifiers will now be recognized as L1-confirmed.
+
+### P2P Interface
+
```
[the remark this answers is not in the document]
protolambda: No, the p2p interface is only tasked with propagating the blocks from L2 to L2. There is no mention of L1 to L2 propagation, because that is not the goal (for L1 reading, the node can just pull it from its L1 RPC connection, which is not a form of P2P, as the node is trusted). And the intention is not to replace "following L1", but to stay closer to the head of the L2, and defer verification with L1 to when the data actually gets on L1.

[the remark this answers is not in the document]
protolambda: I'll try clarify the trust model of L1 vs L2 more here

components/rollup_node/README.md
```diff
@@ -1,62 +1,109 @@
...
+[EIP-2718]: https://eips.ethereum.org/EIPS/eip-2718
+[EIP-4396]: https://eips.ethereum.org/EIPS/eip-4396
 [block-gen]: ./components/rollup_node/block_gen.md
 [exec-engine]: ./exec_engine.md
 [rollup-driver]: ./consensus_layer.md
-[sequencer-replicator]: ./sequencer_replicator.md
 [block-producer]: ./block_producer.md
-[batch-submitter]: ./batch_submitter.md
\ No newline at end of file
+[batch-submitter]: ./batch_submitter.md
```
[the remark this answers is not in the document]
protolambda: Added clarfication about the P2P optionality and L1 serving the worst-case. P2P-interface is not the best name, but at least it clearly distinguishes from the other parts of the rollup-node. And I strongly dislike the "sequencer-replicator" terminology, as there should not be a requirement for a direct connection to the sequencer to replicate its state.

components/rollup_node/p2p_interface.md
protolambda: > Isn't it also used by all nodes to forward L2 nodes acquired from L1 to the execution engine?

No it's not, P2P is not the same as the engine RPC. Nodes talking to eachother are not regarded as P2P communication in L1 specs, because the relation is fully trusted in the case of RPC, and it's the opposite in P2P.

The "p2p-interface" name is indeed sub-optimal, but already used on L1, so I don't see why we should re-invent the naming here when it's functionally the same.

components/rollup_node/p2p_interface.md
protolambda: key-value pairs; each value is identified by a unique key

protolambda: I'll change it to key-value pairs if that reads better

components/rollup_node/p2p_interface.md
protolambda: > We probably want to explain / link to what those are.

`varint` is very standard, it's very common in language standard libraries, and part of protobuf (google encoding). I'll reference that. The idea is to save some bytes for something that technically can be 256 bits, but generally is < 1 byte.

> I'm assuming these mesages will intermingle with discv5 messages sent by L1 nodes and this value signals L2 capabilities? 

Exactly. The DHT is more secure the bigger it is (harder to censor searching paths, more capacity), so the discovery layer is designed to be shared between protocols (already in use by both Eth1, Eth2, and the portal network in the future)

L1 has its own versioning system. Eth2 uses Forkdigests (4 bytes of hashed versioning data), and Eth1 uses an ugly combination of chain IDs, network IDs that are supposed to match, and devp2p protocol versions. In the case of this L2 we want to not collide with others (possibly copies of optimism, or just our own testnets) and support breaking changes with future upgrade, without changing the identity of the network, so it will need to contain the chain ID as well as a versioning identifier. We could hash those together like Eth2 does, but it only becomes less readable that way.

components/rollup_node/p2p_interface.md
protolambda: The discv5 spec itself does not even define the acronym, because it is assumed to be common knowledge to the reader. I'm happy to add it, but I don't think every acronym is bad.

components/rollup_node/p2p_interface.md
protolambda: yes, home devices that might want to talk to each other without servers are one of the non-blockchain use cases.

components/rollup_node/p2p_interface.md
protolambda: The discovery v5 spec is maybe too technical. The interface of the Go implementation is a good practical reference I'll add

components/rollup_node/p2p_interface.md
protolambda: Programming language yes. If you work with multiple p2p implementations, it narrows down bugs and weird behavior: very useful debug information. Similarly for the client-version, we can identify which version of the code was affected when we find a bug in external nodes that we have otherwise not much insight into.

components/rollup_node/README.md
```diff
@@ -1,62 +1,109 @@
 # Rollup Node
 
-The consensus module of Optimistic Ethereum.
+The consensus-layer of Optimism.
```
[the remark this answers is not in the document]
protolambda: We should discuss the naming. We can call it Optimistic Ethereum, but it's not a standard if it doesn't go through the same standardization process as other things. And I'm happy with the Optimism brand.

components/rollup_node/README.md
```diff
@@ -1,62 +1,109 @@
...
+- Input Data: deposits and sequencer data extracted from L1
+- Latest blocks: optimistically via L2 p2p, overrideable by L1
+
+The input data and forkchoice information collected from L1 (and optimistically via p2p)
+are then forwarded to the execution-layer, also known as "execution engine",
+via the standard [Engine API][engine-api].
+
+The design is minimalist:
+- The rollup consensus-layer is largely derived from L1 properties, enabling a simple and stateless L2 consensus.
+- The rollup execution-layer is largely shared with L1, equivalent in all ways except the addition of rollup-deposits.
```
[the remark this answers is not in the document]
protolambda: It's the same VM, with different configuration. And I definitely think we should upgrade to London when 1.0 lands (or earlier).

components/rollup_node/README.md
```diff
@@ -1,62 +1,109 @@
...
 
-1. **Sequencer**: Accepts user-sent and deposit transactions, orders them into L2 blocks, and submits them to L1 in batches. Provides RPC access to pending L2 blocks in advance of L1 submission.
-2. **Verifier**: Watches L1 feeds, reconstructs L2 block inputs, and inserts them into the EE to determine the canonical L2 state. May also track pending sequencer state, in some configurations.
+1. **Sequencer**: Accepts user-sent and deposit transactions, orders them into L2 blocks, and submits them to L1 in batches.
+2. **Verifier**: Watches L1 feeds, reconstructs L2 block inputs, and inserts them into the EE to determine the canonical L2 state.
+
+**Both** update their coupled execution-engine with L2 blocks and fork-choice instructions, to maintain their view of the rollup state.
+
+Optionally the sequencer(s) may share the L2 blocks via the L2 p2p network before the L1 can confirm them,
+which verifiers can optionally use to sync to the very tip (marked "unsafe" in JSON-RPC) of the rollup.
```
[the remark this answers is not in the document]
protolambda: Thought this was discussed in earlier spec discussion / PRs / calls,  but good idea to put a reference to offical specs here.

How I plan to mirror L1:
* unsafe - not confirmed yet
* safe - confirmed on L1
* finalized - undisputed

"latest" will be an alias to "safe" after the Merge upgrade on L1, and "unsafe" is introduced to deal with the blocks that have e.g. 1% of the attesters voting on it, and will get more votes later, but aren't so secured yet.

See https://github.com/ethereum/execution-apis/blob/main/src/engine/specification.md#forkchoicestatev1

components/rollup_node/README.md
```diff
@@ -1,62 +1,109 @@
...
+This enables permissionless L1-like latency for users, but not canonical until available on the L1.
```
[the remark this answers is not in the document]
protolambda: "finalized" would be confusing with both the other field of the forkchoice instruction ("finalizedBlockHash"), as well as the event of finality, which is not what matters here. The "latest" should be safe enough to use. It might still reorg, but is unlikely to. I'll change the wording to clarify this.

components/rollup_node/README.md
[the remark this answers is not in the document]
maurelian: A bit of a bikeshed. I propose we discuss elsewhere.

karlfloersch (approved):

components/rollup_node/p2p_interface.md
norswap: I'd personally never heard of `varint` before. A quick google search seems to show it's indeed a protobuf concept that has spread a bit.

There is a good definition [here](https://go.dev/src/encoding/binary/varint.go) for anyone reading. I think at minimum we should link to that.

We could also just remove the annotation, since we're simply inheriting another spec and this is an explanation of what it mandates. It would also be more consistent, since we don't specify for instance how the IPs are encoded.

components/rollup_node/p2p_interface.md
norswap: Make sense! In general, I'd avoid acronyms that have not been previously introduced.

components/rollup_node/p2p_interface.md
norswap: I see, thanks. Why is it "the same protocol information" vs "protocol version information" or something like that?

norswap:
Rebase on top of master (as this had merge conflicts).

PR #34, opened 2021-11-11
first draft of revamped intro

norswap:
Attempting to flesh out the spec and make thing understanble for a newcomer.

Feel free to review / answer questions on everything **BUT** I want to revisit the section on block numbering.

We had a call with Karl and Proto earlier today about some questions I had up to that point, and block numering (generally, the block generation algorithm, [as described here](https://github.com/ethereum-optimism/optimistic-specs/blob/main/components/rollup_node/block_gen.md) was the main point of discussion / explanation. Took me quite a while to reformulate my understanding in a way that makes sense, but now it needs to be rewritten to be more consistent (the start doesn't really hold with the end), maybe include an example or a super simple chart.

Btw, I figure that another thing that majorly tripped me up on the [block gen chart](https://github.com/ethereum-optimism/optimistic-specs/blob/main/assets/sequencer-block-gen.svg) was that I sort of thought it implied a synchronicity between L1 blocks and epochs. In particular, deposit blocks are always positioned in what looks like "2 blocks later", but for instance the deposit block for epoch 3 is really "synchronous" with the L1 block where S3 appears.

Naming the file `intro2.md` for ease of reviewing, will rename later. Goal is to replace [the introduction](https://github.com/ethereum-optimism/optimistic-specs/blob/main/introduction.md).

### DIFF
--- a/intro2.md
+++ b/intro2.md
@@ -1,0 +1,386 @@
+# The Optimistic Ethereum Specification Overview
+
+Optimistic Ethereum is an _EVM equivalent_, _optimistic rollup_ protocol
+designed to _scale Ethereum_ while remaining maximally compatible with existing
+Ethereum infrastructure. This document provides an overview of the protocol to
+provide context for the rest of the specification.
+
+## Table of Contents
+
+1. [Foundations](#foundations)
+1. [Network Participants](#network-participants)
+1. [User Transactions and The Sequencer](#user-transactions-and-the-sequencer)
+1. [L2 Batches and Blocks](#l2-batches-and-blocks)
+1. [L2 Block Properties](#l2-block-properties)
+1. [L1 Components Overview](#l1-components-overview)
+   - [Data Feeds](#data-feeds)
+   - [Fraud Proof Manager](#fraud-proof-manager)
+
+## Foundations
+
+### What is Ethereum scalability?
+
+Ethereum's limited resources, specifically bandwidth, computation, and storage, constrain the number of transactions which
+can be processed on the network, leading to extremely high fees. Scaling
+Ethereum means increasing the number of useful transactions the Ethereum network can process, by increasing the supply of these limited resources. This also means the fees can be made much lower.
+
+You can [follow this link][rollup-scale] to learn more about how rollups help solve Ethereum scalability.
+
+[rollup-scale]: https://hackmd.io/@norswap/rollups
+
+### What is Optimistic Rollup?
+
+[Optimistic rollup](https://vitalik.ca/general/2021/01/05/rollup.html) is a
+layer 2 scalability technique which increases the computation & storage capacity
+of Ethereum without sacrificing security or decentralization. Transaction data
+is submitted on-chain but executed off-chain. If there is an error in the
+off-chain execution, a fraud proof can be submitted on-chain to correct the
+error and protect user funds. In the same way you don't go to court unless there
+is a dispute, you don't execute transactions on on-chain unless there is an
+error. The rollup is *optimistic* because the execution results are assumed to
+be correct until a fraud proof proves them wrong.
+
+### What is EVM Equivalence?
+
+[EVM
+Equivalence](https://medium.com/ethereum-optimism/introducing-evm-equivalence-5c2021deb306)
+is complete compliance with the state transition function described in the
+Ethereum yellow paper, the formal definition of the protocol. By conforming to
+the Ethereum standard across EVM equivalent rollups, smart contract developers
+can write once and deploy anywhere.
+
+### 🎶 All together now 🎶
+
+#### Optimistic Ethereum is an _EVM equivalent_, _optimistic rollup_ protocol designed to _scale Ethereum_.
+
+## Network Participants
+
+There are three actors in Optimistic Ethereum: users, sequencers, and verifiers.
+
+<!--
+![Network Overview](./assets/network-participants-overview.svg)
+-->
+
+![Network Overview](https://raw.githubusercontent.com/ethereum-optimism/optimistic-specs/main/assets/network-participants-overview.svg)
+
+### Users
+
+At the heart of the network are users (us!). Users can:
+
+1. Deposit or withdraw tokens by sending transactions to Ethereum mainnet.
+2. Send transactions to the sequencer, to use EVM smart contracts on L2, or to
+   send ETH to other L2 users.
+3. View the status of transactions using block explorers provided by network
+   verifiers.
+
+### Sequencer
+
+As a first approximation, you can see the sequencer as the L2 block producer. It
+accrues user transactions into "batches" and submits these batches to contracts
+on L1 in order to make L2 blocks out of them (beware however, that there isn't a
+1-1 mapping between batches and blocks). Unlike on L1, on L2 transactions can be
+confirmed before a full block (or even a full batch) is created!
+
+We are going to give a lot more detail about the operation of the sequencer in
+the rest of this document.
+
+### Verifiers
+
+Verifiers monitor L1 for rollup data. They serve three purpose:
+
+1. Serving rollup data to users; and
+2. Verifying rollup integrity and disputing invalid assertions.
+3. Propagate the L2 state among validators (\*)
+
+In order for the network to remain secure there must be at least one honest
+verifier who is able to verify the integrity of the rollup chain & serve
+blockchain data to users.
+
+(\*) It's important to note that a validator does not rely on this data to
+validate the rollup — only access to the L1 chain is required. However, getting
+the most recent L2 state allows validators to serve the state to users, and to
+pre-validate transactions that should soon be posted to L1. L2 state propagation
+uses the same mechanism as L1 state sync, but must additionally include a
+signature from the sequencer to ensure the legitimacy of the data.
+
+## User Transactions and The Sequencer
+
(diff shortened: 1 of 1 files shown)


norswap:
Rewrote the section about block numbering, which is really a sneaky explanation of block generation. This is clearly too technical & in-depth, and so will need to be separated out in a different file. It also still need an example of some sort.

intro2.md
```diff
@@ -0,0 +1,405 @@
...
+although this decision is constrained by the protocol. For instance, block
+numbers assigned to transactions increase monotonically. See the section [on the
+`NUMBER` opcode](#NUMBER) for more details.
+
+Just like batches, L1 blocks can contain multiple L2 blocks, or even partial
+blocks. They can also have received zero, one, or many batches from the
+sequencer.
+
+**_L2 deposit blocks_**, on the other hand, arise from L1 blocks. There is one
+L2 deposit block per L1 block. These deposit blocks are implicitly created by
```
karlfloersch, line 165: I think that it's better to conceptualize the deposit blocks as being "created" by L1, not by the sequencer. The reason is because the main purpose of deposit blocks is to provide a mechanism by which one can interact with the rollup *without* relying on the sequencer. This is such a hard concept to communicate 😭

intro2.md
```diff
@@ -0,0 +1,405 @@
...
+
+L2 deposit blocks comprise two types of data:
+1. L1 block properties, namely
+   - block hash
+   - block number
+   - timestamp
+   - base fee
+2. L2 transactions submitted on L1
+
+These L2 transactions that have been "enqueued" on L1 are called *deposits*.
```
karlfloersch, line 178: IMO ideally we move away from the "enqueued" terminology because it builds up a mental model where there is some "queue" of L1 transactions which need to be "pulled into" the L2, vs the more modern / IMO simple mental model where they are two streams of blocks which are build built up separately and then interwoven. The sequencer feed & deposit feed are fully separate but then automatically merged by the block generation logic.

intro2.md
norswap: Just making a note to relate our discussion from today, where I make the point I'm not a fan of this framing, because the sequencer decides on L2 deposit block numbers. (The sequencer does however need to respect the rule, i.e. include the deposit block within its sequencing window.)

intro2.md
norswap: No argument here, we can remove that terminology. Was just thinking of making this inclusive of all existing terminology until it gets removed everywhere, but we could also just start here, make a note of it (maybe have a centralized terminology document or github discussion?) and remove it elsewhere afterwards.

intro2.md
```diff
@@ -0,0 +1,405 @@
...
+2. L2 sequencer blocks
+
+As it confirms transactions, the sequencer accumulates these transaction in a
+"batch". The transactions in such batches will become **_L2 sequencer block_**
+once posted to the Optimistic Ethereum contracts on L1. Please note that there
+is no simple mapping between batches and sequencer blocks: a batch may "contain"
+multiple blocks or even partial blocks. Batches are simply a means of grouping
+transactions together for submitting transactions to L1.
+
+The sequencer decides which L@ sequencer block a transaction belongs to,
```
norswap, line 155: typo

intro2.md
```diff
@@ -0,0 +1,405 @@
...
+   - [Fraud Proof Manager](#fraud-proof-manager)
+
+## Foundations
+
+### What is Ethereum scalability?
+
+Ethereum's limited resources, specifically bandwidth, computation, and storage,
+constrain the number of transactions which can be processed on the network,
+leading to extremely high fees. Scaling Ethereum means increasing the number of
+useful transactions the Ethereum network can process, by increasing the supply
```
tynes, line 26: Scalability is increasing throughput without increasing the hardware requirements, see https://twitter.com/_prestwich/status/1284174486674083840?s=20

tynes: Rollups by definition increase the throughput of the system. Users that care about validating the state can opt into validating the state

intro2.md
```diff
@@ -0,0 +1,405 @@
...
+We are going to give a lot more detail about the operation of the sequencer in
+the rest of this document.
+
+### Verifiers
+
+Verifiers monitor L1 for rollup data. They serve three purpose:
+
+1. Serving rollup data to users; and
+2. Verifying rollup integrity and disputing invalid assertions.
+3. Propagate the L2 state among validators (\*)
```
tynes, line 96: Replace `validators` with `verifiers`

intro2.md
```diff
@@ -0,0 +1,405 @@
...
+
+In order for the network to remain secure there must be at least one honest
+verifier who is able to verify the integrity of the rollup chain & serve
+blockchain data to users.
+
+(\*) It's important to note that a validator does not rely on this data to
```
tynes, line 102: I think this note is a bit too theoretical when it comes to the sequencer committing to the pre batched data. If I see a spec with how it could work in practice, then I'd be open to having it. I think for now its safe to say that the verifiers optimistically accept the data from the sequencer because users are interested in low latency transactions and that if the sequencer lies consistently then users will use a different, more reliable system.

We could hint at a cryptographically secure way of doing bond based sequencer pre-commitments if we wanted to though

intro2.md
```diff
@@ -0,0 +1,405 @@
...
+the most recent L2 state allows validators to serve the state to users, and to
+pre-validate transactions that should soon be posted to L1. L2 state propagation
+uses the same mechanism as L1 state sync, but must additionally include a
+signature from the sequencer to ensure the legitimacy of the data.
+
+## User Transactions and The Sequencer
+
+As a user, once you have [bridged] (\*) some ETH over to Optimistic Ethereum,
+you will probably interact with it though your wallet, such as Metamask. Add the
+network (chainID 10, [JSON-RPC] endpoint `https://mainnet.optimism.io/`) and
```
tynes, line 113: Perhaps good to link to this website: https://chainlist.org/

intro2.md
```diff
@@ -0,0 +1,405 @@
...
+(\*) L1 -> L2 deposits will be explained later.
+
+[bridged]: https://www.optimism.io/apps/bridges
+[JSON-RPC]: https://github.com/ethereum/execution-apis
+
+When you send a transaction, it is sent to the node called the sequencer. The
+sequencer will run verify your transaction, execute it if valid, and confirm it.
+This happens much faster than on Ethereum mainnet (around ~1s).
+
+There is currently a single sequencer operated by Optimism PBC. We expect future
```
tynes, line 125: If we are going to mention that there is a single sequencer operated by the company, then we should mention that the profits are being used to fund public goods to prevent the conflict of interest where there is no incentive to actually decentralize. While this is related to the threat model, its unclear how related it is to the protocol specification

intro2.md
```diff
@@ -0,0 +1,405 @@
...
+blocks, which are mined by proof-of-work and will soon be decided by
+proof-of-stake.
+
+Optimistic Ethereum has two kind of blocks:
+1. L2 deposit blocks
+2. L2 sequencer blocks
+
+As it confirms transactions, the sequencer accumulates these transaction in a
+"batch". The transactions in such batches will become **_L2 sequencer block_**
+once posted to the Optimistic Ethereum contracts on L1. Please note that there
```
tynes, line 150: Define a block - I think a good definition of block is a contiguous amount of gas. Transaction at the first index has access to more total gas than the transaction at the second index.

intro2.md
tynes, line 155: `L@` -> `L2`

norswap:
I know we said we're going to reboot the spec effort focusing on a simplified version first, but I'd appreciate a review on the sections going from `## L2 Blocks, Batches & Epochs` up to the end of the file.

Still need to address some of Mark's comments!

intro2.md
karlfloersch: wontfix lol

intro2.md
norswap: This language is no longer present.

intro2.md
norswap: Re-reading this, it gives the impression that there are two classes of validators: those that just trust sequencer-signed L2 state propagation (but do not read L1), and those who just read L1 (to submit fraud proofs).

Would you say that's a correct assessment? (It seems suprising to me.)

If it's true, we should endeavour to remove that distinction, because one easy argument against the verifier's dilemma is that some actors *have to* run a verifier anyway (e.g. Etherscan), so we might as well have them truly *verify* the chain.

intro2.md
norswap: We should totally define what a block is. Not sure if it should be included here or elsewhere, but we can link to the glossary for sure. I'm not a fan of calling a block a "contiguous amount of gas" because it seems to imply that a single transaction could be split amongst multiple blocks. I'd rather call it a bundle of transactions. Then the gas constraint can be expressed as "the total gas used by each transaction in the block cannot exceed the block's gas limit".

intro2.md
norswap: The specification will assume a single sequencer, so that should be mentionned. Should I just remove "operated by Optimism PBC"? That is indeed not related to the spec (as someone else can implement it.)

norswap:
Updated following review. I'm closing this PR now in preparation for the repo reset (https://github.com/ethereum-optimism/optimistic-specs/pull/40).

Please don't delete the branch!

PR #38, opened 2021-11-17
Add glossary.md

maurelian:
A new file glossary.md, which is a first attempt at compiling and clarifying the jargon in the spec.

### DIFF
--- a/glossary.md
+++ b/glossary.md
@@ -2,7 +2,7 @@
 
 ### Layer 1 (L1)
 
-The Ethereum Mainnet.
+An Ethereum network with its own native security mechanism (ie. proof of work, proof of stake or proof of authority).
 
 ### Layer 2 (L2)
 
--- a/glossary.md
+++ b/glossary.md
@@ -1,0 +1,89 @@
+# Glossary of Optimistic Ethereum Terms
+
+### Layer 1 (L1)
+
+The Ethereum Mainnet.
+
+### Layer 2 (L2)
+
+The Optimistic Ethereum Network.
+
+### State Root Oracle
+
+A smart contract on L1, providing access to historical L1 block hashes. The existence of this contract makes it possible to prove the existence of a particular L2 state root within a previous L1 block.
+
+### Block Generation (AKA Block Derivation)
+
+All of the data necessary to reconstruct the history and state of L2 is contained within L1 blocks. Block Generation is the process of converting data contained in L1 blocks into L2 blocks.
+
+### Epoch
+
+A continuous sequence of 0 or more L2 blocks. There is a one to one correspondence between L2 Epochs, and L1 blocks.
+
+**Example usage:** "Epoch 20 corresponds to L1 block 637, epoch 21 corresponds to block 638, etc".
+
+### Feeds
+
+Feeds are append only lists, each of which is contained in a smart contract on Layer 1. The state of the Rollup Chain is determined by the data written to these contracts.
+
+### Sequencer Feed
+
+The Sequencer Feed is a contract which contains the data representing sequencer submitted blocks, each element in this feed will be used to generate a "sequencer block".
+
+### Deposit Feed
+
+The Deposit Feed is a contract which contains deposit data. Each element in this feed will be used to generate a "deposit block".
+
+### Sequencing Window
+
+The number of L1 blocks during which the Sequencer _may_ insert Sequencer blocks before a Deposit block.
+
+If the Sequencer Window is 2 blocks, and Alice makes a Deposit in Block 420, her deposit WILL be included (by the Block Generation process) at block 422.
+
+### Execution Environment (EE)
+
+The virtual machine in which transactions are applied to the pre-state to generate the post-state.
+
+### Rollup Node AKA Rollup Driver
+
+The core client software which reads data from an L1 node, and transforms it into L2 data. Both the Sequencer, and Verifier nodes run a Rollup Node, albeit in different configurations.
+
+### Consensus Layer
+
+By analogy to the architecture of ETH2 we can also describe Layer 1 as the "Consensus Layer", as it defines the canonical history and state of the Layer 2 chain.
+
+### Sequencer
+
+The privileged entity which accepts L2 transactions from users, and determines the ordering of those transactions with respect to each other and deposit transactions.
+
+
+### Verifier
+
+A verifier is any Rollup Node operator who is not the Sequencer. Verifiers compute the state of the L2 chain, and compare it to the states proposed by the Sequencer.
+
+### Deposit transaction (via the Deposit Feed)
+
+A transaction which originates on L1, by way of a message sent to the Deposit Feed contract. This message can be sent either by a contract or an externally owned account.
+
+### Sequencer transaction (via the Sequencer feed)
+
+A transaction which is sent directly to the sequencer from user.
+
+### L2 State Oracle (AKA L2 Block Oracle)
+
+The on-chain component to which the Sequencer writes the L2 state roots.
+The L2 State oracle also implements the k-section game, and single step verifier.
+
+### Single Step Verifier
+
+An on-chain
+
+### Dispute game
+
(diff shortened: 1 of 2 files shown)


glossary.md
```diff
@@ -0,0 +1,89 @@
+# Glossary of Optimistic Ethereum Terms
+
+### Layer 1 (L1)
+
+The Ethereum Mainnet.
```
tynes, line 5: This technically could be a testnet. Perhaps a better definition could be an ethereum network that derives its security from proof of work/proof of stake

glossary.md
```diff
@@ -0,0 +1,89 @@
...
+
+### Layer 2 (L2)
+
+The Optimistic Ethereum Network.
```
tynes, line 9: A network that derives its security from a layer one

glossary.md
```diff
@@ -0,0 +1,89 @@
...
+
+### State Root Oracle
```
tynes, line 11: Is this supposed to be the block hash oracle? Also could link to this? https://eips.ethereum.org/EIPS/eip-210

glossary.md
```diff
@@ -0,0 +1,89 @@
...
+### Deposit Feed
+
+The Deposit Feed is a contract which contains deposit data. Each element in this feed will be used to generate a "deposit block".
+
+### Sequencing Window
+
+The number of L1 blocks during which the Sequencer _may_ insert Sequencer blocks before a Deposit block.
+
+If the Sequencer Window is 2 blocks, and Alice makes a Deposit in Block 420, her deposit WILL be included (by the Block Generation process) at block 422.
+
```
norswap, line 42: I like this! We should probably mention the concepts of epoch: the sequencing windows is the number of L1 blocks during which the sequencer may insert L2 sequencer blocks for a given epoch, i.e. after the epoch's L2 deposit block, but before the next epoch's deposit block.

Also a nitpick, but a sequencing window is a block range whose *size* is the number of L1 blocks [...]. e.g. the epoch for L1 block w/ number `X` is `[X+1, X+k]` where `k` is the window size.

glossary.md
```diff
@@ -0,0 +1,89 @@
...
+### Execution Environment (EE)
+
```
norswap, line 44: Isn't EE "Execution Engine"?

glossary.md
```diff
@@ -0,0 +1,89 @@
...
+The virtual machine in which transactions are applied to the pre-state to generate the post-state.
+
+### Rollup Node AKA Rollup Driver
+
```
norswap, line 48: We should probably choose between both of these - no particular opinion here.

glossary.md
```diff
@@ -0,0 +1,89 @@
...
+The core client software which reads data from an L1 node, and transforms it into L2 data. Both the Sequencer, and Verifier nodes run a Rollup Node, albeit in different configurations.
+
```
norswap, line 50: As far as I understand (notably referring to this diagram we charted together), the tx batch submitted and proposal submitter are also part of the rollup node, and go on the opposite direction from this (they submit L2 tx and block hashes to L1).

We might want to be more precise thant just "data" here. In particular, the thing being read from L1 are "block inputs" (or some other term we come to agreement on, but the goal is to do block derivation).

glossary.md
```diff
@@ -0,0 +1,89 @@
...
+### Consensus Layer
+
+By analogy to the architecture of ETH2 we can also describe Layer 1 as the "Consensus Layer", as it defines the canonical history and state of the Layer 2 chain.
+
```
norswap, line 54: Do we want to add these analogies / overloaded terms? I think if we do this it becomes confusing because suddenly "consensus layer" can mean either the L1 consensus layer or the whole of L1.

glossary.md
```diff
@@ -0,0 +1,89 @@
...
+### Sequencer
+
+The privileged entity which accepts L2 transactions from users, and determines the ordering of those transactions with respect to each other and deposit transactions.
+
+
+### Verifier
+
+A verifier is any Rollup Node operator who is not the Sequencer. Verifiers compute the state of the L2 chain, and compare it to the states proposed by the Sequencer.
+
```
norswap, line 63: Could be good to reuse part of the (existing) intro here that lists the goals of the verifier (validating + fraud proof, distributing state to users, and amongst validators).

glossary.md
```diff
@@ -0,0 +1,89 @@
...
+### Deposit transaction (via the Deposit Feed)
+
+A transaction which originates on L1, by way of a message sent to the Deposit Feed contract. This message can be sent either by a contract or an externally owned account.
+
+### Sequencer transaction (via the Sequencer feed)
+
+A transaction which is sent directly to the sequencer from user.
```
norswap, line 70: user**s**

glossary.md
```diff
@@ -0,0 +1,89 @@
...
+
+### L2 State Oracle (AKA L2 Block Oracle)
+
```
norswap, line 73: L2 proposal manager (cf. our conversation about how I didn't like "oracle" because it implies a very simple question-answer interface which is not the case here)

glossary.md
```diff
@@ -0,0 +1,89 @@
...
+### Layer 2 (L2)
+
+The Optimistic Ethereum Network.
+
+### State Root Oracle
+
+A smart contract on L1, providing access to historical L1 block hashes. The existence of this contract makes it possible to prove the existence of a particular L2 state root within a previous L1 block.
+
+### Block Generation (AKA Block Derivation)
+
```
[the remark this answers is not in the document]
maurelian: Only because I was expecting this to be read during an in-person session, and that some of those people would be used to the former terminology.

glossary.md
maurelian:
```suggestion
A network that derives its security from a layer one.
```

glossary.md
maurelian:
```suggestion
An Ethereum network with its own native security mechanism (ie. proof of work, proof of stake or proof of authority).
```

glossary.md
maurelian: Yes, it's the same as the `Block Hash Oracle`, but we've changed the name to this (actually I think we're even shifting towards `L2 State Root Oracle`. 
Yes, it provides block hashes, but the purpose for it is more closely tied to getting state roots for L2.

maurelian:
```suggestion
### L2 State Root Oracle
```

glossary.md
```diff
@@ -0,0 +1,89 @@
...
+All of the data necessary to reconstruct the history and state of L2 is contained within L1 blocks. Block Generation is the process of converting data contained in L1 blocks into L2 blocks.
+
+### Epoch
+
+A continuous sequence of 0 or more L2 blocks. There is a one to one correspondence between L2 Epochs, and L1 blocks.
+
```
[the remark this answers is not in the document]
maurelian: Yes, good point!
```suggestion
A continuous sequence of L2 blocks, including a single deposit block, followed by 0 or more L2 blocks. There is a one to one correspondence between L2 Epochs, and L1 blocks.```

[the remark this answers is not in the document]
Comment (norswap) [reply glossary.md:22]: Really not important, but is that last comma in the suggested change grammatical?

**Comment by norswap** [comment]
Regarding "repo regenesis" (https://github.com/ethereum-optimism/optimistic-specs/pull/40), I suggest we introduce a glossary and everyone introduces new definitions in it as needed.

Let's also immediately get into the habit of linking terms to their glossary page entry.

File: glossary.md [hunk 14]
```diff
@@ -0,0 +1,89 @@
+# Glossary of Optimistic Ethereum Terms
+
+### Layer 1 (L1)
+
+An Ethereum network with its own native security mechanism (ie. proof of work, proof of stake or proof of authority).
```
snario, line 5:
```suggestion
An Ethereum network with its own native security mechanism (e.g., proof of work, proof of stake or proof of authority).
```
https://english.stackexchange.com/questions/1629/e-g-versus-i-e

norswap:

PR #40, opened 2021-11-19
Reset repo to restart from scratch

maurelian:
Prior to this commit, the specs in this repo have grown somewhat haphazardly, making them hard to edit and evolve in the desired direction. We've agreed to start fresh. Much of the content can and will be reused as we begin filling in these specs again, though it will be subjected to review again before merging.


norswap:
A few options:
1. merge this, keep building on `main` BUT must make a new branch to preserve easy github navigation to the old stuff
2. move the `main` branch to another branch and reset the history of the `main` branch
3. create a new branch and make it the default branch on github

Option 2 is the cleanest but it edits history, which is potentially problematic when syncing up local code.
No strong opinions otherwise.

ben-chain:
Let's do 1. from @norswap 's suggestions, so we can merge this as soon as we clone current to another branch

norswap:
Modified + rebase on top of main to remove files that were added by the recently merged PRs.

Issue #5, opened 2021-09-02
What is the deposit specification?


karlfloersch:
**Is your feature request related to a problem? Please describe.**
Deposit transactions must be included in the rollup. How do we do this?

TODO: Flesh out this issue

karlfloersch:
At a high level we've come to understand that there are 2 major "types" of L1->L2 messages (deposits) that we expect to see:
1. An L1 context update (eg. block hash, block number, base fee, etc)
2. A deposit message coming from an end user.

One big question that we haven't yet answered is exactly how the block processing should work. What currently seems the most promising is updating the baseFee every time a new L1Context has been processed, and then allowing the sequencer to submit as many blocks as they'd like between L1Context updates up to a limit. This would be a EIP1559-like mechanism but over L1Contexts.

lightclient:
Is there a problem with sequencers not including L1 contexts when the base fee drops so they can keep earning a higher data fee?

lightclient:
For the actual data, maybe this?

```rust
struct Context {
    block_hash: Hash,
    number: u64,
    state_root: Root,
    transactions: Hash,
    receipts: Hash,
    base_fee: u256,
    time: u64,
}

struct Message {
    index: u64,
    timestamp: u64,
    sender: Address,
    target: Address
    gas_limit: u64,
    data: Vec<u8>,
}
```

These objects can be i) ABI encoded or ii) hand encoded. I think it would be pretty easy group the values and hand decode them context / msg manager, saving some unnecessary data posted to L1.

karlfloersch:
Noting down some findings after conversations over the weekend debating the need for sequencer priority / deposit transactions in general. Here is a short writeup of the situation:

# Block proposal mechanisms for rollups
1. Free for all
2. Sequencer priority

## Free for all
In a free for all model, anyone can submit a batch to L1 and the batch will be immediately applied to the head of the L2 state. Anyone may propose a next batch, there is no need for a special "deposit" transaction type because there is no risk of censorship by a privileged sequencer. 

#### Benefits
- No deposit transaction type
- Simple to conceptualize

#### Drawbacks
- Creates a race between block proposers which results in:
[...]

PR #43, opened 2021-11-30
Rollup Node Spec

norswap:
[>> Rendered markdown view <<](https://github.com/ethereum-optimism/optimistic-specs/tree/c65b5491affa3f3388e43fdfef1e90952be3c425)

Draft of the rollup node spec. Not final, not perfect, but should be a fairly good idea of what I'm aiming at.
To be used as a basis to further huddle on in order to refine the concept.

I've used **TODO**s liberally in the document to mark missing links and missing parts. Notably:

- [x] ~Deposits must be better specified (this might be in-scope for the deposit contract or the execution enginerather than the rollup driver).~ -> specify in deposit feed contract (issue: https://github.com/ethereum-optimism/optimistic-specs/issues/5)
- [x] Speak to some kind of finalization guarantee. Or remove that TODO, as that may be meaningless in the deposits-only stage. 
- [x] ~List out error cases, in particular resulting from the interaction with the execution engines (are there others?) and specify what must be done if encountered.~ -> temporarily postponing, tracked in https://github.com/ethereum-optimism/optimistic-specs/issues/47


rollup-node.md
```diff
@@ -0,0 +1,284 @@
...
+
+> **TODO** Enunciate our finalization guarantees.
+
+> **TODO** Must enunciate error cases more precisely, as well as state what
+> should be impossible and should cause the node to stop deriving the L2 chain.
+
+## Block Gossip
+
+Another responsability of the rollup node is to transmit the L2 blocks it
+derives from the L1 chain to other L2 nodes via a process called *block gossip*.
```
norswap, line 282: missing glossary link for *block gossip*

rollup-node.md
```diff
@@ -0,0 +1,284 @@
...
+its "cousins" (i.e. the descendant of an ancestor of the previous head). In
+those case, the rollup driver must:
+
+1. Locate the *common ancestor*, a block that is an ancestor of both the
+   previous and new head.
+2. Isolate the range of L1 blocks \]common ancestor, ..., new head\].
+3. For each such block, call [`engine_forkchoiceUpdatedOPV1`] and
+   [`engine_getPayloadV1`].
+   - Fill the [`PayloadAttributesOPV1`] object according to [the section on payload attributes][payload-attr].
+   - Fill the [``ForkchoiceStateV1`] object according to the [the section on the
```
protolambda, line 264:
```suggestion
   - Fill the [`ForkchoiceStateV1`] object according to the [the section on the
```

Fix text formatting

glossary.md
```diff
@@ -0,0 +1,241 @@
...
+engine] to construct L2 blocks.
+
+"Payload attributes" is a term that originates and is specified in the [Ethereum
+Engine API specification][engine-api].
+
+In addition to the [specified][engine-api] `timestamp`, `random`and
+`suggestedFeeRecipient` fields, we add a field to specify the transactions that
+the execution engine must include in the block and execute.
+
+> **TODO** The transaction field must be specified.
```
protolambda, line 142: This is just the glossary, not the full specification. I would reference the engine API spec instead for the details.

glossary.md
```diff
@@ -0,0 +1,241 @@
...
+still hear some people using this term.
+
+## L1 Attributes Transaction
+[l1-attributes-tx]: /glossary.md#l1-attributes-transaction
+
+A transaction with an Optimistic-Ethereum-specific transaction type, that is
+used to register the L1 block attributes (number, timestamp, ...) on L2.
+
+The L1 attributes for a given L1 block can be read on L2 from the [L1 Attributes
+Precompiled Contract][l1-attr-precompile].
```
protolambda, line 154: I don't believe this precompile is the right approach. And although we can list it as a labeled transaction kind, I would say it's the same type of transaction as a deposit: some EVM execution authenticated by the rollup-node.

glossary.md
```diff
@@ -0,0 +1,241 @@
...
+
+cf. [L1 attributes transaction format](/rollup-node.md#payload-transaction-format)
+(in the section on [payload attributes])
+
+> **TODO** We might want to move this the format spec to the execution engine.
```
protolambda, line 159: If it's just the same as the deposit-tx type, then we don't have to define a new type of transaction.

glossary.md
```diff
@@ -0,0 +1,241 @@
...
+
+## L1 Attributes Precompiled Contract
+[l1-attr-precompile]: /glossary.md#l1-attributes-precompiled-contract
+
+A [precompiled contract][precompile] on L2 that can be used to retrieve the L1
+block attributes of L1 blocks with a given block number or a given block hash.
+
+cf. [L1 Attributes Precompiled Contract Specification](TODO)
+
+> **TODO LINK** L1 attributes precompiled contract spec
```
protolambda, line 169: It looks like a precompile, but we don't need it to be one, and I would prefer not to change geth when we do not have to. In this case we want upgradeability, storage, and updates through transactions. Reusing the EVM smart contracts, state storage, and deposit transactions covers everything we need. No need for a precompile.

glossary.md
```diff
@@ -0,0 +1,241 @@
...
+
+--------------------------------------------------------------------------------
+
+# Execution Engine Concepts
+
+## Execution Engine
+[execution engine]: /glossary.md#execution-engine
+
+The execution engine is responsible for executing transactions in blocks and
+computing the resulting state roots, log roots and block hash.
```
protolambda, line 216: There is no such thing as a "log root". The logs are part of the receipt (encoded in RLP etc.), but not merkleized or hashed individually. The receipts form a MPT. They are combined into a bloom filter though.

rollup-node.md
```diff
@@ -0,0 +1,285 @@
...
+2. Pass the payload attributes to the [execution engine], so that [output block
+   properties][block] may be computed.
+
+While this process is conceptually a pure function from the L1 chain to the L2
+chain, it is in practice incremental. The L2 chain changes whenever new L1
+blocks (and hence new L2 derivation inputs) are added to the L1 chain, or
+whenever the L1 chain [re-organizes][reorg].
+
+Additionally, the rollup node is responsible for [block gossip], i.e.
+transmitting the L2 blocks it derives over a peer-to-peer network.
```
protolambda, line 31: Same as other comment about p2p/transmitting: it is the job of the sequencer to seed the gossip, and unrelated to local derivation at the start, until the node figures out it either derives the same thing it already propagated, or derived something that will reorg away the previous sequencer claim.

rollup-node.md
```diff
@@ -0,0 +1,285 @@
...
+driver, while block gossip is [specified in a separate document][gossip-spec].
+
+[gossip-spec]: TODO
+
+> **TODO LINK** block gossip spec
+
+## L2 Chain Derivation
+[l2-chain-derivation]: #l2-chain-derivation
+
+This section specifies how the [rollup driver] derives one L2 block from every
```
protolambda, line 44: Are we fully committed to doing 1:1 derivation? I thought we would map 1 L1 block to 1 *or more* L2 blocks.

rollup-node.md
```diff
@@ -0,0 +1,285 @@
...
+
+[payload-attr]: #From-L2-derivation-inputs-to-payload-attributes
+[`PayloadAttributesOPV1`]: #From-L2-derivation-inputs-to-payload-attributes
+
+The rollup reads the following [L2 derivation inputs] for each L1 block:
+
+- L1 block attributes
+   - block number
+   - timestamp
+   - basefee
```
protolambda, line 61: Not important now, but something we might want to add here later: creating that "logs root" (something more neat than the receipt root on L1). Merkleizing the L1 log events more efficiently, filtered to only those of the successful receipts, would be a nice thing to do as rollup. If L2 doesn't pay for a log-subscription in advance (the extension we talked about), at least it can efficiently provide a proof later on that something happened on L1.

rollup-node.md
```diff
@@ -0,0 +1,285 @@
...
+
+> **TODO LINK** deposit feed contract specification
+
+From these L2 derivation inputs, the rollup node constructs an expanded version
+of the [Engine API PayloadAttributesV1 object][PayloadAttributesV1]:
+
+[PayloadAttributesV1]: https://github.com/ethereum/execution-apis/blob/main/src/engine/specification.md#payloadattributesv1
+
+```
+PayloadAttributesOPV1: {
```
protolambda, line 86: We should move this into its own block-derivation doc to keep this doc light enough to cover all rollup-node responsibilities.

rollup-node.md
```diff
@@ -0,0 +1,285 @@
...
+which are not accepted by the network anymore. This is not an [EIP-2718] typed
+transaction, but can be distinguished from those by the fact that their
+RLP-encoding's first byte will always be `>= 0xC0`.
+
+[EIP-155]: https://eips.ethereum.org/EIPS/eip-155
+
+#### Payload Transaction Format
+
+The `transactions` array is filled with the deposits, prefixed by the (single)
+[L1 attributes transaction]. The deposits (which must be similarly RLP-encoded)
```
protolambda, line 139: The typed-transaction EIP does not require us to RLP-encode the deposits if we don't want to. And it requires us to prefix it with a type byte. Let's keep the rollup-node doc general, and discuss the deposit format in a separate PR (@maurelian can lead this)

rollup-node.md
```diff
@@ -0,0 +1,285 @@
...
+  types. We specify how to select the transactions below.
+
+[unix type]: https://en.wikipedia.org/wiki/Unix_time
+[merge]: https://ethereum.org/en/eth2/merge/
+[EIP-2718]: https://eips.ethereum.org/EIPS/eip-2718
+
+> For details on RLP encoding, refer to the yellow paper, or look at [this
+> hyperlinked implementation][encode-tx].
+
+[encode-tx]: https://github.com/norswap/nanoeth/blob/cc5d94a349c90627024f3cd629a2d830008fec72/src/com/norswap/nanoeth/transactions/Transaction.java#L84-L130
```
protolambda, line 120: EIP-2718 defines all we need, we should not leak the details of other transaction types into this specification, or we will have to keep updating it with every L1 change. Also, like the other review comment, we should not rely on external implementations in this specification.

rollup-node.md
```diff
@@ -0,0 +1,285 @@
...
+- [EIP-1559]: the current "regular" Ethereum L1 transaction type, which
+  separates gas fees between a base fee and a tip
+- the *[L1 attributes transaction]*, see below
+
+[EIP-1559]: https://eips.ethereum.org/EIPS/eip-1559
+
+Historical Optimistic Ethereum transactions also include [EIP-155] transactions,
+which are not accepted by the network anymore. This is not an [EIP-2718] typed
+transaction, but can be distinguished from those by the fact that their
+RLP-encoding's first byte will always be `>= 0xC0`.
```
protolambda, line 132: Why? IMHO we should keep supporting it. Or we will have to modify geth to break support, which seems very backwards.

rollup-node.md
```diff
@@ -0,0 +1,285 @@
...
+[unix type]: https://en.wikipedia.org/wiki/Unix_time
+[merge]: https://ethereum.org/en/eth2/merge/
+[EIP-2718]: https://eips.ethereum.org/EIPS/eip-2718
+
+> For details on RLP encoding, refer to the yellow paper, or look at [this
+> hyperlinked implementation][encode-tx].
+
+[encode-tx]: https://github.com/norswap/nanoeth/blob/cc5d94a349c90627024f3cd629a2d830008fec72/src/com/norswap/nanoeth/transactions/Transaction.java#L84-L130
+
+In particular, we allow the following two transaction types:
```
protolambda, line 122: You are forgetting about the transaction type introduced in the Berlin hardfork (a subset of the EIP-1559 spec, supporting access-lists without the base-fee functionality)

rollup-node.md
```diff
@@ -0,0 +1,285 @@
...
+The `transactions` array is filled with the deposits, prefixed by the (single)
+[L1 attributes transaction]. The deposits (which must be similarly RLP-encoded)
+are simply copied byte-for-byte — it is the role of the [execution engine] to
+reject invalidly-formatted transactions.
+
+> **TODO** must offer some precisions on the format of deposits: sender,
+> receivers both in-tx-as-encoded, and on-L2-tx. What about the fees?
+
+The Optimistic Ethereum specific *[L1 attributes transaction]* has the following
+[EIP-2718]-compatible format: `0x7E || [block_number, timestamp, basefee]`
```
protolambda, line 147: Wasn't @maurelian going to specify the deposits? We should separate this out, and keep it in the deposits specification doc.

glossary.md
```diff
@@ -0,0 +1,241 @@
...
+[deriv-inputs]: /glossary.md#l2-chain-derivation-inputs
+
+This term refers to data that is found in L1 blocks and is read by the [rollup
+node] to construct [payload attributes].
+
+Chain derivation attributes include:
+- L1 block attributes
+   - block number
+   - timestamp
+   - basefee
```
tynes, line 125: Perhaps premature optimization, but I could imagine this also having the value of the `RANDOM` opcode post merge, see https://eips.ethereum.org/EIPS/eip-4399

rollup-node.md
```diff
@@ -0,0 +1,285 @@
...
+  the future. We don't pick `0x7F` itself in case it becomes used for a
+  variable-length encoding scheme.
+- `block_number` is the L1 block number as a 64-bit integer (4 bytes)
+- `timestamp` is the L1 block timestamp as a 64-bit integer (4 bytes)
+- `basefee` is the L1 block basefee as a 64-bit integer (4 bytes)
+
+When included in the `transactions` array, this transaction should be
+RLP-encoded in the same way as other transactions.
+
+Here is an example valid `PayloadAttributesOPV1` object, which contains an L1
```
protolambda, line 162: Let's move this into a doc specific to the execution-engine?

rollup-node.md
tynes, line 120: Perhaps link to geth's RLP package or ethers as well since its widely used?

rollup-node.md
```diff
@@ -0,0 +1,285 @@
...
+Optimistic Ethereum's execution engine API is built upon [Ethereum's Engine API
+specification][eth-engine-api], with a couple of modifications. That
+specification builds upon [Ethereum's JSON-RPC API specification][JSON-RPC-API],
+which itself builds upon the [JSON-RPC specification][JSON-RPC].
+
+[eth-engine-api]: https://github.com/ethereum/execution-apis/blob/main/src/engine/specification.md
+[JSON-RPC]: https://www.jsonrpc.org/specification
+
+In particular, the [Ethereum's Engine API specification][eth-engine-api]
+specifies a [JSON-RPC] endpoint with a number of JSON-RPC (REST) routes, which
```
protolambda, line 198:
```suggestion
specifies a [JSON-RPC] endpoint with a number of JSON-RPC routes, which
```
JSON-RPC is not REST. We kept the engine API design simple enough that it could be, but it's still JSON-RPC.

rollup-node.md
tynes, line 122: Is it intentional to block type 1 transactions? The ones with the accesslist that aren't eip1559

glossary.md
```diff
@@ -0,0 +1,241 @@
...
+
+A re-organization, or re-org for short, is whenever the head of a blockchain
+(its last block) changes (as dictated by the [fork choice rule]) to a block that
+is not a child of the previous head.
+
+L1 re-orgs can happen because of network conditions or attacks. L2 re-orgs are a
+consequence of L1 re-orgs, mediated via [L2 chain derivation][derivation].
+
+## Precompiled Contract ("Precompile")
+[precompile]: /glossary.md#precompiled-contract-precompile
```
tynes, line 61: Precompile vs predeploy? Does predeploy need to be specified as well?

rollup-node.md
```diff
@@ -0,0 +1,285 @@
...
+are the means through which the rollup driver interacts with the execution
+engine.
+
+Instead of calling [`engine_forkchoiceUpdatedV1`], the rollup driver must call
```
protolambda, line 202: I'm not sure if the two should be different. They can be made to be compatible:
- if `transactions` is empty, do the normal thing
- if `transactions` is not empty, do the optimism version (i.e. all the same, except pulling transactions into block from this list, instead of the tx pool)

And the `transactions` field itself is optional, defaults to an empty list.

rollup-node.md
```diff
@@ -0,0 +1,285 @@
...
+  [`PayloadAttributesV1`]. The execution engine must include the valid
+  transactions supplied in this object in the block, in the same order as they
+  were supplied, and only those. See the [previous section][payload-attr] for
+  the specification of how the properties must be set.
+
+- we repurpose the [`ForkchoiceStateV1`] structure with the following property
+  semantics:
+  - `headBlockHash`: block hash of the last block of the L2 chain, according to
+    the rollup driver.
+  - `safeBlockHash`: same as `headBlockHash`.
```
protolambda, line 221: This is changing on L1, and just a placeholder semantic, the two are not the same. And in the rollup they are definitely different. Also, this should be covered in the execution engine doc, not here in the general rollup-node doc.

rollup-node.md
```diff
@@ -0,0 +1,285 @@
...
+
+The L1 chain may occasionally re-organize, meaning the head of the L1 chain
+changes to a block that is not the child of the previous head but rather one of
+its "cousins" (i.e. the descendant of an ancestor of the previous head). In
+those case, the rollup driver must:
+
+1. Locate the *common ancestor*, a block that is an ancestor of both the
+   previous and new head.
+2. Isolate the range of L1 blocks \]common ancestor, ..., new head\].
+3. For each such block, call [`engine_forkchoiceUpdatedOPV1`] and
```
protolambda, line 261: This is closer to what manual sync would look like, but in some cases it may be better to first try a forkchoice update, and have the execution engine resolve the sync (like it would do in the L1 case).

rollup-node.md
```diff
@@ -0,0 +1,285 @@
...
+> **TODO** Must enunciate error cases more precisely, as well as state what
+> should be impossible and should cause the node to stop deriving the L2 chain.
+
+## Block Gossip
+
+Another responsability of the rollup node is to transmit the L2 blocks it
+derives from the L1 chain to other L2 nodes via a process called *[block
+gossip]*.
+
+This is specified in [a separate document][gossip-spec].
```
protolambda, line 285: When linking docs we should just use the document title, instead of "separate document" etc.

glossary.md
norswap: How does that relate to the `random` field we need to pass in the [payload attributes](https://github.com/ethereum/execution-apis/blob/main/src/engine/specification.md#payloadattributesv1)?

Reading through, it seems to be the same:

> Beginning with TRANSITION_BLOCK, client software MUST set the value of the mixHash, i.e. the field with the number 13 (0-indexed) in a block header, to the output of the randomness beacon provided by the beacon chain.

That being said, that only means that it does not mean to be included in the special transaction to convey the attributes... it still needs to be read to construct the attributes, so it might be better to mention it here indeed.

If the `RANDOM` opcode become reality, we'll need to think about wether we want it to return something different on every block. In that case, it might be good to include randomness in the L1 attributes in the transaction, and change the `random` value in the payload attributes.

rollup-node.md
norswap: My goal with the link was to specify / explain how to reimplement rather than point to things that can be used. I must say RLP encoding is the thing that I looked up in Geth that confused me the most by far. This is probably being due to me not being super familiar with the language + this part of the code relying heavily on multiple kind of indirections and reflection, but I expect other people to trip on the same things. So for understanding purposes, it doesn't seem ideal.

I'm not strictly against linking libraries that could be used for implementation, but is this something we want to do in general?

rollup-node.md
norswap: I figured it makes for easier spec and easier implementation, and shouldn't be an impediment at the wallet level (since they needed to update for us anyway, after EIP-1559).

That being said, maybe there are some legacy things we do absolutely wish to support or that will byte us in the ass if we don't?

I'm thinking that over the long run Ethereum is going to deprecate the non EIP-1559 transactions anyway (similarly to what is already underway with pre-EIP-155 transactions that are being soft-deprecated by Geth (there might be an EIP?)).

glossary.md
norswap: What's the trade-off between these two things? A pre-compile does necessitate a Geth transaction, but maybe it can be done very cleanly?

While we're at it, what do we expect the precompile/predeploys to be besides this L1 attributes oracle?

glossary.md
```diff
@@ -0,0 +1,241 @@
...
+used in contrast to [layer 1][L1], which refers to the Ethereum blockchain.
+
+## Block
+[block]: /glossary.md#layer-1-L1
+
+Can refer to an [L1] block, or to an [L2] block, which are structured similarly.
+
+A block is a sequential list of transactions, along with a couple of properties
+stored in the *header* of the block. A description of these properties can be
+found in code comments [here][nano-header], or in the [Ethereum yellow paper
```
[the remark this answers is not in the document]
norswap: It's a bit overkill since this is the glossary. The goal wasn't to be specificative necessarily but informative.
But you're right that there is the potential to fall behind, and it's confusing because, well this is a specification.

I'm not too sure what to do in this context. Listing everything like you say is not the goal here. I just wanted to give the reader a quick idea of what's in a header, and that seemed the easiest way to do it. Got any thoughts?

That being said, the problem of precise specification will be present in the execution engine spec. I'm thinking we shoud link to the yellow paper + the front page of the execution-specs repo that has the whole EIP list per hard fork, does that seem alright?

glossary.md
```diff
@@ -0,0 +1,241 @@
...
+does not need to result in the execution of EVM bytecode. Instead, its behaviour
+is defined by specification.
+
+Ethereum L1 precompiles are specified in Appendix E of the [yellow
+paper][yellow].
+
+Optimistic Ethereum exposes all of Ethereum's precompiles, and adds the
+following precompiles:
+
+- [L1 Attributes Precompiled Contract][l1-attr-precompile]
```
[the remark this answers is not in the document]
norswap: cf. the comment that Mark made about this above

A few notes:
- a genesis contract is still part of the spec.
- a pre-deploy requires either a re-genesis or a hard fork. On the flip side, a precompile can retroactively exist since the start, since nobody called it before the new version of the spec. That being said, I suspect we might need a hard fork anyway.
- the L1 attributes are carried by a custom transaction type, which will require special handling.

I think the fact we'll need to modify Geth is a good argument against a precompile, especially since we'll need to store these L1 attributes somewhere (that's the real annoying thing, I figure precompiles can be added fairly easily).

glossary.md
```diff
@@ -0,0 +1,241 @@
...
+
+The rollup node is responsible for [deriving the L2 chain][derivation] from the
+[L2 derivation inputs][deriv-inputs] available on L1. This is done by its
+[rollup driver] component.
+
+The rollup node is also responsible for [block gossip], the process of
+transmitting the L2 blocks it derives over a peer-to-peer network.
+
+- cf. [Rollup Node Specification](/rollup-node.md)
+- cf. [Block Gossip Specification](TODO)
```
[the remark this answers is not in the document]
norswap: Yeah, I was thinking to just drop the whole block gossip part since it doesn't serve any purpose at the deposit-only stage.

glossary.md
```diff
@@ -0,0 +1,241 @@
...
+L1.
+
+## L2 Chain Derivation
+[derivation]:  /glossary.md#L2-chain-derivation
+
+A process that reads [L2 derivation inputs][deriv-inputs] from L1 in order to
+derive the L2 chain.
+
+cf. [L2 Chain Derivation (in Rollup Node
+Specification)](/rollup-node.md#l2-chain-derivation)
```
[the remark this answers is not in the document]
norswap: 80 lines baby! But yeah, we can do 100 or 120. Any thoughts?

And of course we can make tasteful exceptions.

glossary.md
norswap: I guess it depends how exactly deposits end up working (especially with regards to whether they include a signature). If they're signed, then it has to be different.

glossary.md
norswap: For the sake of brevity and concentrating the discourse, could we avoid making the same comment in three different places ;)

glossary.md
norswap: You're right, "receipts root" is really what is meant here.

rollup-node.md
norswap: It's 1-1 derivation because this is the deposits-only stage. It will change later on when we add sequencer blocks.

rollup-node.md
norswap: Could you write up your thoughts on that as an issue with a specific tag? Like `ideas`, `improvements` or something like that?

rollup-node.md
norswap: I don't get it, block derivation is 90% of the content of the present document, so you would move the whole document??

rollup-node.md
norswap: Mmh that is interesting. We should definitely talk about how things work on L1 here.

rollup-node.md
norswap: Yup, they're not the same but for the deposits-only stage we have nothing better to put in there, so I specified away.

I do think it belongs here too, because the rollup-node needs to fill in the arguments to make the call.

rollup-node.md
norswap: I'm just basing myself of the current version of the spec that doesn't have the `transactions` fields.
But if it changes, then I agree we should do it as you say.

rollup-node.md
norswap: Mmh, that is an interesting perspective indeed, I think I agree.

Regarding the impl link, the point is again to be informative, but your point is well-taken. Maybe we should try to mark it more explicitly? I really believe in the value-add here, the encoding scheme is dead simple, but it's hard to piece together from the yellowpaper (and actually the EIPs were most of them are specified & which I don't link here).

rollup-node.md
norswap: Yeah, good point. I was thinking this was somehow simpler but I agree we should just follow L1 Ethereum.

rollup-node.md
norswap: Didn't forget, was an explicit choice, but one that doesn't make a lot of sense, cf. https://github.com/ethereum-optimism/optimistic-specs/pull/43#discussion_r759679488

rollup-node.md
norswap: Yeah, this could be replaced by a link to that spec. Seemed like a simple enough strawman to put in.

rollup-node.md
norswap: What about having the person that writes the spec move it?

rollup-node.md
norswap: Big philosophical debate, goal was just to give a quick idea, but not a hill that I'm willing to die on, will remove.

rollup-node.md
norswap: On further reflection, I think it's a mistake to try to constrain the type of transactions that are allowed here. See proto's comments to that effect: https://github.com/ethereum-optimism/optimistic-specs/pull/43#discussion_r759678629

maurelian (commented):
Very preliminary review.

glossary.md
```diff
@@ -0,0 +1,241 @@
+# Glossary
+
+--------------------------------------------------------------------------------
+
+# General Terms
+
+## Layer 1 (L1)
+[L1]: /glossary.md#layer-1-L1
```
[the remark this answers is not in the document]
norswap: My thinking was that with the absolute link, we could easily copy the links to any other place where we need them. It's a bit cute, no strong feeling about it. I've used relative anchors in the other file.

glossary.md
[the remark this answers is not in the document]
norswap: resolving this, followup here: https://github.com/ethereum-optimism/optimistic-specs/pull/44

glossary.md
norswap: Added a TODO that we might want to revisit this once the L1 deposit feed contract spec is up.

rollup-node.md
norswap: Actually revisiting this, this is wrong, this transaction does not appear at all on the L1 side since it is derived from L1 block attributes. Should probably be specified in the execution engine instead. Needs to be mentionned here too in any case, so duplication is warranted. We can change it later, obviously.

rollup-node.md
norswap: No discussion on this, really agnostic on what it should be. There's still the chance that the upstream spec updates to accomodate our needs.

norswap: Resolving this, this will be a concern for the execution engine.

rollup-node.md
```diff
@@ -0,0 +1,285 @@
...
+processing the whole L1 chain since the [L2 chain inception].
+
+The L1 chain may occasionally re-organize, meaning the head of the L1 chain
+changes to a block that is not the child of the previous head but rather one of
+its "cousins" (i.e. the descendant of an ancestor of the previous head). In
+those case, the rollup driver must:
+
+1. Locate the *common ancestor*, a block that is an ancestor of both the
+   previous and new head.
+2. Isolate the range of L1 blocks \]common ancestor, ..., new head\].
```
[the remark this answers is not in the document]
norswap: You're correct. I'm adding `code` markers. What would you replace it with? I'm not aware of a notation that would be as universal as this one.

glossary.md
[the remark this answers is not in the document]
norswap: discussion here: https://github.com/ethereum-optimism/optimistic-specs/issues/46

rollup-node.md
```diff
@@ -0,0 +1,296 @@
...
+
+The Optimistic Ethereum specific *[L1 attributes transaction]* has the following
+[EIP-2718]-compatible format: `0x7E || [block_number, timestamp, basefee]`
+where:
+
+- `0x7E` is the transaction type identifier. It is selected because transaction
+  type identifiers are currently allowed to go up to `0x7F`. Picking a high
+  identifier minimizes the risk that the identifier will be used by Ethereum in
+  the future. We don't pick `0x7F` itself in case it becomes used for a
+  variable-length encoding scheme.
```
maurelian, line 148: Is this referring to some discussion elsewhere about using 0x7F? 
Perhaps we should choose some address space closer to the middle of the range, which is a less likely schelling point than the beginning or end.

rollup-node.md
```diff
@@ -0,0 +1,296 @@
...
+[Ethereum JSON-RPC API specification][JSON-RPC-API], as this structure will need to
+be sent over JSON-RPC. `array` refers to a JSON array.
+
+[HEX value encoding]: https://eth.wiki/json-rpc/API#hex-value-encoding
+[JSON-RPC-API]: https://eth.wiki/json-rpc/API
+
+The object properties must be set as follows:
+
+- `timestamp` is set to the current [unix time] (number of elapsed seconds since
+  00:00:00 UTC on 1 January 1970), rounded to the closest multiple of 2 seconds.
```
maurelian, line 110:
```suggestion
- `timestamp` is set to the current [unix time], rounded to the closest multiple of 2 seconds.
```

I think this is a case where we can just link out to the reference without adding context.

maurelian: I think its worth also indicating that the timestamps must be unique, ie. no two L2 blocks have the same one.

rollup-node.md
maurelian: I agree that what's currently written here is fine. It's a placeholder for when there is a sequencer and thus `unsafe` is more meaningful.

glossary.md
```diff
@@ -0,0 +1,220 @@
...
+
+L1 re-orgs can happen because of network conditions or attacks. L2 re-orgs are a
+consequence of L1 re-orgs, mediated via [L2 chain derivation][derivation].
+
+## Predeployed Contract ("Predeploy")
+[predeploy]: /glossary.md#predeployed-contract-predeploy
+
+A contract that was deployed on-chain at genesis (i.e. at the start of the chain).
+
+Optimistic Ethereum has the following pre-compiles:
```
ben-chain, line 65:
```suggestion
Optimistic Ethereum has the following predeploys:
```

glossary.md
```diff
@@ -0,0 +1,220 @@
...
+(its last block) changes (as dictated by the [fork choice rule]) to a block that
+is not a child of the previous head.
+
+L1 re-orgs can happen because of network conditions or attacks. L2 re-orgs are a
+consequence of L1 re-orgs, mediated via [L2 chain derivation][derivation].
+
+## Predeployed Contract ("Predeploy")
+[predeploy]: /glossary.md#predeployed-contract-predeploy
+
+A contract that was deployed on-chain at genesis (i.e. at the start of the chain).
```
ben-chain, line 63:
```suggestion
A contract placed in the L2 genesis state (i.e. at the start of the chain).
```

rollup-node.md
```diff
@@ -0,0 +1,296 @@
...
+[l2-chain-derivation]: #l2-chain-derivation
+
+This section specifies how the [rollup driver] derives one L2 block from every
+L1 block. The L2 block will carry the L1 block attributes (as a *[L1 attributes
+transaction]*) well as all L2 transactions deposited by users in the L1 block
+(*[deposits]*).
+
+[L1 attributes transaction]: /glossary.md#l1-attributes-transaction
+
+### From L2 derivation inputs to payload attributes
```
ben-chain, line 54:
```suggestion
### From L1 blocks to L2 payloads
```

Is what I would suggest given top-level comment of this review.

rollup-node.md
```diff
@@ -0,0 +1,296 @@
...
+[payload-attr]: #From-L2-derivation-inputs-to-payload-attributes
+
+[`PayloadAttributesOPV1`]: #From-L2-derivation-inputs-to-payload-attributes
+
+The rollup reads the following [L2 derivation inputs] for each L1 block:
```
ben-chain, line 59:
```suggestion
The L2 chain derivation function will parse the following info from each L1 block:
```

glossary.md
```diff
@@ -0,0 +1,220 @@
...
+
+This term refers to data that is found in L1 blocks and is read by the [rollup
+node] to construct [payload attributes].
+
+Chain derivation attributes include:
+- L1 block attributes
+   - block number
+   - timestamp
+   - basefee
+- [deposits]
```
ben-chain, line 114: I would suggest either deleting these lines (which duplicate info on `rollup-node.md`) due to the confusion mentioned in the top-level comment, or putting something about feeds:

## Feed

This term refers to any array of data which is generated from L1 blocks by the [rollup node] in constructing [payload attributes] during block derivation.  Example feeds might be:
- a feed of L1 block attributes like `BASEFEE` and `TIMESTAMP` (computed from the L1 `blockHeader`)
- a feed of decoded deposit events (e.g. ABI decoded from events within the L1 `block.receiptsRoot`)

ben-chain (commented):
This is looking awesome.  I think my only wish for this PR would be some wording clarification around "derivation" and "derivation input".  At one point, we say:
> This section specifies how the rollup driver derives one L2 block from every L1 block.

I like this, and I think that the general definition that "block derivation is the process of generating new L2 execution inputs from new L1 blockdata" will hold up well, even beyond the deposit-only rollup.  However, from this perspective, I think that "L2 derivation inputs" is a little confusing.

Given the above definition, I think of pure L1 blocks as the real "input" to derivation. The chain of deposits (which will be generated based on events within the L1 blocks, as TBD in `deposit-feed-spec`) is really an intermediate representation that appears *during* the block derivation, as opposed to being an input to the derivation function.

I have usually thought of these representations as the "feeds". In this sense, block derivation becomes a function like this:
`derive_payload(l1_block) -> payload`
And the feeds are an IR when you examine the function:
`derive_payload = process_feed(generate_feed(l1_block))`

Left some suggestions accordingly, LMK what you think :)

norswap:
@ben-chain Somehow can't reply inline to the top review comment, so here goes.

**1. Agreed the L1 blocks are the real inputs**

But "L2 block inputs" has a different, useful meaning, see (2).

**2. The feed abstraction doesn't really work for me**

I'm not a fan because these things don't act like feeds at all. There is a feed of L1 blocks, but within those, it's just a plain list of deposits and some block attributes.

That is the crux of the distinction - the rollup driver acts as a pipeline that ingests a feed of L1 blocks, and outputs a feed of L2 blocks. In the deposits-only version, each L1 block is then processed to yield a L2 block. Part of this processing is extracting the useful information from the L1 block. This information needs a name, which is where I came up with "L2 derivation inputs".

Schematically:
```
for (L1_block in L1):
    inputs = extract_inputs(L1_block)
    L2_block = process(inputs)
    emit L2_block
```

You proposed `derive_payload = process_feed(generate_feed(l1_block))`
which I can rewrite as

```
L1.map(L1_block -> extract_inputs(L1_block))
  .map(inputs -> process(inputs)
```

so essentially `generated_feed = L1_block -> extract_inputs(L1_block)`
and `process_feed = inputs -> process(inputs)`

But I feel a more useful formulation equivalent to my for loop above is:

```
L1.map(L1_block -> process(extract_inputs(L1_block))
```

i.e. that secondary feed of inputs is pure conceptual overhead.

It's much more useful to have a name for what those "inputs" (things within a single block) are, rather than for a feed of them. To be very concrete, it allows me to split up the document in two sections, one that explains `extract_inputs` above ("From L2 derivation inputs to payload attributes") and one that explains `process` ("Building the L2 block with the execution engine").

If you don't like "L2 derivation inputs", we could find another name though. I just think "feed" doesn't really fit what actually happens in the derivation.

rollup-node.md
norswap: No this is all me.

idk, we should do what seems likely. I think EIPs are likely to just pick in sequential orders.

But in the very unlikely case where they'd need a variable-length bytecode encoding, the traditional way to do that is to use the last byte in the range as the variable-length marker.

In any case, whatever we pick will be relevant to the design of future EIPs anyway, so I'm not too afraid, but I feel like this is both a bit safer & cleaner.

rollup-node.md
norswap: Which reference would you link though? We're not copying this from L1 (1)! This is also different from the L1 spec (which doesn't round at 2s), and that spec is hard to link (because it's inside a pdf).

Also even if we reused the spec AND it was reusable, this is literally 10 words, redundancy seems worth it here.

(1) Although maybe we should at this stage - good question! In the context of future stages, do we want deposit timestamp == L1 timestamp or do we want to reflect our potential processing lag in the timestamp (imagining we go with the 1 block / 2s proposal of Karl)?

maurelian:
Metadata (for [internal tracking](https://linear.app/optimism/issue/ENG-1731/rollup-node-spec))
Fixes ENG-1731

norswap:
@protolambda comment: the input is really block + receipts, which can be verifier against the block. Useful to add this precision, do not need to change the high-level language ("derivation of the L2 chain from the L1 chain").

rollup-node.md
norswap: Marking this resolved, feel free to open an issue if you think it warrants further discussion.

rollup-node.md
norswap: "L2 derivation inputs" are no more!

PR #44, opened 2021-12-01
Add markdown linting and formatting with markdownlint

maurelian:
**Description**

Let's outsource style debates to a linter!

This PR adds [markdownlint](https://github.com/DavidAnson/markdownlint) and [markdownlint-cli2](https://github.com/DavidAnson/markdownlint-cli2), which looks like a very mature and well maintained stack for linting markdown. 

I've gone with the default settings for now, we can modify the [rules](https://github.com/DavidAnson/markdownlint/blob/main/doc/Rules.md) as we see fit in the future.


norswap:
Let's set the line length to something > 80 (the default) since that was literally the only markdown complain in the other PR :) Both 100 or 120 seem like sensible values to me.

Also what does that github workflow do? My naive reading says it lints when pushing to main, but not on PRs, which does not seem extremely useful? Could we make it so that github runs `lint:check` on PRs and warns if the PR is not properly linted?

protolambda (approved):
LGTM. Let's just try it, we can adjust the linting rules later if we see the need to

Edit: I see it defaults to 80 line length. Will update rules there before merge. I like 120.

And I don't think it will trigger on pull-request before it being merged on the target branch, we'll experiment in the next PRs.

Issue #48, opened 2021-12-02
Figure out fees


norswap:
1) We will need to implement EIP-1559 on the rollup. Do we need to modify it or tweak it? In particular how do deposits figure in the computation? This will also depend on block timings (see https://github.com/ethereum-optimism/optimistic-specs/discussions/42)

2) https://github.com/ethereum-optimism/optimistic-specs/discussions/32 : do depositors pay for L2 gas on L1 (simple, but means there are two markets for L2 gas with different prices), or do they pay for L2 gas on L2 (meaning the account must be provisionned with ETH on L2).

norswap:
Regarding (1), if we assume that we are going with the constant block time where the deposit block takes one slot, then it makes sense to count deposits in the EIP-1559 computation.

Regarding (2), I'm personally very much in favour of the second option. Considering the extra costs of submitting the transaction on L1 in the first place, it's unlikely to ever be more profitable to submit on L1 than L2. And if that were to happen anyway, I'm pretty okay with it - it isn't a terrible outcome for anyone.

tynes:
Can we deprecate the L2 `GasPriceOracle`? If we implement a consensus level EIP1559 like mechanism, then we can remove the concept of the L2 gas price being held in the `GasPriceOracle` as well as the offchain `gas-oracle` that pushes the L2 gas price to the chain. If we deposit the L1 basefee into L2, then we can eliminate the L1 gas price being held in the `GasPriceOracle`. The `gas-oracle` also pushes the L1 `basefee` into the `GasPriceOracle` as the L1 gas price. The total fee is computed as L1 fee + L2 fee where the L1 fee is based on the size of the transaction and the L2 fee is based on the amount of execution that happens on L2.

norswap:
@tynes Two followup questions:

1. Do we need to have the L2 gas price on the L1 chain?

2. I think the issue of L1 fee + L2 fee for L2-native transactions is another concern we need to figure out that I didn't touch on above. We just had some slack conversation about this. It seems like you're suggesting we modify EIP-1559 on L2 to account for the L1 basefee, and then get rid of the L1 fee, right?
    - There is a lot of concern about breaking people's gas estimators. Is this something we're willing to push on?
    - Currently, the L1 fee on L2 is not paid via gas at all, it is directly deducted from the user's OETH balance. This works well, but it can be sneaky when using tooling that is not aware of the L1 fee, because it won't be shown. I ran into this issue myself.

[...]

PR #54, opened 2021-12-07
Deposits spec

### LINKED ISSUE #5: What is the deposit specification?

maurelian:
This is as the title says a first pass at the Deposits spec. 

Some remaining open items (along with much discussion and editing of course)
- [ ] Rebase onto #43, update the glossary, and add links.
- [ ] Consider moving relevant content over from the rollup-node and execution-engine docs.
- [ ] Fix line widths and other linting issues

Closes #5


deposit-transactions.md
```diff
@@ -0,0 +1,165 @@
...
+## Deposit Transaction Type
+
+We define a new [EIP-2718] compatible transaction type with the prefix `0x7E`, and the following
+fields:
+
+<!-- ToDo: set to more GoLang like type defs? -->
+- `address to`
+- `address from`
+- `uint256 value`
+- `bytes data`
```
maurelian, line 25: We agreed to defer on how to handle fees. 
What are the implications of excluding the nonce though? Without a sequencer none I suppose, but worth discussion in the future.

deposit-transactions.md
```diff
@@ -0,0 +1,165 @@
...
+  function depositTransaction(
+    address to,
+    uint256 value,
+    bytes memory _data
+  ) external {
+    address from;
+    if (msg.sender == tx.origin) {
+        from = msg.sender;
+    } else {
+        from = msg.sender + 0x1111000000000000000000000000000000001111;
```
maurelian, line 146: Implemented [here](https://github.com/ethereum-optimism/optimism/blob/02c23083d2ec29b0a82755653735f12a839ee827/packages/contracts/contracts/standards/AddressAliasHelper.sol#L21).

deposit-transactions.md
```diff
@@ -0,0 +1,165 @@
...
+    address indexed to,
+    uint256 value,
+    bytes _data
+  );
+
+  event TransactionDeposited(
+    address indexed from,
+    uint256 value,
+    bytes _data
+  );
```
maurelian, line 135: I created two events because felt it was important to differentiate between `address(0)` and `null`. However on reflection, the impl below makes it impossible to emit the first event with `to == address(0)`, so it doesn't really matter. 
Questions: 
Should it be possible to send to `address(0)` on L2? 
Should we consider adding an `isCreation` boolean argument?

deposit-transactions.md
```diff
@@ -0,0 +1,165 @@
...
+the `TransactionDeposited` event(s) emitted by the Deposit Feed contract.
+
+The Deposit Feed handles two special cases:
+
+1. A `to` value of `0`, which results in a contract creation on L2.
+2. A call from a contract account, in which case the `from` value is transformed to its L2 alias.
+   This prevents attacks in which a contract on L1 has the same address as a contract on L2 but
+   doesn't have the same code. We can safely ignore this for EOAs because they're guaranteed to have
+   the same "code" (i.e. no code at all). This also makes it possible for users to interact with
+   contracts on L2 even when the Sequencer is down.
```
protolambda, line 118: To clarify: we don't need to know about L2 state during block derivation to determine if the address is a contract or not, since it can be checked using the L1 during deposit-time instead?

deposit-transactions.md
protolambda: > Should it be possible to send to address(0) on L2?

Yes, although it's weird. Adding a byte for bitflags like `isCreation` and possible future toggles does not sound so bad to me.

protolambda (commented):
Looks good! The `to` case for contract creation needs a bit tidy-up (double event, bitflag/boolean, or zeroed address are all options) and you can remove the draft status.

deposit-transactions.md
```diff
@@ -0,0 +1,165 @@
...
+This transaction type contains a subset of the fields used in [EIP-155], but does not include
+signature information.
+
+[EIP-155]: https://eips.ethereum.org/EIPS/eip-155
+
+Although in practice we define only one new Transaction Type we can distinguish between two distinct
+transactions which occur in the deposit block, based on their positioning. The first transaction
+MUST be the [L1 Attributes Deposit Transaction][l1-attributes-deposit-transaction], followed by a
+dynamic array of [Deposited Transactions][deposited-transactions] submitted to the Deposit Feed
+contract by accounts on L1.
```
norswap, line 32: - I would name the new transaction type, so we can refer to it by name here (and later?)
- There is no notion of "deposit block" at this stage of the spec.
- Are we calling them "deposited transactions" or "deposits"?
- Also shows that this term should have been defined earlier since it's being used here :)
- I'd say "zero or more" instead of a "a dynamic array of" which seems to imply things on storage format.

deposit-transactions.md
```diff
@@ -0,0 +1,165 @@
...
+
+## L1 Attributes Deposit Transaction
+
+[l1-attributes-deposit-transaction]: #l1-attributes-deposit-transaction
+
+This transaction is a call to the [Layer 1 Attributes Predeploy][l1-attributes-predeploy] contract.
```
norswap, line 38: I would mention it's deposit-type transaction. I mean it's sort of implicit in the name, but it's easy to get confused, for a brief second, I thought this was speccing the L1 call to submit a deposit.

deposit-transactions.md
```diff
@@ -0,0 +1,165 @@
...
+
+This transaction MUST have the following values:
+
+1. `from` is the Depositor Account `0xdeaddeaddeaddeaddeaddeaddeaddeaddead0001`.
```
norswap, line 42: [really pedantic comment] Below this is "L1 Attributes Deposit Account" - makes it clear it's not used for anything else. Maybe add a link to section below.

deposit-transactions.md
```diff
@@ -0,0 +1,165 @@
...
+`0xdeaddeaddeaddeaddeaddeaddeaddeaddead0001`. Its value returned by the `CALLER` and `ORIGIN`
+opcodes during execution of the L1 Attributes Deposit Transaction.
+
+### L1 Attributes Predeploy
+
+[l1-attributes-predeploy]: #l1-attributes-predeploy
+
+A predeployed contract on L2 at address `0x4200000000000000000000000000000000000014`, which holds
+certain block variables from the corresponding L1 block in storage, so that they may be accessed
+during the execution of the subsequent deposited transactions.
```
norswap, line 70: Would suggest:

> which maps L1 block numbers to some of the attributes from that block, so that ...

deposit-transactions.md
maurelian: Ah, good call, will include that.

deposit-transactions.md
```diff
@@ -0,0 +1,165 @@
...
+
+The contract implements an authorization scheme, such that it only accepts state-changing calls from
+the [Depositor Account].
+
+The contract has the following solidity interface, and can be interacted with according to the
```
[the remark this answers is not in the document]
maurelian: I guess we could also save 4 bytes by just using the fallback function.
Might be worth doing now, or maybe save optimizations for the next phase.

deposit-transactions.md
```diff
@@ -0,0 +1,165 @@
...
+### Deposit Feed Contract
+
+[deposit-feed-contract]: #deposit-feed-contract
+
+The Deposit Feed contract is deployed to L1. Deposited Transactions are derived from the values in
+the `TransactionDeposited` event(s) emitted by the Deposit Feed contract.
+
+The Deposit Feed handles two special cases:
+
+1. A `to` value of `0`, which results in a contract creation on L2.
```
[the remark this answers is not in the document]
maurelian: That was my thought, I will make it explicit.

deposit-transactions.md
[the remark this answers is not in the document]
norswap: Noob question: why would there be a dynamic length argument for that call?

Quick calculation, 32 bytes of savings per L1 block means saving 7200*32*365 bytes = 84MB/year ~ 8GB/100 years. Might not be worth bothering.

deposit-transactions.md
maurelian: I think this works just based on L1 state, but I'm going to give this thought and respond properly.

deposit-transactions.md
```diff
@@ -0,0 +1,165 @@
...
+  function l1Hash() view;
+}
+```
+
+## Deposited Transactions
+
+[deposited-transactions]: #deposited-transactions
+
+"Deposited Transactions" are derived from logs emitted by the [Deposit Feed
+contract][deposit-feed-contract] on L1.
```
norswap, line 102: Maybe explain that they are meant to be perfectly "normal" L2 transaction behaviour-wise, except they are submitted on L1 -- if not done before? Could steal the explanation from rollup node spec.

deposit-transactions.md
```diff
@@ -0,0 +1,165 @@
...
+
+### Deposit Feed Contract
```
norswap, line 104: What about calling it "Deposit Contract" - I feel like as though no info would be lost?

Deposits don't really behave like a feed, because they get "chunked" into L1 blocks for the purpose of assigning them to an epoch (at a later specification stage).

deposit-transactions.md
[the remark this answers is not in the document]
norswap: This is an important distinction, on L1 an **empty** address (RLP-byte-sequence of zero bytes) is used for contract creation, not the 0 address (20 zeroes).
cf. https://github.com/norswap/nanoeth/blob/cc5d94a349c90627024f3cd629a2d830008fec72/src/com/norswap/nanoeth/data/Address.java#L20-L37

deposit-transactions.md
norswap: The assumption is that if an address is a contract on L1 it won't be a user on L2, and vice-versa (impossible in general to find collision etc). The only clash possible is L1 and L2 contracts with same address because of `CREATE2` with same constructor bytecode deploying two different contract bytecode. So yeah, we don't need L2 state.

norswap: Explain what the L2 alias is here. Perhaps mention L2 contracts can nevertheless check for specific L1 and L2 contract callers.

deposit-transactions.md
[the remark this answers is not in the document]
protolambda: Nice trick with the fallback, but yes, let's optimize in a later phase.

deposit-transactions.md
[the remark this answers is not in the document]
protolambda: If we're doing the flag thing, we can just ignore the to-address bytes in the consumer side (result is the same)

deposit-transactions.md
norswap: Seems quite important to allow `address(0)` since it's used for burning tokens a lot. No reason that shouldn't be allowed for deposited transactions.

I think having a single event with the boolean flag will be simpler to handle in the rollup node, no?

norswap (commented):
I really like the structure of this 👍

deposit-transactions.md
```diff
@@ -0,0 +1,165 @@
...
+the [Depositor Account].
+
+The contract has the following solidity interface, and can be interacted with according to the
+[contract ABI specification][ABI].
+
+[ABI]: https://docs.soliditylang.org/en/v0.8.10/abi-spec.html
+
+```solidity
+interface L1BlockValues {
+
```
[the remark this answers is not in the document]
protolambda: Good call, indeed! In a production setting we should probably make the deposit contract upgrade-able with a delay, to be able to add/change future L1 properties. For the deposit-rollup it's ok to go without.

deposit-transactions.md
tynes: We currently do not support L1 -> L2 contract creations. In the case of the sequencer going down, it is nice functionality to have. Is it worth the extra complexity from a user perspective when a contract factory on L2 could be used to do a L1 -> L2 deployment?

deposit-transactions.md
```diff
@@ -0,0 +1,165 @@
+# Deposit Transactions
+
+Although the L2 execution engine is nearly identical to that of L1, an important difference is the
+introduction of a new 'Deposit' Transaction Type. Deposit transactions have the following notable
+distinctions from existing transaction types:
+
+1. They are initiated by the System as part of the protocol.
+2. Do not require signature validation.
```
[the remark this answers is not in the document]
maurelian: Agree, auth is the primary gap in the current content.

deposit-transactions.md
maurelian: I know how it works currently. :)  As of now there is no way to make a call on L1 and have a `CREATE` on L2. 
The relevant code path is through the L1 and L2 Cross Domain Messenger contracts, and neither contains a `CREATE` op.
So this is one area in which the new design already makes things better, with much less complexity.

deposit-transactions.md
maurelian: I'll just go with the more verbose approach for now. 
I think we should find a shorter name for this in the future though. :)

deposit-transactions.md
norswap: Apologies, I forget not everyone here is new 😂 
What (in the current version of Optimistic Ethereum) prevents a deposit from doing a L2 contract call which calls `CREATE`?
Are deposits purely separate from the cross domain messengers? Also for tokens?

deposit-transactions.md
norswap: It's a good question.
- Option A: add an extra field in the even and an extra argument in the function (unless there's a way to detect the nul address from the argument?).  We'll alsoneed to check the field in the new transaction type handler and call contract creation logic. And finally, increment nonces on the basis of deposits too.
- Option B: Users have to call a predeployed contract factory instead.

Option B seems simpler, though a bit less identical to what you'd do on L1. I could be convinced either way.

deposit-transactions.md
ben-chain: Historically we have just only implemented cross-contract calls--there is no fundamental reason that creations won't work, it's just about implementing it. Two notes:
- There needs to be a way to indicate intent to create on L1 within the Deposit Feed contract. Options would be a separate function on L1 and a new event field, or just specifying `0x000` as target to do a CREATE.
- We should specify how a deposit transaction is interpreted as a CREATE by the EE, i.e. if `to == nil` (which would be compatible with either of the options in the previous bullet).

I think it's worth specifying this here.

ben-chain: FWIW I would advocate for adding a `gasLimit` to this type.  I think it's unlikely that we would not have this field -- even if we defer on how it is populated or validated right now, I think it will exist.  It could just be an argument to the deposit feed's function right now.

deposit-transactions.md
```diff
@@ -0,0 +1,165 @@
...
+    bytes memory _data
+  ) external {
+    address from;
+    if (msg.sender == tx.origin) {
+        from = msg.sender;
+    } else {
+        from = msg.sender + 0x1111000000000000000000000000000000001111;
+    }
+
+    if(_to == address(0)) {
```
ben-chain, line 149: I think using addr(0) is *probably* fine, but raising [this](https://dapp.org.uk/reports/optimism.html#org9346693) issue we've seen in the past. I don't think it's the same issue, but there's always risk of confusion where a dev allows an L1 contract to call an arbitrary L2 contract, and a user can trick the contract into CREATEing instead of making a call to `addr(0)`

deposit-transactions.md
```diff
@@ -0,0 +1,165 @@
...
+[l1-attributes-deposit-transaction]: #l1-attributes-deposit-transaction
+
+This transaction is a call to the [Layer 1 Attributes Predeploy][l1-attributes-predeploy] contract.
+
+This transaction MUST have the following values:
+
+1. `from` is the Depositor Account `0xdeaddeaddeaddeaddeaddeaddeaddeaddead0001`.
+1. `to` is `0x4200000000000000000000000000000000000014` (the address of the L1 attributes predeploy
+   contract).
+1. `value` is `0`
```
ben-chain, line 45: Does this imply that moving ETH in and out of the rollup is out-of-scope for deposit-only rollup?

deposit-transactions.md
```diff
@@ -0,0 +1,165 @@
...
+1. `data` is an abi encoded call to the [L1 Attributes Predeploy] contract's `setL1BlockValues()`
+   function with correct values associated with the corresponding L1 block.
+
```
ben-chain, line 48: I would recommend that we implement an "Execution" section here which outlines exactly how the EE handles payloads with these transactions.  Something like:

"Creates an L2 call frame which:
- sets `CALLER` and `ORIGIN` to `depositTx.from`
- sets context.calldata to `depositTx.data`
- sets context.gas to `depositTx.gasLimit`
- handles `depositTx.value` somehow -- TBH do we know exactly how yet?

deposit-transactions.md
maurelian: Good point to consider. Going to resolve this comment and refer to it in the thread above which is closely related.

deposit-transactions.md
maurelian: Reposting another relevant comment made [below](https://github.com/ethereum-optimism/optimistic-specs/pull/54/files#r765064416) by @ben-chain:
> I think using addr(0) is probably fine, but raising this issue we've seen in the past. I don't think it's the same issue, but there's always risk of confusion where a dev allows an L1 contract to call an arbitrary L2 contract, and a user can trick the contract into CREATEing instead of making a call to addr(0)

deposit-transactions.md
maurelian: This only applies to the first deposit block transaction which submits the L1 attributes.

HOWEVER, it is true that we have not yet specified a method for withdrawing ETH. 
That would look a bit weird, because it would require submitting a transaction to the Deposit Feed contract on L1, in order to burn ETH on L2, and then the Feed contract would transfer the value.
But it's also odd to have ETH permanently locked in the Feed. 🤔

maurelian: I'm going to add a TODO for now.

deposit-transactions.md
norswap: @ben-chain Could you expand on why we need this for? (Are there any use cases if we don't charge L2 fees on L2 for deposits?)

deposits.md
```diff
@@ -0,0 +1,225 @@
...
+- `address to`
+- `address from`
+- `uint256 value`
+- `bytes data`
+- `uint256 gasLimit`
+
+This is a subset of the fields used in [EIP-155], but does not include signature information.
+
+We select `0x7E` because transaction type identifiers are currently allowed to go up to `0x7F`.
+Picking a high identifier minimizes the risk that the identifier will be used by Ethereum in the
```
tynes, line 28: I think this should be reworded to maintain the meme that OP is Ethereum. We should also consider opening an EIP for this once its standardized

deposits.md
```diff
@@ -0,0 +1,225 @@
+# Deposits
```
tynes, line 1: Its definitely the cleanest way to handle deposit transactions as a new transaction type but will require a lot of work getting them added to all of the libraries. It seems like we will need to write the serialization in go (hopefully get it into geth, will need to deserialize from the engine RPC), in JS (hopefully get into ethers, required for fetching transactions) and would be nice to get it in Rust (ethers-rs)

deposits.md
```diff
@@ -0,0 +1,225 @@
...
+
+## L1 Attributes Deposit
+
+[l1-attributes-deposit]: #l1-attributes-deposit
+
+This is a deposit sent to the [Layer 1 Attributes Predeploy][l1-attributes-predeploy] contract.
+
+This transaction MUST have the following values:
+
+1. `from` is the L1 Attributes Depositor Account `0xdeaddeaddeaddeaddeaddeaddeaddeaddead0001`.
```
tynes, line 83: If `from` and `to` have to be constant values, why are they serialized in the tx? If I understand correctly, `from` in the tx is the L1 `msg.sender` and `to` is the target contract being called from L1 and these hard coded values are set for the EVM context and the original data in the tx is encoded in the calldata

deposits.md
```diff
@@ -0,0 +1,225 @@
...
+    address from;
+    if (msg.sender == tx.origin) {
+        from = msg.sender;
+    } else {
+        from = msg.sender + 0x1111000000000000000000000000000000001111;
+    }
+
+    if(isCreation && _to != address(0)) {
+        revert('Contract creation deposits must not specify a recipient address.');
+    } else {
```
tynes, line 206: We can get rid of this `else` clause

deposits.md
```diff
@@ -0,0 +1,225 @@
...
+      emit TransactionDeposited(
+        msg.sender,
+        to,
+        msg.value,
+        isCreation,
+        initCode
```
tynes, line 212: `initCode` should be `_data`?

deposits.md
```diff
@@ -0,0 +1,225 @@
...
+        from = msg.sender + 0x1111000000000000000000000000000000001111;
+    }
+
+    if(isCreation && _to != address(0)) {
+        revert('Contract creation deposits must not specify a recipient address.');
+    } else {
+      emit TransactionDeposited(
+        msg.sender,
+        to,
+        msg.value,
```
tynes, line 210: This is missing `gasLimit`

deposits.md
```diff
@@ -0,0 +1,225 @@
...
+  event TransactionDeposited(
+    address indexed from,
+    address indexed to,
+    uint256 value,
+    uint256 gasLimit,
+    bool isCreation,
+    bytes _data
+  );
+
+  function depositTransaction(
```
tynes, line 190: This doesn't specify the dos protection/burn, is that still going to exist?

deposits.md
protolambda: It was brought up earlier, first we concluded it may be better without to keep scope small of deposit rollup, but it makes sense to have indeed.

deposits.md
```diff
@@ -0,0 +1,225 @@
...
+
+First, the balance of the `from` account MUST be increased by the amount of `value`.
+
+Then, the execution environment for a deposit transaction is initialized based on the transaction's
+values, in exactly the same manner as it would be for an EIP-155 transaction.
+
+Specifically, a new EVM call frame targeting the `to` address is created with values initialized as
+follows:
+
+- `CALLER` and `ORIGIN` set to `from`
```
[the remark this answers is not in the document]
norswap: Right it's transformed before the event is emitted. Maybe worth mentioning here.

deposits.md
norswap: Do you have a suggestion on the rewording?

deposits.md
norswap: Yes, I think we've come to the idea that we were not going to do any fees for the deposit rollup (since L2 congestion is L1 congestion as there are no L2 native tx).

I've been suggesting somewhere that we should a line to this effect to this spec.

Relevant strawman proposal (including "congestion fees"  which we get built-in with EIP-1559): https://github.com/ethereum-optimism/optimistic-specs/issues/48#issuecomment-989839408

deposits.md
norswap: It's listed in the transaction type spec, so it needs to be here for consistency. Also without it, it's possible to generate infinite computation for the rollup, which is not acceptable even for the deposit rollup imho.

deposits.md
```diff
@@ -0,0 +1,225 @@
+# Deposits
+
+Deposits are transactions which are initiated on L1, and executed on L2. This document outlines a new
+[Transaction Type][transaction-type] for deposits. It also describes how deposits are initiated on
+L1, along with the authorization and validation conditions on L2.
```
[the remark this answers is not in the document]
maurelian: Not really, I actually always feel a bit confused about which to use, so don't mind the call out. 

But it's also something we can fix in a 'polish up pass'.

deposits.md
```diff
@@ -0,0 +1,225 @@
...
+We define a new [EIP-2718] compatible transaction type with the prefix `0x7E`, and the following
+fields:
+
+- `address to`
+- `address from`
+- `uint256 value`
+- `bytes data`
+- `uint256 gasLimit`
+
+This is a subset of the fields used in [EIP-155], but does not include signature information.
```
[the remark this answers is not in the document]
maurelian: I moved them around to be more consistent with rollup-node.md

deposits.md
maurelian: Good callout @tynes. For now I'll go with this: 
> Picking a high identifier minimizes the risk that the identifier will be used be claimed by another
> transaction type on the L1 chain in the future.... 
And hard agree that we should open an EIP as soon as we're ready.

glossary.md
```diff
@@ -1,6 +1,30 @@
...
+  - [Rollup Driver](#rollup-driver)
+  - [L2 Chain Derivation](#l2-chain-derivation)
+  - [L2 Derivation Inputs](#l2-derivation-inputs)
+  - [Payload Attributes](#payload-attributes)
+  - [L1 Attributes Transaction](#l1-attributes-transaction)
+  - [L1 Attributes Predeployed Contract](#l1-attributes-predeployed-contract)
+  - [Deposits](#deposits)
+  - [Deposit Feed Contract](#deposit-feed-contract)
+- [Execution Engine Concepts](#execution-engine-concepts)
+  - [Execution Engine](#execution-engine)
```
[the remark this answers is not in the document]
maurelian: Yep, auto-gen...

deposits.md
```diff
@@ -0,0 +1,225 @@
...
+### Validation and Authorization of Deposit Transaction Types
+
+[authorization]: #authorization
+
+As noted above, the Deposit Transaction Type does not include a signature for validation. Rather,
+authorization is handled by the [L1 Deposit Feed contract][deposit-feed-contract] and the
+[L2 chain Derivation][derivation] process itself.
+
+If a Deposit Transaction is included which is not derived using the correct derivation algorithm,
+the resulting state transition would be invalid.
```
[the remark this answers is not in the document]
maurelian: OK, added a link from here to the glossary entry.

The exec engine spec has a [TODO](https://github.com/ethereum-optimism/optimistic-specs/blob/e124d332f39a36d5a36050bc65a004f05fdf8c2c/exec-engine.md#L63) to link to the derivation spec. I think that's still a WIP right?
Certainly not in scope for this PR.

deposits.md
```diff
@@ -0,0 +1,225 @@
...
+
+### Execution
+
+In order to execute a deposit transaction:
```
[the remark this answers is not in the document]
maurelian: I could see splitting this document into 'Deposit contract', and 'Deposit transaction type' in a forthcoming polish PR.

deposits.md
```diff
@@ -0,0 +1,225 @@
...
+
+  function l1Number() view;
+  function l1Timestamp() view;
+  function l1BaseFee() view;
+  function l1Hash() view;
+}
+```
+
+## L1 Transaction Deposits
+
```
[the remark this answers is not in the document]
maurelian: deferring for now: https://github.com/ethereum-optimism/optimistic-specs/issues/61

deposits.md
maurelian: coding in markdown is hard.

deposits.md
```diff
@@ -0,0 +1,225 @@
...
+> **TODO** Specify and link to deposit blocks
+
+### Validation and Authorization of Deposit Transaction Types
+
+[authorization]: #authorization
+
+As noted above, the Deposit Transaction Type does not include a signature for validation. Rather,
+authorization is handled by the [L1 Deposit Feed contract][deposit-feed-contract] and the
+[L2 chain Derivation][derivation] process itself.
```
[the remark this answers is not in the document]
maurelian: How about this? 
```
authorization is handled by the [L2 chain Derivation][derivation] process, which when 
correctly processed will only derive transactions with a `from` address attested to 
by the logs of the [L1 deposit feed contract][deposit-feed-contract].
```

deposits.md
norswap: This is the L1 attributes deposit transaction, which sends L1 header values to a L2 contract when they can subsequently be read by transsactions in the block. This transacton is always the very first in a L2 block.

Things are hardcoded because the L1 attributes deposit transaction has the same EIP-2718 type as user deposit transactions. It's just generated automatically from L1 blocks instead of being submitted by a user.

The rationale was that it would be easier to process (smaller code diff) to do it this way rather than introducing another separate tx type.

@maurelian maybe this needs to be introduced a bit better? it's a bit free floating right now, which is understandable because these transactions are not processed by the deposit feed contract (it's debatable whether this should be specified here imho) -- let's consider this for a subsequent PR?

maurelian: I modified some of the earlier content to make the distinction between the two uses of this tx type more explicit, and explain the rationale for only creating a single new type.

https://github.com/ethereum-optimism/optimistic-specs/commit/4729f5ec12e92417314d6727d97cd488e3e59292

PR #56, opened 2021-12-07
add more linting exemptions

norswap:
See the file with the justifications.

Additional note on allowing multiple H1, going too deep into the heading hierarchy makes them stand out less. I like reusing h1 sometimes, as in the glossary: https://github.com/ethereum-optimism/optimistic-specs/blob/main/glossary.md


protolambda:
Just opened a PR to fix the linting on `main` as well: #57 . I'm fine with either.

protolambda:
Although, emphasis shouldn't be a heading, and using one `<h1>` per document is normal.

norswap:
Merging this after clearing with proto. For `<h1>`:  the glossary really looks more cleaner. For empahsis paragraphs, it's just nice to allow us writing things like:

**Beware: only do this if you know what you're doing.**

PR #58, opened 2021-12-07
rollup node: start on ref implementation (deposit-only rollup prototype)

protolambda:
Changes:
- Move spec docs to `/spec`
- Create `/opnode`
- `go.mod` for opnode go module in repo root
- `opnode/main.go`: boilerplate, but fine-grained access to shutdown, graceful exit, dynamic commands, etc.
- `opnode/node/node.go`: initialization of node (starts RPC clients, will start health service/API etc. later), and main loop (go pattern to synchronize core logic, useful to avoid race conditions like starting multiple sync processes with engine)
- `opnode/node/log.go`: see `--help` for logging options (level/color/format)
- `opnode/node/api.go`: Engine API types, bindings, and block -> execution-payload (API type) conversion
- `opnode/node/driver`: incomplete, next up, in separate PR.

Libraries:
- go-ethereum for ethereum types and JSON RPC client, as well as the logger in the rollup-node
- [`protolambda/ask`](https://github.com/protolambda/ask) for CLI interactions, prev. used to for [ZCLI](https://github.com/protolambda/zcli) and other eth2 debug tooling of mine (interactive shells etc.). Can compose any command structure, and dynamically expand etc.

TODO:
- [ ] automate `go fmt` check in CI (probably better in separate PR)
- [ ] start reading L1 blocks from RPC
- (new PRs) expand driver: (will work with @norswap)
  - update deposit transaction code
  - L2 block derivation
  - call engine and complete the derivation outputs to full blocks

```
go mod download
go run ./opnode run --help
```


opnode/node/driver.go
```diff
@@ -0,0 +1,105 @@
...
+}
+
+func (dep *DepositTransaction) Decode(data []byte) (valid bool) {
+	// TODO
+	return false
+}
+
+type OpaqueTransaction []byte
+
+type L2Block struct {
```
karlfloersch, line 40: Is this an "L2BlockInput" ? Or is this a stand-in for the actual L2 blocks?

opnode/node/node.go
```diff
@@ -0,0 +1,97 @@
...
+func (c *OpNodeCmd) RunNode() {
+	c.log.Info("started OpNode")
+
+	heartbeat := time.NewTicker(time.Millisecond * 700)
+	defer heartbeat.Stop()
+
+	for {
+		select {
+		case <-heartbeat.C:
+			// TODO poll data, process blocks, etc.
```
karlfloersch, line 75: This would also include reorging the L2 chain, is that right? 

Also, could consider adding a comment or a placeholder function to explicitly call out the handling of reorgs will be done in the driver. I just want to make sure that as we implement we **do not** put off handling reorgs till the end and instead are thinking about graceful reorg logic from the beginning.

opnode/node/api.go
```diff
@@ -0,0 +1,274 @@
...
+type ExecutePayloadResult struct {
+	// the result of the payload execution
+	Status ExecutePayloadStatus `json:"status"`
+	// the hash of the most recent valid block in the branch defined by payload and its ancestors
+	LatestValidHash Bytes32 `json:"latestValidHash"`
+	// additional details on the result
+	ValidationError string `json:"validationError"`
+}
+
+type ForkchoiceState struct {
```
karlfloersch, line 155: Did not realize 'forkchoice' was one word ! (presumably the lack of camel case is not an error)

opnode/node/api.go
```diff
@@ -0,0 +1,274 @@
...
+
+func (b PayloadID) MarshalText() ([]byte, error) {
+	return hexutil.Bytes(b[:]).MarshalText()
+}
+
+func (b PayloadID) String() string {
+	return hexutil.Encode(b[:])
+}
+
+type ExecutionPayload struct {
```
karlfloersch, line 107: Are these types able to be imported from geth?

opnode/node/driver.go
karlfloersch: If this is a real L2 block, should this just be a normal `Block` type?

karlfloersch (commented):
Left a few comments/questions but overall this is lookin' reallllly good! Extremely solid base IMO

opnode/node/driver.go
protolambda: It's just the input, the contents a sequencer posts on-chain. I'll remove it for now while we're only doing the deposit-rollup. I'm curious though, how do we determine the coinbase address? Do we encode it as part of the transactions batch, or do we maybe use the sender of the transaction that made the data available?

protolambda: Probably better to discuss in an issue later. Will focus on deposit-rollup for now.

opnode/node/node.go
protolambda: I'll implement collection of L1 data in a later PR to keep the size of this one reasonable. But yes, definitely want to handle reorgs. And subscribing, instead polling, would be neat too. I already put a TODO for it a few lines down. :+1: 

> // TODO: open a channel with L1 RPC to listen for new blocks and reorgs?

opnode/node/api.go
protolambda: See https://github.com/ethereum/execution-apis/blob/main/src/engine/specification.md#forkchoicestatev1

Engine-API spec regards it as one word. But the consensus specs fork choice doc uses both "fork choice" and "fork-choice": https://github.com/ethereum/consensus-specs/blob/dev/specs/merge/fork-choice.md And other places mix it up as well.

I'll stick with the API naming here.

opnode/node/api.go
protolambda: Some of the API types are, but I was not happy with all of the types.

- To match the quantity naming in the spec, I aliased: `type Uint64Quantity = hexutil.Uint64`
- To properly enforce a maximum of 32 bytes as per the spec, I added `BytesMax32`
- To enforce fixed-size 256 byte logs-bloom, I added `Bytes256`
- To not deal with untyped payload IDs or byte slices, I created a type for it: `PayloadID`
- For hashes and addresses we just use the geth types (`common.Hash`, common.Address`)

TLDR: more strictly typed and closer to spec in a few cases.

maurelian:
Is this going to be moved [here](https://github.com/ethereum-optimism/reference-optimistic-geth)?

karlfloersch (approved):
I think this is a great start -- good enough for a merge. I did double check that it works & definitely the code is nice scaffolding for future work.

maurelian:
**High level comments:** 

1. I need more info in order to be able to run this, for both of the nodes required (`--l1` and `--l2`):
    1. What repo and git ref should I build it from, or otherwise how should I install it?
    2. What command and config should I use to run it?

2. I think this needs some polish (at least more commenting) to get to the level of readability that @snario has encouraged for a 'reference impl'. However I also recognize the need to **prototype and not die a death by nits in PR review**. To that end I think we should create a `staging` branch, where this could be merged on the basis of "it runs, and the code looks reasonable". Then we could give it more scrutiny to ensure:
	1. consistency with the existing spec content
	2. Readability, clarity, etc.

karlfloersch:
> I think we should create a `staging` branch

Yep I agree with this suggestion so that we can iterate quickly but also keep `main` super refined / detail oriented

maurelian:
Great! 
Still would like to run it before it's merged tho. :)

opnode/node/api.go
```diff
@@ -0,0 +1,274 @@
...
+			code := ErrorCode(rpcErr.ErrorCode())
+			e.Warn("Unexpected error code in forkchoice-updated response", "code", code)
+		} else {
+			e.Error("Failed to share forkchoice-updated signal")
+		}
+		return result, err
+	}
+}
+
+func BlockToPayload(bl *types.Block, random Bytes32) (*ExecutionPayload, error) {
```
norswap, line 240: Do we expect to ever go from block to payload?

(That being said, not opposed to have this here in any case, it's pretty instructive and seems like a useful utility to have.)

opnode/node/driver.go
```diff
@@ -0,0 +1,87 @@
...
+}
+
+func (item *InputItem) Epoch() uint64 {
+	return item.Block.NumberU64()
+}
+
+// sanity check that the receipts are consistent with the block data
+func (item *InputItem) CheckReceipts() bool {
+	hasher := trie.NewStackTrie(nil)
+	computed := types.DeriveSha(types.Receipts(item.Receipts), hasher)
```
norswap, line 22: the `types.Receipts(...)` performs an array copy, right?

opnode/node/driver.go
```diff
@@ -0,0 +1,87 @@
...
+
+func (dep *DepositTransaction) Decode(data []byte) (valid bool) {
+	// TODO
+	return false
+}
+
+type OpaqueTransaction []byte
+
+type OutputItem struct {
+	// TODO: convert to L1 attributes deposit
```
norswap, line 43: Since we made the L1 attributes transaction a deposit-type transaction, shouldn't we bundle this along with `UserDepositTxs` (probably under a different name)?

opnode/node/driver.go
```diff
@@ -0,0 +1,87 @@
...
+
+	// To be converted to full block with Engine API
+	UserDepositTxs []*DepositTransaction
+}
+
+var DepositGatewayAddr = common.HexToAddress("0xaaaaaaaaaTODO")
+
+// TODO: maybe submit batches to a real deployed contract, if we can avoid the EVM running?
+// To save the ETH and other things that get accidentally get here.
+var BatchSubmissionAddr = common.HexToAddress("0x0000000000000000000000006f7074696d69736d") // "optimism"
```
norswap, line 55: Not a bad idea, can you open an issue about this?
A counterpoint is that it not being possible for us to refund mistakes might be an advantage, as that absolves us from having to deal with that at all.

opnode/node/driver.go
```diff
@@ -0,0 +1,87 @@
...
+type OutputItem struct {
+	// TODO: convert to L1 attributes deposit
+	L1BlockHash   common.Hash
+	L1BlockNumber uint64
+
+	// To be converted to full block with Engine API
+	UserDepositTxs []*DepositTransaction
+}
+
+var DepositGatewayAddr = common.HexToAddress("0xaaaaaaaaaTODO")
```
norswap, line 51: What if we changed this to `DepositContractAddr`?

opnode/node/driver.go
```diff
@@ -0,0 +1,87 @@
...
+				// TODO: logs from other addresses may still be converted to other types of deposit txs,
+				// if we implement payment for this type of L1 -> L2 subscription service.
+			}
+		}
+	}
+
+	for _, tx := range input.Block.Transactions() {
+		if to := tx.To(); to != nil && *to == BatchSubmissionAddr {
+			// TODO: parse L2 block data from transaction
+		}
```
norswap, line 84: I think it's fine to have this stub, but it should be flagged explicitly as being out-of-sync/not-in-scope with current state of the spec. e,g. something like `// future stub: not relevant for deposit rollup`

norswap (approved):

opnode/node/api.go
protolambda: If we distribute block data in RLP encoded form, and want to insert it through the API, then it's useful to have. We can remove it later if we do not end up doing that. It's there for completeness of the API bindings.

opnode/node/driver.go
protolambda: No it does not. Go distinguishes between `slice` and `array`. This is just a slice: the type contains the element type, the memory is a pointer to the start, the slice length (element count, not memory size), and capacity (element count of underlying array that backs the slice). The memory of the slice-header is copied, but is only 24 bytes. The type is cast and doesn't cost us anything. This cast works because the element type is the same, and it's only the slice type that changes.

opnode/node/driver.go
protolambda: I kept them separate so we wouldn't have to define how it translates from typed properties to opaque encoded deposit transaction. In a follow-up PR we can implement actual block-derivation. And maybe we can do it in steps: decode into typed data first, then encode into a list of opaque transactions, and then complete the full block contents with execution through engine-api.

opnode/node/driver.go
protolambda: I think there are contracts named like ETH and token gateway, but it's really just a placeholder. Let's address naming in the block-derivation work.

opnode/node/driver.go
protolambda: Opened, see #62

opnode/node/api.go
norswap: I think it would be useful to have a comment that says exactly this in the code!

PR #66, opened 2021-12-13
Add Deposit Feed contract

maurelian:
**Description**

This PR does the following:

1. Adds a new contracts sub-package.
2. The Deposit Feed contract and tests.
2. Removes the pseudocode from deposits.md, and just points to the solidity file instead.

**Note to reviewers**

Although this PR adds 17 new files, most are just config and scaffolding. You can focus on: 
1. DepositFeed.sol
2. depositFeed.spec.ts
3. Minor changes to deposits.md

**Metadata**
- Fixes #ENG-1738
- #68


deposits.md
```diff
@@ -189,53 +189,10 @@ The deposit feed handles two special cases:
 
 1. A contract creation deposit, which is indicated by setting the `isCreation` flag to `true`.
    In the event that the `to` address is non-zero, the contract will revert.
-2. A call from a contract account, in which case the `from` value is transformed to its L2 .
+2. A call from a contract account, in which case the `from` value is transformed to its L2 alias.
```
norswap, line 192: I'd explain what this means inline or provide a link.

packages/contracts/contracts/DepositFeed.sol
```diff
@@ -0,0 +1,51 @@
...
+        address indexed to,
+        uint256 value,
+        uint256 gasLimit,
+        bool isCreation,
+        bytes data
+    );
+
+    /**
+     * Accepts deposits of ETH and data, and emits a TransactionDeposited event for use in deriving
+     * deposit transactions.
```
norswap, line 25: The "ETH deposit" aspect of it is something that isn't really explained in the spec, is it?

This makes the deposit contract acts like a bridge for ETH (i.e. locking up the ETH) so that it can be minted on L2 (something we haven't specced either).

I have a few concerns with this:
1. This makes ETH deposits special instead of using the generic mechanism (i.e. use L1 to deposit in a locking contract, emit a minting transaction deposit to the deposit contract).
2. There is currently no way to reclaim ETH from this contract, which means that the ETH we'd mint on L2 would be "unbacked" (i.e. they'd be more OETH than ETH that can be reclaimed from our bridge).

packages/contracts/test/DepositFeed.ts
```diff
@@ -0,0 +1,155 @@
+import { expect } from 'chai'
+import { ethers } from 'hardhat'
+import { Contract, Signer, BigNumber } from 'ethers'
+import { applyL1ToL2Alias } from '@eth-optimism/core-utils'
+
```
norswap, line 5: What happened to the idea of using Foundry for testing?

packages/contracts/test/DepositFeed.ts
```diff
@@ -0,0 +1,155 @@
...
+        true,
+        '0x'
+      )
+    ).to.be.revertedWith(
+      'Contract creation deposits must not specify a recipient address.'
+    )
+  })
+
+  describe('Should emit the correct log values when:', async () => {
+    it('a simple ETH-only transaction is deposited.', async () => {
```
maurelian, line 62: Proto reminded me that I need to check balances here.

packages/contracts/test/DepositFeed.ts
```diff
@@ -0,0 +1,155 @@
...
+import { applyL1ToL2Alias } from '@eth-optimism/core-utils'
+
+const ZERO_ADDRESS = '0x' + '00'.repeat(20)
+const ZERO_BIGNUMBER = BigNumber.from(0)
+const NON_ZERO_ADDRESS = '0x' + '11'.repeat(20)
+const NON_ZERO_GASLIMIT = BigNumber.from(50_000)
+const NON_ZERO_VALUE = BigNumber.from(100)
+const NON_ZERO_DATA = '0x' + '11'.repeat(42)
+
+const decodeDepositEvent = (
```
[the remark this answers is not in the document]
norswap: Doesn't typechain auto-create nice structure for events? o.o

packages/contracts/test/DepositFeed.ts
```diff
@@ -0,0 +1,155 @@
...
+        from: applyL1ToL2Alias(dummy.address),
+        to: ZERO_ADDRESS,
+        value: ZERO_BIGNUMBER,
+        gasLimit: NON_ZERO_GASLIMIT,
+        isCreation: true,
+        data: NON_ZERO_DATA,
+      })
+    })
+  })
+})
```
norswap, line 155: Do we want to use a solidity code coverage tool?

norswap (commented):
Looks good to me, but I need we need to discuss the ETH deposit aspect further. I didn't really think about it when reviewing the spec!

packages/contracts/test/DepositFeed.ts
maurelian: Sure, I have some experience [setting it up](https://github.com/ethereum-optimism/optimism/commit/46d30311d69af4a7f757bc9ec1863d4bf90442fd). I'd say it's a nice to have though considering the size of the code. Created the issue here: #80

packages/contracts/test/DepositFeed.ts
maurelian: it doesn't easily support testing events, which is 90% of what we need to test.

packages/contracts/test/DepositFeed.ts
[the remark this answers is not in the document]
maurelian: > Is declaring a function as a const holding a lambda how the cool kids do it these days? Is the point to avoid polluting the implicit this?

Yes and yes. :)

> Doesn't typechain auto-create nice structure for events? o.o

Good point, I'm using that now, but the custom function is still helpful to make the data from typechain a bit more usable.

packages/contracts/contracts/DepositFeed.sol
maurelian: 1. Well, ETH is special. But I take your point that we don't want the UX to be too different from other tokens. Ultimately I think we'll probably want to reuse the same bridge contracts we have now, but I do think we might want to. 
2. That's definitely not great, I agree, but I do think it's a necessary simplification right now as withdrawals would explode the design surface quite a bit.

norswap:
A suggestion for testing events, we can use the waffle library, in particular the `emit` Chai (another library) matcher: https://ethereum-waffle.readthedocs.io/en/latest/matchers.html#emitting-events

Otoh, less dependencies is also cool with me.

norswap:
Note: the project refuses to compile if the `ETHERSCAN_API_KEY` env variable is not set, which is not ideal.

norswap:
Trying to run, I get:
```
# yarn test
yarn run v1.22.17
$ hardhat test
An unexpected error occurred:

hardhat.config.ts:28:3 - error TS2322: Type '{ solidity: string; networks: { rinkeby: { url: string; accounts: string[]; }; }; gasReporter: { enabled: boolean; currency: string; }; etherscan: { apiKey: string | undefined; }; }' is not assignable to type 'HardhatUserConfig'.
  Object literal may only specify known properties, and 'etherscan' does not exist in type 'HardhatUserConfig'.

28   etherscan: {
     ~~~~~~~~~~~~
29     apiKey: process.env.ETHERSCAN_API_KEY,
   ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
30   },
   ~~~

error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

maurelian:
@norswap, you need to pull again :)

maurelian:
We had a bit of a side bikeshed about the path forward here WRT to handling deposits with ETH value. 
Here is some historical context as to why we have an ERC20 contract for tracking ETH balances on L2: #84.

I plan to move forward with the following: 
1. Assume that ETH is deposited to an external bridge contract, which we will not implement (though will likely have a dummy contract for testing purposes). 
2. ETH is no longer sent to or held in the Deposit Feed Contract. This means that minting on L2 will be done purely based on an authorization scheme, much like now. That is, the L2 Bridge contract will mint ETH if and only if the L1 caller is the L1 Bridge, informing it that it has received an ETH deposit.

packages/contracts/hardhat.config.ts
```diff
@@ -0,0 +1,33 @@
...
+
+const config: HardhatUserConfig = {
+  solidity: '0.8.10',
+  networks: {
+    rinkeby: {
+      url: process.env.RINKEBY_URL || '',
+      accounts:
+        process.env.PRIVATE_KEY !== undefined ? [process.env.PRIVATE_KEY] : [],
+    },
+  },
```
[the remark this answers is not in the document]
maurelian: by default :)

maurelian:
Contrary to my previous comment, I now think that we _should_ accept ETH directly within the `DepositFeed` contract. 

1. Not accepting ETH directly would require that we add another edge case (`require(msg.sender == address(L1EthBridge))`). 
2. In that edge case, we would need to parse the `calldata`, in order to extract the `value` and `data` emit.
3. The main motivation for NOT accepting ETH directly is backwards compatibility with the existing L1 bridge. If that's our priority, then we should also preserve backwards compatibility the L2 bridge, which would require many more changes:
  - ie. ETH deposits would be minted as the `OVM_ETH` token, by a privileged contract (the L2 bridge).
4. ETH is the native token of the network, and so should be treated like a first class member (alongside `data`). The alternative approach is an ugly abstraction we should try to avoid if we can.

It may be the case that we eventually decided that backwards compatibility IS important enough to do this, but as protocol designers, we should default to simplicity, and let others talk us into adding complexity.

PR #83, opened 2021-12-16
Rollup Driver Implementation

norswap:
Very much a work in progress, but just wanted to checkpoint what I've been up to.


opnode/node/driver.go
```diff
@@ -1,40 +1,79 @@
...
+
 type DepositTransaction struct {
-	Sender common.Address
-	// TODO some flag/bool for contract-deploy transaction. Or it can be *To with nil
-	Flags uint8
-	To    common.Address
-	Value uint256.Int
-	Data  []byte
+	From     common.Address
+	To       AddressOrNil // nil for the empty/nil address (∅)
```
protolambda, line 66: This can just be a one-liner, see upstream go-ethereum implementation of the regular tx type with optional `To` field:
https://github.com/ethereum/go-ethereum/blob/4e599ee469223ccc452f40f36c85bac3bea9c16b/core/types/dynamic_fee_tx.go#L31

opnode/node/driver_test.go
```diff
@@ -0,0 +1,157 @@
...
+
+		// ignored (zeroed):
+		PostState:         []byte{},
+		CumulativeGasUsed: 0,
+		TxHash:            zeroHash,
+		ContractAddress:   zeroAddress,
+		GasUsed:           0,
+		BlockHash:         zeroHash,
+		BlockNumber:       zero,
+		TransactionIndex:  0,
```
protolambda, line 97: If fields are zeroed, you can just omit them completely. (the byte slice will be `nil`, but generally acts the same, and allocates a slice when necessary on first append). No need to specify a `zeroHash` or `zeroAddress`! (And if you really need to be explicit, `common.Address{}` and `common.Hash{}` would be the idiomatic Go way.

opnode/node/driver_test.go
```diff
@@ -0,0 +1,157 @@
...
+		ParentHash: zeroHash,
+		UncleHash:  zeroHash,
+		Coinbase:   zeroAddress,
+		Root:       zeroHash,
+		Difficulty: zero,
+		Number:     zero,
+		GasLimit:   0,
+		MixDigest:  zeroHash,
+		Nonce:      types.EncodeNonce(0),
+		BaseFee:    zero,
```
protolambda, line 120: See other comment about default zero values.

opnode/node/driver_test.go
norswap: Yup, it was my intention to be explicit, in case we needed to tweak some other fields in the generation process.

Good stuff about `common.Address{}`, only realized you could do that with array aliases later!

norswap:
Unfinished, but the part that reconstructs the deposits from log is quite clean now + fully tested with a testing infrastructure that allows us to generate sample blocks/transactions/receipts.

protolambda:
Rebased onto latest end-to-end work (staging + minor improvement on `ChainMissing` case in L1 tracker that passed by in code review). Will pull in latest ref optimistic geth (updated deposit tx type to split `Mint` from `Value` so both transfers of L2 ETH and L1-ETH deposits can work), and then update driver/tests to encode the deposit tx & build full block through API.

protolambda:
Some tests need to be fixed for derivation: I updated test utils, and new test runners to iterate test cases, but the cases themselves are broken because we now use the `types.DepositTx` of opti geth, and since block-derivation is now split into three functions (instead of previous object-heavy approach):
- `DeriveL1InfoDeposit(block *types.Block) *types.DepositTx`
- `DeriveUserDeposits(block *types.Block, receipts []*types.Receipt) ([]*types.Transaction, error)`
- `DerivePayloadAttributes(block *types.Block, receipts []*types.Receipt) (*PayloadAttributes, error)` (calls above two functions, essentially creates the L2 block-inputs)
- `DeriveBlock(ctx context.Context, engine BlockPreparer, fcState *ForkchoiceState, attributes *PayloadAttributes) (*ExecutionPayload, error)` (computes full block from inputs, using only engine API calls, with specified forkchoice context)

(The fault proof code would use the first 3, and replace the last call with a direct computation of the block, instead of calling an external API)

I'd rather merge it into `staging` so we can rebase onto `main` and get consistency with the contracts PR work

opnode/l2/driver.go
```diff
@@ -2,88 +2,203 @@ package l2
...
+// parse log data for:
+//     event TransactionDeposited(
+//    	 address indexed from,
+//    	 address indexed to,
+//    	 uint256 value,
+//    	 uint256 gasLimit,
+//    	 bool isCreation,
+//    	 data data
+//     );
+func UnmarshalLogEvent(blockNum uint64, txIndex uint64, ev *types.Log) (*types.DepositTx, error) {
```
norswap, line 35: A general comment, I don't think it was quite necessary to change my code so much for stylistical reasons. It's alright, and definitely no hard feelings, but some people might not find this cool (and I sometimes like it much less, here I think it's relatively minor so whatever).

opnode/l2/driver.go
```diff
@@ -2,88 +2,203 @@ package l2
...
+	if len(ev.Topics) != 3 {
+		return nil, fmt.Errorf("expected 3 event topics (event identity, indexed from, indexed to)")
+	}
+	if ev.Topics[0] != DepositEventABIHash {
+		return nil, fmt.Errorf("invalid deposit event selector: %s, expected %s", ev.Topics[0], DepositEventABIHash)
+	}
+	if len(ev.Data) < 160 {
+		return nil, fmt.Errorf("deposit event data too small (%d bytes): %x", len(ev.Data), ev.Data)
+	}
```
norswap, line 44: Are all those sanity checks really necessary? What are your trust assumptions regarding the input (might be worth documenting that)?

opnode/l2/driver.go
```diff
@@ -2,88 +2,203 @@ package l2
...
+	dep.From = common.BytesToAddress(ev.Topics[1][12:])
+	// indexed 1
+	to := common.BytesToAddress(ev.Topics[2][12:])
+
+	// unindexed
+	// 0:32: value - big-endian
+	dep.Value = new(big.Int).SetBytes(ev.Data[0:32])
+	// 32:64: gas - big-endian
+	gas := new(big.Int).SetBytes(ev.Data[32:64])
+	if !gas.IsUint64() {
```
norswap, line 61: Where does this restriction come from?

opnode/l2/driver.go
```diff
@@ -2,88 +2,203 @@ package l2
...
+		return nil, fmt.Errorf("bad gas value: %x", ev.Data[32:64])
+	}
+	dep.Gas = gas.Uint64()
+	// 64:96: isCreation - boolean, aligned with end.
+	// If False == 0 then it will create a contract using L2 account nonce to determine the created address.
+	if ev.Data[95] == 0 {
+		dep.To = &to
+	}
+	// 96:128: data offset
```
norswap, line 70: Not sure these comments add anything to the code.

opnode/l2/driver.go
```diff
@@ -2,88 +2,203 @@ package l2
...
+	// unindexed
+	// 0:32: value - big-endian
+	dep.Value = new(big.Int).SetBytes(ev.Data[0:32])
+	// 32:64: gas - big-endian
+	gas := new(big.Int).SetBytes(ev.Data[32:64])
+	if !gas.IsUint64() {
+		return nil, fmt.Errorf("bad gas value: %x", ev.Data[32:64])
+	}
+	dep.Gas = gas.Uint64()
+	// 64:96: isCreation - boolean, aligned with end.
```
norswap, line 65: Maybe "0 or 1, encoded as a big-endian big integer"? "aligned with end" doesn't really do it for me.

opnode/l2/driver.go
```diff
@@ -2,88 +2,203 @@ package l2
...
+	dataOffset.SetBytes(ev.Data[96:128])
+	if dataOffset.Eq(uint256.NewInt(128)) {
+		return nil, fmt.Errorf("incorrect data offset: %v", dataOffset[0])
+	}
+
+	// 128:160: data length
+	var dataLen uint256.Int
+	dataLen.SetBytes(ev.Data[128:160])
+	if dataLen.Eq(uint256.NewInt(uint64(len(ev.Data) - 160))) {
+		return nil, fmt.Errorf("inconsitent data length: %v", dataLen[0])
```
norswap, line 81: Typo: "inconsitent"

opnode/l2/driver.go
```diff
@@ -2,88 +2,203 @@ package l2
...
-	return false
+// CheckReceipts sanity checks that the receipts are consistent with the block data.
+func CheckReceipts(block *types.Block, receipts []*types.Receipt) bool {
+	hasher := trie.NewStackTrie(nil)
+	computed := types.DeriveSha(types.Receipts(receipts), hasher)
+	return block.ReceiptHash() == computed
 }
 
-type OpaqueTransaction []byte
+const L1InfoDepositIndex uint64 = 0xFFFF_FFFF_FFFF_FFFF
```
norswap, line 99: At first this was called "L1 attributes deposits", then it was going to be "system deposit". I love a good naming bikeshed, but it would be nice to preserve a naming consistency, while we have the naming conversation elsewhere.

opnode/l2/driver.go
```diff
@@ -2,88 +2,203 @@ package l2
...
+	// 32:64: gas - big-endian
+	gas := new(big.Int).SetBytes(ev.Data[32:64])
+	if !gas.IsUint64() {
+		return nil, fmt.Errorf("bad gas value: %x", ev.Data[32:64])
+	}
+	dep.Gas = gas.Uint64()
+	// 64:96: isCreation - boolean, aligned with end.
+	// If False == 0 then it will create a contract using L2 account nonce to determine the created address.
+	if ev.Data[95] == 0 {
+		dep.To = &to
```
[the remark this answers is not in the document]
protolambda: `dep.To` needs to be nil as a contract creation. Otherwise it would be a call to the zero address with contract data as calldata. We can map `to == 32 zero bytes` case to `dep.To == nil`, but then you couldn't do that transaction to the zero address anymore (it would be weird, but technically breaks compatibility)

opnode/l2/driver.go
protolambda: Just picked something descriptive at that moment, I'll try to make naming more consistent with the remaining refactoring/polishing.

opnode/l2/driver.go
protolambda: gas is limited to 30 M on mainnet, even 1000 M is still well within 64 bits. Thus the go-ethereum implementation represents gas as a uint64. Any value higher than that is invalid. The contract must not output gas values higher than 64 bits. (or alternatively, we can clip it to max uint64 in the driver, will need spec discussion)

opnode/l2/driver.go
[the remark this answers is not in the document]
norswap: Ah right, I forgot we don't have the boolean in the structure here, sorry about that!

PR #91, opened 2021-12-21
Add L1Block values contract

maurelian:
This PR:
1. Splits the contracts package into L1 and L2 contracts
2. Adds a new contract `L1Block.sol`, which conforms to the spec for the [L1 Attributes Predeploy](https://github.com/ethereum-optimism/optimistic-specs/blob/1b9c7eb618a218b3464d0258360d69ccd958bfc6/deposits.md#L129).

To get the `deployedBytcode` for the L2 genesis file: 

1. Pull this branch and build using `yarn build`
2. Print the bytecode using 
   `cat specs/packages/contracts/artifacts/contracts/L2/L1Block.sol/L1Block.json | jq .deployedBytecode`


norswap:
Should we add the instructions for generating the genesis file somewhere in the repo too?

protolambda (approved):
LGTM

norswap:
@protolambda @maurelian I don't think this was complete, e.g. there was a test case that was not implemented at least.

Was the deployed bytecode also supposed to be part of this PR?

protolambda:
@norswap good catch, we'll need to add that test, I'll create an issue for it.

Bytecode is an output artifact, I don't think we should host that in the repo (or we need CI to check that it's consistent with the source code)

norswap:
Right, it's part of the output of solc, that checks out.

Issue #85, opened 2021-12-16
Specify ETH & ETH deposit handling


norswap:
cf. this comment: https://github.com/ethereum-optimism/optimistic-specs/pull/66
and historical context: https://github.com/ethereum-optimism/optimistic-specs/discussions/84

On the current EVM, the storage for balances are always 0 BUT this doesn't change anything for users: things like the `BALANCE` opcode are hardwire to a ERC20 contract whose transfer and approval functions all revert.

One downside of that difference is that a Merkle proof scheme for proving user balances on L1 would need to be modified to work on L2.

On the other hand, using native ETH has design implications: we can't use the bridge pattern anymore (since you can't "mint" the native currency in a contract) and would need to add a new deposit transaction type, which would be interpreted specially by the execution engine and cause ETH to be minted.

Questions:
- Do we want to track ETH in the proper storage slots in Optimism 1.0?
- If so, how do we design the mechanism for minting native ETH? Is the custom transaction type the only option?

Random remarks:
- We can't expose ETH as both an ERC20 and the normal balances, because that would violate programmer's assumptions (you don't expect that transaction with an ERC20 would change the ETH balance underneath you).
- We don't need to support sending ETH as part of a deposit: i.e. calling a contract X + minting ETH in the same deposit. Instead, a contract always has the option to perform two deposits instead: one that calls contract X and one that mints ETH (via the standard ETH bridge).

maurelian:
Related discussion on #68

norswap:
One major issue with the current design (deposits have a "value" that is ETH locked in the deposit contract and subsequently minted on L2) is that this does not allow deposit to spend ETH from their associated L1 contracts.

We thought of two solutions:
1. Do not lock ETH in the deposit contract, any sent value is L2 ETH. The behaviour of sending (or calling a contract with) L1 ETH can be reproduced by first depositing the ETH, then sending it. To deposit the ETH, a message would be sent to a special pre-compiled contract that mints L2 ETH. Only a L1 ETH bridge contract would be allowed to call this conctract to mint.
2. Have two fields on deposit transaction, one for the L1 ETH value, one for the L2 ETH value. In another PR, proto has named these `mint` and `value` respectively, which I kinda like (althought it should be clarified that they are both like the normal value field: they are both sent with the transaction).

[...]

PR #95, opened 2021-12-22
Split 'value' into 'mint' and 'value'

maurelian:
Replaces the `value` argument of the `TransactionDeposited` event, and Deposit Transaction type with two arguments:

1. `depositValue` is the ETH amount sent to the contract, to be minted on L2.
2. `sendValue` is the ETH amount to be sent to the recipient (analogous to `value` in s a typical L1 transaction).

### DIFF
--- a/packages/contracts/contracts/L1/DepositFeed.sol
+++ b/packages/contracts/contracts/L1/DepositFeed.sol
@@ -14,8 +14,8 @@ contract DepositFeed {
     event TransactionDeposited(
         address indexed from,
         address indexed to,
-        uint256 depositValue,
-        uint256 sendValue,
+        uint256 mint,
+        uint256 value,
         uint256 gasLimit,
         bool isCreation,
         bytes data
@@ -25,14 +25,14 @@ contract DepositFeed {
      * Accepts deposits of ETH and data, and emits a TransactionDeposited event for use in deriving
      * deposit transactions.
      * @param _to The L2 destination address.
-     * @param _sendValue The ETH value to send in the deposit transaction.
+     * @param _value The ETH value to send in the deposit transaction.
      * @param _gasLimit The L2 gasLimit.
      * @param _isCreation Whether or not the transaction should be contract creation.
      * @param _data The input data.
      */
     function depositTransaction(
         address _to,
-        uint256 _sendValue,
+        uint256 _value,
         uint256 _gasLimit,
         bool _isCreation,
         bytes memory _data
@@ -49,6 +49,6 @@ contract DepositFeed {
             }
         }
 
-        emit TransactionDeposited(from, _to, msg.value, _sendValue, _gasLimit, _isCreation, _data);
+        emit TransactionDeposited(from, _to, msg.value, _value, _gasLimit, _isCreation, _data);
     }
 }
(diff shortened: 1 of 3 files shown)


deposits.md
```diff
@@ -23,11 +23,12 @@ fields (rlp encoded in the order they appear here):
...
 
-- `address from`
-- `address to`
-- `uint256 value`
-- `bytes data`
-- `uint256 gasLimit`
+- `address from`: The address of the sender account.
+- `address to`: The address of the recipient account.
+- `uint256 depositValue`: The ETH value to mint on L2.
+- `uint256 sendValue`: The ETH value to send to the recipient account.
```
maurelian, line 29: Another option would be something like this, which might be cleaner because it adds a new field alongside the existing `value` field in other transaction types, rather than replacing it with 2 new fields.

```suggestion
- `uint256 deposited`: The ETH value to mint on L2.
- `uint256 value`: The ETH value to send to the recipient account.
```

deposits.md
```diff
@@ -82,7 +83,7 @@ follows:
   [aliased][address-aliasing] by the deposit feed contract).
 - `context.calldata` set to `data`
 - `context.gas` set to `gasLimit`
-- `context.value` set to `value`
+- `context.value` set to `sendValue`
```
maurelian, line 86: Could there be a need for app-devs to access the `depositedValue` amount in the context?

norswap:
Relevant issue: https://github.com/ethereum-optimism/optimistic-specs/issues/85

maurelian:
One potential pitfall here, which @tynes pointed out, is that an L1 contract making a deposit could end up stranding ETH on L2. This would occur in the case that the contract calls the deposit function with both `msg.value > sendValue`. In that case, the difference (`msg.value - sendValue`) would be minted to the contract's address alias on L2. 

Unless the contract has been specifically designed to recover this balance, it would not be spendable as the private key is unknown.

However this possibility exists with many other design choices, including with the currently operating bridges (a contract could send value to itself). 

If we reject this change, and stick with requiring that the deposit value be the same as the send value, then it would be impossible (in a Deposit-only Rollup, or with a censoring Sequencer), to transfer an existing L2 balance.

protolambda:
When Norswap mentioned we needed to both deposit and move funds, I was interpreting it as:
- `value` stays the same as regular tx: it moves ETH from one account to the other, and is available through EVM execution etc.
- `mint` is added: it creates ETH on the L2. It is up to the L1 to lock it up during the deposit flow, for later withdrawals
- If `value == mint`: mint first in creator account, before EVM execution.
- If `value > mint`: mint first, and then move value from that account. If `account_funds + mint < value` then fail the transaction.
- If `value <= mint`: mint first, and then move value from that account. Note that there will always be enough funds during execution.
- If EVM fails, the mint is still persisted, and stays at the tx origin account of the L2 execution (the From address). No lost ETH if EVM fails somehow.

This also means that the contract sets `mint` to `msg.value` (whatever it locks up will be minted on L2) and `value` is arbitrary (and won't be used for minting, and might cause the tx to fail on L2 if it's too high)

protolambda (approved):
LGTM! Once the deposit fee design moves towards completion we can deal with enhanced `msg.value` testing (some of that value must be subtracted to cover fees, it's not all `mint` if we pay fees in some designs)

PR #101, opened 2021-12-23
block derivation ref impl: Tx index fix (staging)

protolambda:
Deposit transactions were not unique when multiple are emitted by the same L1 transaction.

The fix changes the tx-index to match the deposit-index within the block (the L1-info tx that is prepended and not emitted based on a log gets a different special value).

We may change this later, the specs skip over uniqueness of deposit transaction hashes, this is just for end-to-end test 0.

### DIFF
--- a/opnode/l2/driver.go
+++ b/opnode/l2/driver.go
@@ -23,6 +23,7 @@ var (
 )
 
 // UnmarshalLogEvent decodes an EVM log entry emitted by the deposit contract into typed deposit data.
+//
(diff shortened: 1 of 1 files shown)


protolambda:
Merging and rebasing staging on `main` again. Seeing what can be done for end-to-end. In the new year we can make a specs PR to `main` that defines deposit-uniqueness (with l1 height + deposit index, or based on L2 height/tx index, what's in `staging` doesn't have to be final).

PR #107, opened 2022-01-03
Add a build command to package.json

maurelian:
**Description**
A very small change to add a missing yarn command, so that we can run `yarn build` in the specs dir.


protolambda (approved):
LGTM

PR #108, opened 2022-01-03
Add missing auth test to L1Block spec

maurelian:
**Description**

Addresses the unimplemented test in #92 and Fixes #103. 

Also fixes some formatting warnings that slipped through.

PR #109, opened 2022-01-03
Use custom solidity errors

maurelian:
**Description**

Uses solidity's new [custom errors](https://docs.soliditylang.org/en/latest/contracts.html#errors) instead of reason strings, this approach saves gas and IMO is cleaner.

This PR builds on #108, otherwise they would conflict.


norswap:
Sounds good to me!

PR #118, opened 2022-01-07
specs: Add README with design goals

### LINKED ISSUE #82: Doc: Outline Design Goals
We should clarify our design goals, to simplify decision making.

This would be good to include in an introduction document (#77)

maurelian:
**Description**
Adds a minimal README file to the specs directory, outlining the design goals. 

It could be argued that this file should include more information, such as usage info for installing or linting, but that should be separate PR.

**Metadata**
- Fixes #82

### DIFF
--- a/specs/README.md
+++ b/specs/README.md
@@ -3,6 +3,13 @@
 This directory contains the plain english specs for Optimism, an minimal optimistic rollup protocol
 that maintains 1:1 compatibility with Ethereum.
 
+## Specification Contents
+
+- [Glossary](glossary.md)
+- [Deposits](deposits.md)
+- [Execution Engine](exec-engine.md)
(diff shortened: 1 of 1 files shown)


norswap (approved):
Looks good to me.

I can't resist to add (sorryyyy!)  that I'm always a bit skeptical of such design goals sections. Who doesn't want to be fast, scalable and cheap? It's more useful to specify preferences in trade-offs or non-goals which are things  that other people would find valuable. Anyway, this is really fine as-is, and nobody won't be losing any sleep over this.

maurelian:
> It's more useful to specify preferences in trade-offs or non-goals which are things that other people would find valuable. 

This is a good point, I strongly agree, but also think this is good enough for now. 
We may want to update this in the future based on our history of decision making.

PR #117, opened 2022-01-07
Make rollup node spec consistent with deposit spec

norswap:
Closes https://github.com/ethereum-optimism/optimistic-specs/issues/81

This is the first part of a series of spec improvement PR. This one makes the rollup node and the deposit spec consistent.

The main change there was that we settled on the "L1 attributes transaction" to be an " L1 attributes deposit" sharing the same format as user-submitted deposits (now called "transaction deposits"). Most of the change in this PR reflect this change in the rollup node spec and in the glossary.

Subsequent PRs will:
- further adopted [standardized language for deposit](https://github.com/ethereum-optimism/optimistic-specs/issues/87) when we reach agreement on terms
- fix links
- rework the rollup node spec to be simpler, notably including a [high-level pseudo-code description](https://github.com/ethereum-optimism/optimistic-specs/issues/74)

(Note a few parasitic changes in the `deposits.md` file because it was aligned to a 100 characters boundary except for a few paragraphs that I have realigned. Sorry about the noise!)

Issue #116, opened 2022-01-07
Lint for broken links in markdown


maurelian:
We should be automating the process of finding and fixing broken links to save time. 

This is lower priority than #115, and can happen in the next milestone.

maurelian:
closed by #120

PR #120, opened 2022-01-10
Fix broken links

### LINKED ISSUE #115: Fix Links
When moving the specs to the `specs` subdir, a lot of links died. Fix them by making them relative.

norswap:
Closes #115 and #116 

Built on top on https://github.com/ethereum-optimism/optimistic-specs/pull/117, don't merge!

Uses https://github.com/lycheeverse/lychee for link checking.

~A bit unsure about the github workflow, needs testing.~ It works.

~I'm expecting the workflow to fail at first, because we have one link to a build output, so either
we must add contract building to the workflow, or add an exception.~
It passes, but that itself is a problem, because it's caused by lychee not checking absolute (`/`-prefixed) links.


specs/deposits.md
```diff
@@ -148,21 +155,25 @@ The contract has the following solidity interface, and can be interacted with ac
...
 
-[L1Block.sol]: /packages/contracts/contracts/L1Block.sol
+[L1Block.sol]: ../packages/contracts/contracts/L2/L1Block.sol
 
-The bytecode to add to the genesis file will be located in the `deployedBytecode` of the
-[JSON artifact file][l1-block-artifacts] corresponding to L1Block.sol.
+The bytecode to add to the genesis file will be located in the
+`deployedBytecode` of the [JSON artifact file][l1-block-artifacts] corresponding
+to L1Block.sol (this link will be broken unless you've built the contracts with
+`yarn build` in the `packages/contracts` directory).
```
norswap, line 167: Maybe just remove this link and display the path textually instead. @maurelian wdyt?

norswap:
I couldn't get it to check absolute links. Opened an issue @ lychee here: https://github.com/lycheeverse/lychee-action/issues/70

In the meantime, I've changed the absolute links from `deposits.md` to be relative, except for the artifact links, which will therefore be conveniently ignored (though I think we should probably change that by a textual path, see my comment).

specs/deposits.md
```diff
@@ -197,13 +209,13 @@ The deposit feed handles two special cases:
 [address-aliasing]: #address-aliasing
 
 If the caller is not a contract, the address will be ed by adding
```
trianglesphere, line 211: Not quite in PR, but what's `the address will be ed by adding...`

specs/glossary.md
```diff
@@ -186,57 +187,76 @@ cf. [Execution Engine Specification](exec-engine.md)
...
-(number, timestamp, ...) on L2.
+A deposit is an L2 transaction derived from an L1 block.
 
-The L1 attributes for a given L1 block can be read on L2 from the [L1 Attributes Predeployed
-Contract][l1-attr-predeploy].
+There are two kinds of deposits:
 
-cf. [L1 attributes transaction format](/rollup-node.md#payload-transaction-format) (in the section on [payload
-attributes])
+- [L1 Attributes Deposit][l1-attribute-tx], which submit the L1 block's attributes to the [L1 Attributes Predeployed
```
trianglesphere, line 198: Github doesn't recognize the first pair of square brackets as a link

specs/glossary.md
```diff
@@ -186,57 +187,76 @@ cf. [Execution Engine Specification](exec-engine.md)
...
+While transaction deposits are notably (but not only) used to "deposit" (bridge) ETH and tokens to L2, the word
+*deposit* should be understood as "a transaction *deposited* to L2 from L1".
 
-> **TODO** We might wish to make this a "normal transaction" if deposits end up
-> not carrying a signature.
+The term *deposit* performs the double duty of referring to both:
 
-## L1 Attributes Predeployed Contract
+1. The transaction information submitted on L1 to the [deposit feed contract][deposit-feed]. This is a kind of [L2
+   derivation input][deriv-input].
```
trianglesphere, line 209: ditto with `derives-input`

specs/glossary.md
norswap: Yup, that should be `[l1-attributes-deposit]`.

norswap:
Following @trianglesphere 's comment, I rechecked and found quite a few more broken link references (e.g. `[text][link-ref]`). Would be nice if the tool could catch those automatically too.

specs/deposits.md
maurelian, line 167:
```suggestion
After running `yarn build` in the `packages/contracts` directory, the bytecode to add to the 
genesis file will be located in the `deployedBytecode` field of the build artifacts file at 
`/packages/contracts/artifacts/contracts/L2/L1Block.sol/L1Block.json`.
```

maurelian: Agree. I made a suggestion below that I think gets this across well.

maurelian (commented):
I think it's good with these small changes

norswap:
- had to exclude `twitter.com` because of rate limiting
- had to exclude email address [because this](https://github.com/lycheeverse/lychee/issues/471)
- I'm explicitly listing files to lint more restrictively, otherwise this gets quite slow locally as it's trying to file markdown files in `.git` and `node_modules`
   - this does mean that I'm not currently linting markdown file in the javascript packages
   - we could add markdown files one by one though
   - [this would fix it](https://github.com/lycheeverse/lychee/issues/470)

PR #123, opened 2022-01-10
Input deriv test cleanup

protolambda:
A mostly testing-only PR + `DepositEventABI` update (bug found with e2e testing by @maurelian), final polish before moving on with `staging` PRs to `main`.

Changes:
- L1 info deposit testing happens in reader-tests already (we go back and forth to encoded form, and check if the roundtrip doesn't lose info. With Go 1.18 we can fuzz that roundtrip later)
- `DeriveUserDeposits` returns a list of types deposit transactions, instead of the the geth transaction wrapper.
- No need to mock a whole block anymore, we only use the block-height for user deposits.
- Re-instate user deposits testing, with non-deposit logs, and less code

### DIFF
--- a/opnode/l2/input_derivation_test.go
+++ b/opnode/l2/input_derivation_test.go
@@ -6,72 +6,12 @@ import (
 	"math/big"
 	"math/rand"
 	"testing"
-	"time"
 
 	"github.com/stretchr/testify/assert"
 
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/core/types"
-	"github.com/ethereum/go-ethereum/trie"
 )
-
-var (
-	// testnet chain ID, not meaningful here
-	chaindID = big.NewInt(69)
-)
-
-type deriveTxsTestInput struct {
-	block    *types.Block
-	receipts []*types.Receipt
-}
-
-// Generates a test case with nSuccess successful transactions and nFailed failed transactions,
-// as well as corresponding deposits.
-// The transactions are dummy transactions, that do not match the log entries contained the deposits.
-func GenerateTest(nSuccess uint64, nFailed uint64, rng *rand.Rand) *deriveTxsTestInput {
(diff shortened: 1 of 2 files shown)


opnode/l2/input_derivation_test.go
```diff
@@ -241,65 +125,71 @@ func TestUnmarshalLogEvent(t *testing.T) {
...
-	input    *deriveTxsTestInput
-	expected []*types.Transaction
-}
-
-func TestDeriveUserDeposits(t *testing.T) {
-	testCases := []DeriveUserDepositsTestCase{
-		{"no deposits", GenerateTest(0, 0), []*types.Transaction{}},
-		{"success deposit", GenerateTest(1, 0), []*types.Transaction{nil}}, // TODO
-		{"failed deposit", GenerateTest(0, 1), []*types.Transaction{}},
-		{"many deposits", nil, nil}, // TODO
```
trianglesphere, line 256: Are these test cases being ran in `reading_test.go` or is that just testing round trip?

trianglesphere (commented):
One question about test cases for future work, but this seems good

opnode/l2/input_derivation_test.go
protolambda: See new `TestDeriveUserDeposits` below, the diff isn't clear here. Tested a bunch of edge cases with deposits success/fail/logs/non-deposit-logs/etc. And no need to generate a full block or hardcode transactions, since we first generate the deposits, then convert those to logs (along with other non-deposit logs etc. as the test case specifies), and then check if `DeriveUserDeposits` derives equal deposits from those logs.

It's indeed similar to the L1-info deposit round trip test, but for user-deposits here, instead of the L1-info deposit.

protolambda: Regarding the diff, #128 includes the changes and may be easier to review (also includes the `reading_test.go` etc.)

Issue #119, opened 2022-01-08
EPIC CHECKLIST: Review/Integrate staged deposit-only ref impl work


protolambda:
In the `staging` branch you can find a `opnode` dir with the reference implementation for the deposit-only milestone.
You can run it end-to-end with [these instructions](https://github.com/protolambda/rollup-node-experiments/).

To simplify review, clean up the commit history, and help organize, we can split the staging branch into indepdent PRs:
- [x] `opnode/cmd` + Go CI + go.sum file: boring boilerplate to run it as binary + CI setup, minimum compiling Go:
  #124
- [x] `opnode/eth`: API interfaces, api utils and types:
  #125
- [x] `opnode/l1`: parallized/cached blocks and receipts downloader with tests:
  #126
- [x] `opnode/l2/api.go`: Execution Engine API client bindings:
  #127
- [x] `opnode/l2/{reading.go, input_derivation.go}`: convert block/receipts (interfaces for mocking in tests) to deposit transactions:
  #128
- [x] `opnode/l2/{output_derivation.go, driver_step.go}`: where the we derive outputs and apply the block to the engine:
  #129
- [x] `opnode/l2/{sync_reference.go, sync_start.go}`: how we resolve where sync needs to start:
  #130
- [x] `opnode/l2/driver.go`: the main event system of engine sync:
  #131
- [x] `opnode/node/`: rollup-node start (init, manage and shut down all L1 and L2 processes):
  #132

protolambda:
completed all PRs, closing this

PR #124, opened 2022-01-10
ref impl: CI + opnode Go base

protolambda:

This:
- Go CI (title)
- Updates linting docs with Go instructions
- Adds Go instructions from CI
- Defines the `optimistic-specs` go module with `opnode` package
- Adds the necessary boilerplate to run the process, shut it down. A.k.a. `opnode/cmd`
- Adds logging command options

*Side-note: Not cherry-picking from `staging`, squashing noise + cleaning up tweaks*


norswap (approved):
Looks pretty sweet to me!

In the future, we'll probably want to write some short markdown explanation of the command-based architecture. This is not directly relevant to this PR, but the fact that some of the state is held within the commands `struct`s for instance.

protolambda:
@norswap good point, will add a `opnode/README.md` as part of #132 (where we define the node options and actually gets interesting to run)

PR #125, opened 2022-01-10
ref impl: eth utils - watch head, hash/height id type, RPC source interfaces

protolambda:

This:
- Introduces different types of `Source` interfaces: abstract away the JSON RPC from the ref impl. And more granular than e.g. the `ethereum.ChainReader`
- Simple combined-source: to support multiple L1 sources (can replace with smarter load-balancing later on)
- Wrap a header subscription to provide a channel of `HeadSignal` updates. Used to track L1 and L2 chain (requires websocket RPC, note that `staging` does have a fallback)
- `BlockID`: combine hash and height, makes the sync implementation more readable

**Depends on #124**

Review: any team. No specs affected other than defining `BlockID`. Sources are named based on Go-ethereum method naming.

### DIFF
--- a/opnode/eth/canonical.go
+++ b/opnode/eth/canonical.go
@@ -1,0 +1,36 @@
+package eth
+
+import (
+	"context"
+	"fmt"
+	"math/big"
+)
+
+// BlockLinkByNumber Retrieves the *currently* canonical block-hash at the given block-height, and the parent before it.
+// The results of this should not be cached, or the cache needs to be reorg-aware.
+type BlockLinkByNumber interface {
+	BlockLinkByNumber(ctx context.Context, num uint64) (self BlockID, parent BlockID, err error)
+}
+
+type BlockLinkByNumberFn func(ctx context.Context, num uint64) (self BlockID, parent BlockID, err error)
+
+func (fn BlockLinkByNumberFn) BlockLinkByNumber(ctx context.Context, num uint64) (self BlockID, parent BlockID, err error) {
+	return fn(ctx, num)
+}
+
+// CanonicalChain presents the block-hashes by height by wrapping a header-source
+// (useful due to lack of a direct JSON RPC endpoint)
+func CanonicalChain(l1Src HeaderByNumberSource) BlockLinkByNumber {
+	return BlockLinkByNumberFn(func(ctx context.Context, num uint64) (BlockID, BlockID, error) {
+		header, err := l1Src.HeaderByNumber(ctx, big.NewInt(int64(num)))
+		if err != nil {
+			// w%: wrap the error, we still need to detect if a canonical block is not found, a.k.a. end of chain.
+			return BlockID{}, BlockID{}, fmt.Errorf("failed to determine block-hash of height %d, could not get header: %w", num, err)
+		}
+		parentNum := num
+		if parentNum > 0 {
+			parentNum -= 1
+		}
+		return BlockID{Hash: header.Hash(), Number: num}, BlockID{Hash: header.ParentHash, Number: parentNum}, nil
+	})
+}
--- a/opnode/eth/canonical.go
+++ b/opnode/eth/canonical.go
@@ -12,6 +12,7 @@ type BlockLinkByNumber interface {
 	BlockLinkByNumber(ctx context.Context, num uint64) (self BlockID, parent BlockID, err error)
 }
 
+// BlockLinkByNumberFn implements BlockLinkByNumber to implement the interface as anonymous function
 type BlockLinkByNumberFn func(ctx context.Context, num uint64) (self BlockID, parent BlockID, err error)
 
 func (fn BlockLinkByNumberFn) BlockLinkByNumber(ctx context.Context, num uint64) (self BlockID, parent BlockID, err error) {
--- a/opnode/eth/heads.go
+++ b/opnode/eth/heads.go
@@ -13,10 +13,10 @@ type HeadSignal struct {
 	Self   BlockID
 }
 
-// HeadSignalFn is called to inform a chain with a new head
+// HeadSignalFn is used as callback function to accept head-signals
 type HeadSignalFn func(sig HeadSignal)
 
-// WatchHeadChanges wraps a new-head subscription from ChainReader to feed the given Tracker
+// WatchHeadChanges wraps a new-head subscription from NewHeadSource to feed the given Tracker
 func WatchHeadChanges(ctx context.Context, src NewHeadSource, fn HeadSignalFn) (ethereum.Subscription, error) {
 	headChanges := make(chan *types.Header, 10)
 	sub, err := src.SubscribeNewHead(ctx, headChanges)
--- a/opnode/eth/heads.go
+++ b/opnode/eth/heads.go
@@ -1,0 +1,48 @@
+package eth
+
+import (
+	"context"
+
+	"github.com/ethereum/go-ethereum"
+	"github.com/ethereum/go-ethereum/core/types"
+	"github.com/ethereum/go-ethereum/event"
+)
+
+type HeadSignal struct {
+	Parent BlockID
+	Self   BlockID
+}
+
+// HeadSignalFn is called to inform a chain with a new head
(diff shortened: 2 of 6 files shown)


opnode/eth/canonical.go
```diff
@@ -0,0 +1,36 @@
...
+type BlockLinkByNumber interface {
+	BlockLinkByNumber(ctx context.Context, num uint64) (self BlockID, parent BlockID, err error)
+}
+
+type BlockLinkByNumberFn func(ctx context.Context, num uint64) (self BlockID, parent BlockID, err error)
+
+func (fn BlockLinkByNumberFn) BlockLinkByNumber(ctx context.Context, num uint64) (self BlockID, parent BlockID, err error) {
+	return fn(ctx, num)
+}
+
```
norswap, line 20: This is some non-intuitive Go hackery, could we
- make `BlockLinkByNumberFn` private?
- add a comment to the effect that we need to introduce `blockLinkByNumberFn` so that it can implement `BlockLinkByNumber` and let us use an anonymous function a an implementation of `BlockLinkByNumber`?

opnode/eth/heads.go
```diff
@@ -0,0 +1,48 @@
...
+	"github.com/ethereum/go-ethereum/core/types"
+	"github.com/ethereum/go-ethereum/event"
+)
+
+type HeadSignal struct {
+	Parent BlockID
+	Self   BlockID
+}
+
+// HeadSignalFn is called to inform a chain with a new head
```
norswap, line 16: Not sure what is meant by "inform a chain with a new head"? Does it just mean a new L1 head was detected?

opnode/eth/heads.go
```diff
@@ -0,0 +1,48 @@
...
+type HeadSignalFn func(sig HeadSignal)
+
+// WatchHeadChanges wraps a new-head subscription from ChainReader to feed the given Tracker
```
norswap, line 19: What is `ChainReader`? Outdated comment and it's now `NewHeadSource`?
Same with `Tracker` - `HeadSignalFn`?

opnode/eth/source.go
```diff
@@ -0,0 +1,127 @@
...
+func (fn BlockByNumFn) BlockByNumber(ctx context.Context, number *big.Int) (*types.Block, error) {
+	return fn(ctx, number)
+}
+
+// CombinedL1Source balances multiple L1 sources, to shred concurrent requests to multiple endpoints
+type CombinedL1Source struct {
+	i       uint64
+	sources []L1Source
+}
+
```
norswap, line 94: I would document how the combined source work (i.e. round-robin, using `i` to track the last used source).

Also maybe "dispatch between" or "load-balance between" instead of "shred to"?

opnode/eth/source.go
```diff
@@ -0,0 +1,127 @@
...
+func NewCombinedL1Source(sources []L1Source) L1Source {
+	if len(sources) == 0 {
+		panic("need at least 1 source")
+	}
+	return &CombinedL1Source{i: 0, sources: sources}
+}
+
+func (cs *CombinedL1Source) HeaderByHash(ctx context.Context, hash common.Hash) (*types.Header, error) {
+	return cs.sources[atomic.AddUint64(&cs.i, 1)%uint64(len(cs.sources))].HeaderByHash(ctx, hash)
```
norswap, line 103: Suggestion: isolate `atomic.AddUint64(&cs.i, 1)%uint64(len(cs.sources))` to its own private function, since it's used quite a bit. e.g. `func (cs *CombinedL1Source) nextSource() L1Source`

norswap (approved):

opnode/eth/heads.go
protolambda: typo, missing word after renaming things, my bad

opnode/eth/heads.go
protolambda: `ChainReader` is a super common ethereum interface, in the very root package: https://github.com/ethereum/go-ethereum/blob/045e90c8971cddbabeac0abd54240cf02bc1d94d/interfaces.go#L54
Will update with package name to clarify where it's coming from.

opnode/eth/canonical.go
protolambda: It's a common pattern to define a function type that implements single-function interfaces. Interfaces are more flexible to modify and extend, whereas functions make implementing it really easy.

With most of the RPC interfaces and functions I hope we get different implementations later on; more advanced caching, load-balancing and additional test implementations. E.g. the `l2` package might use it during testing.

protolambda: Keeping the `BlockLinkByNumberFn` public, updated doc to describe the purpose better

opnode/eth/heads.go
```diff
@@ -0,0 +1,48 @@
...
+type HeadSignal struct {
+	Parent BlockID
+	Self   BlockID
+}
+
+// HeadSignalFn is used as callback function to accept head-signals
+type HeadSignalFn func(sig HeadSignal)
+
+// WatchHeadChanges wraps a new-head subscription from NewHeadSource to feed the given Tracker
+func WatchHeadChanges(ctx context.Context, src NewHeadSource, fn HeadSignalFn) (ethereum.Subscription, error) {
```
trianglesphere, line 20: I'm a little less familiar with the internals of the Eth subscription system, but is there a design where the subscription returns the `HeadSignal` events and we don't need a callback?

protolambda: I like channels better than callbacks too, but we're sending it to an `event.Feed` to fan out to multiple channels anyway.
See https://github.com/ethereum-optimism/optimistic-specs/blob/061eb488c2ba02e3d5ec8b70abeda5c9ea6d0645/opnode/node/node.go#L193

opnode/eth/canonical.go
trianglesphere: Necroing b/c I haven't seen this pattern. Interfaces are really nice when you truly don't care about the implementation (`io.Reader` or `io.Writer`), and also when you want a test vs real implementation.

I get splitting up the interface (particularly further down), but I'm curious to see how the type/anonymous function thing works out.

protolambda: https://pkg.go.dev/net/http#HandlerFunc is an example of this pattern in the standard library, and you've probably used it somewhere without realizing already, it's everywhere :)

norswap: I saw it was being used in another PR and thought maybe it was a common pattern. Does it have a name?
For sure explaining it everywhere is very impractical. If it has a well-known googlable name, e.g. "interaface adapter function type" we could tag uses of the pattern with something like `// interface adapter function type`.

Or we could just let people figure it out I suppose. We do assume a fair amount of Go understanding.

protolambda: "anonymous function interface implementation" I guess? You can find it on google that way, not that well documented though.

opnode/eth/source.go
```diff
@@ -0,0 +1,135 @@
...
+
+type BlockByNumFn func(ctx context.Context, number *big.Int) (*types.Block, error)
+
+func (fn BlockByNumFn) BlockByNumber(ctx context.Context, number *big.Int) (*types.Block, error) {
+	return fn(ctx, number)
+}
+
+// CombinedL1Source implements round-robin between multiple L1 sources,
+// to divide concurrent requests to multiple endpoints
+type CombinedL1Source struct {
```
trianglesphere, line 94: This is something that I want to watch closely going forwards. I'm a little worried that L1 nodes are not in sync with each other which could cause issues with figuring out the canonical chain or followup requests about block state.

opnode/eth/canonical.go
trianglesphere: I like `function to interface adapter` or something along those lines. Important to me is to keep `adapter` and I don't really care for `anonymous` as the function does not need to be anonymous.

I'm fine leaving this as is, but from looking at and playing with it, my gut feeling that it's a little overkill remains. I think the function adapter pattern is really useful when you have a pure function and it avoids having to create an empty struct and define the pure function over that struct, but once you start to rely on closures, I'd just create a struct to implement it.

PR #126, opened 2022-01-10
ref impl: L1 downloader of blocks and receipts

protolambda:

This:
- Downloads L1 blocks, and parallelizes the fetching of receipts (since JSON-RPC doesn't bundle it nicely)
- Provides Downloader interface. In the future we could also implement it with GraphQL (geth supports it), to download everything in one Go, if we provide a geth L1 source. More efficient but less generic.
- Download jobs and history are limited with a LRU (multiple engines might be using the downloader)
- Downloader receipt work is split between workers (don't blow up goroutine-count with high receipts numbers)
- Receipt fetching will be retried (without retrying forever, mixed with worker queue)
- Evicted jobs get cancelled
- Duplicate requests get de-duplicated (sync multiple L2 engines faster)
- With tests for different failures / receipts / worker counts

**Depends on #125**

Review: systems team preferred. No spec affected (just implements a simple downloader interface: `Fetch(ctx context.Context, id eth.BlockID) (*types.Block, []*types.Receipt, error)`)


opnode/l1/downloader.go
```diff
@@ -0,0 +1,220 @@
...
+			case <-quit:
+				return nil
+			}
+		}
+	})
+}
+
+// AddReceiptWorkers can add or remove (negative value) worker routines to parallelize receipt downloads with.
+// It returns the number of active workers.
+func (l1t *downloader) AddReceiptWorkers(n int) int {
```
protolambda, line 206: Maybe good to add a `Close` that essentially does `AddReceiptWorkers(-len(l1t.receiptWorkers))` to clean up? Current closing would be a bit hacky: `l1t.AddReceiptWorkers(-l1t.AddReceiptWorkers(0))`.

opnode/l1/downloader_test.go
```diff
@@ -0,0 +1,206 @@
...
+		return nil, errors.New("no receipt here")
+	}
+	ret.tries -= 1
+	if ret.tries < 0 {
+		return ret.receipt, nil
+	}
+	return nil, fmt.Errorf("receipt takes %d more tries to download", ret.tries)
+}
+
+func RandomL1Block(txCount int) (*types.Block, []*types.Receipt) {
```
norswap, line 49: This seems like a very useful function to have in some kind of test util library! Similarly with its subparts (e.g. we could have `RandomTx`, etc).

norswap (changes requested):
This is a super neato piece of code! Really like the architecture.

My main feedback is that this could **really** use a top comment briefly describing the architecture. I'm up for writing it, if that's fine?

opnode/l1/downloader.go
```diff
@@ -0,0 +1,220 @@
...
+type downloader struct {
+	// cache of ongoing/completed block tasks: block hash -> block
+	cacheEvict *lru.Cache
+	cacheLock  sync.Mutex
+
+	receiptTasks       chan *receiptTask
+	receiptWorkers     []ethereum.Subscription
+	receiptWorkersLock sync.Mutex
+
+	chr DownloadSource
```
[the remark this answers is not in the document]
protolambda: Started with the `ethereum.ChainReader` interface, then later decided I only need a single method of the interface and called it the `DownloadSource`, then forgot to update the var name. Good catch

opnode/l1/downloader.go
```diff
@@ -0,0 +1,220 @@
...
+	evict := func(k, v interface{}) {
+		// stop downloading things if they were evicted (already finished items are unaffected)
+		v.(*blockAndReceipts).Finish(wrappedErr{downloadEvictedErr})
+	}
+	// 500 at 100 KB each would be 50 MB of memory for the L1 block inputs cache
+	dl.cacheEvict, _ = lru.NewWithEvict(500, evict)
+	return dl
+}
+
+func (l1t *downloader) Fetch(ctx context.Context, id eth.BlockID) (*types.Block, []*types.Receipt, error) {
```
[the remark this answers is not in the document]
protolambda: Good catch, used to be part of the "l1 tracker", `l1d` or `dl` makes more sense now yes

opnode/l1/downloader.go
```diff
@@ -0,0 +1,220 @@
...
+				total := atomic.AddUint32(&task.Dest.DownloadedReceipts, 1)
+				if total == uint32(len(task.Dest.Receipts)) {
+					// block completed without error!
+					task.Dest.Finish(wrappedErr{nil})
+					continue
+				}
+				// task completed, but block is not complete without other receipt tasks finishing first
+			case <-quit:
+				return nil
+			}
```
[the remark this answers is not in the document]
protolambda: it's not a head comment, but says something about the `case` above being done

opnode/l1/downloader.go
```diff
@@ -0,0 +1,220 @@
...
+	return event.NewSubscription(func(quit <-chan struct{}) error {
+		for {
+			select {
+			case task := <-l1t.receiptTasks:
+				// scheduled tasks may be stale if other receipts of the block failed too many times
+				if task.Dest.finished > 0 { // no need for locks, a very rare stale download does not hurt
+					continue
+				}
+				// limit fetching to the task as a whole, and constrain to 10 seconds for receipt itself
+				ctx, cancel := context.WithTimeout(task.Dest.ctx, time.Second*10)
```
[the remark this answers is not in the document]
protolambda: Instead of applying many smaller context time-outs, we maintain one time-out for all the work combined. Will clarify comment

opnode/l1/downloader.go
```diff
@@ -0,0 +1,220 @@
...
+var downloadEvictedErr = errors.New("evicted")
+
+func NewDownloader(chr DownloadSource) Downloader {
+	dl := &downloader{
+		receiptTasks: make(chan *receiptTask, 100),
+		chr:          chr,
+	}
+	evict := func(k, v interface{}) {
+		// stop downloading things if they were evicted (already finished items are unaffected)
+		v.(*blockAndReceipts).Finish(wrappedErr{downloadEvictedErr})
```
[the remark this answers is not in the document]
protolambda: > If new request tend to supercede/replace old requests, then the current behaviour is fine.

Generally they should, since we sync is a linear process. With multiple engines it may be more of a risk, but the cache size is 500 block entries, which should be more than plenty.

opnode/l1/downloader.go
```diff
@@ -0,0 +1,220 @@
...
+)
+
+type blockAndReceipts struct {
+	// Track if we finished (with error or not). May increment >1 when other sub-tasks fail.
+	// First field, aligned atomic changes.
+	finished uint32
+
+	// Count receipts to track status
+	// First field of struct for memory aligned atomic access
+	DownloadedReceipts uint32
```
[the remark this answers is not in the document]
protolambda: ah good catch, two atomic fields, but all 32-bit aligned, should be fine. Will update comment

opnode/l1/downloader_test.go
protolambda: Not too sure, we mostly just care about the receipts. Was even thinking of embedding empty/nil transactions to simplify it. Hiding the block data behind an interface, and only dealing with the minimal instead of generating a full block, would be nice.

opnode/l1/downloader.go
```diff
@@ -0,0 +1,220 @@
...
+
+	"github.com/ethereum-optimism/optimistic-specs/opnode/eth"
+
+	"github.com/ethereum/go-ethereum"
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/core/types"
+	"github.com/ethereum/go-ethereum/event"
+)
+
+type blockAndReceipts struct {
```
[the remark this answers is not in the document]
protolambda: Good suggestion, but think the name is just too complex already, and indeed misses the "task" idea. Renaming it to `downloadTask`, the primary task of the downloader.

opnode/l1/downloader.go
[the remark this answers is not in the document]
norswap: But this isn't done in this file, is it? In `Fetch` we create a timeout, but we also derive this from `blockAndReceipts.ctx`. So the timeout you're referring too is potentially set upstream, by the `Fetch` caller, then?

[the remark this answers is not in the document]
protolambda: It starts with the upstream ctx, which stays with the download-task, and which the block and receipt calls wrap with timeouts.

opnode/l1/downloader.go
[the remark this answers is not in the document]
protolambda: Related edge-case here: on eviction we finish the task, but we should cancel the task context as well, so we free up resources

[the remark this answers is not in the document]
norswap: Mmh, I guess the annoying bit is that the conjugation of concerns: the cache is both a cache but also the mechanism that prevents us from having too many requests running at the same time. If we had some form of task queue, then we could reject new requests while the queue is full.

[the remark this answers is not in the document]
protolambda: We actually do have that: the workers consume from a task queue (`receiptTasks` channel). When the workers start applying back-pressure (full channel, sending a new task will block when consumers are slower than producer and capacity is full) the main `Fetch` calls get blocked. Eviction of a running task should only happen if there's one super slow worker and other parallel fast workers that keep the download cache going.

opnode/l1/downloader.go
```diff
@@ -0,0 +1,271 @@
...
+
+	receiptTasks       chan *receiptTask
+	receiptWorkers     []ethereum.Subscription
+	receiptWorkersLock sync.Mutex
+
+	// source to pull data from. Nil when downloader is closed
+	src DownloadSource
+}
+
+func NewDownloader(dlSource DownloadSource) Downloader {
```
trianglesphere, line 109: I tend to prefer to return a struct and let package that uses the downloader define the interface at the point of usage. (https://blog.chewxy.com/2018/03/18/golang-interfaces/)

protolambda: Me too, a struct is generally preferred indeed, but everything except the interface methods is private in this case. And if returning the (`*downloader`), it's easy to make a bad copy by mistake (it's not supposed to be copied because of the mutex and attached workers). Hiding it with the interface in this case protects us from that mistake.

trianglesphere: I don't really mind that it's only public methods are in the interface. And for the copying, it might be a little more suggestive, but it's still a pointer under the hood and go lets you copy objects that are typed as interfaces..

opnode/l1/downloader.go
```diff
@@ -0,0 +1,271 @@
...
+	dl.cacheLock.Lock()
+	if dl.src == nil {
+		dl.cacheLock.Unlock()
+		return nil, nil, DownloadClosedErr
+	}
+
+	var dlTask *downloadTask
+	if dlTaskIfc, ok := dl.cacheEvict.Get(id.Hash); ok {
+		dlTask = dlTaskIfc.(*downloadTask)
+		dl.cacheEvict.Add(id.Hash, dlTask) // add it again, so it moves to the front and avoid eviction
```
trianglesphere, line 134: `lru.Get` updates the frequency. `lru.Contains` and `lru.Peek` don't update the frequency.

opnode/l1/downloader.go
```diff
@@ -0,0 +1,271 @@
...
+
+type DownloadSource interface {
+	eth.BlockByHashSource
+	eth.ReceiptSource
+}
+
+type downloader struct {
+	// cache of ongoing/completed block tasks: block hash -> block
+	cacheEvict *lru.Cache
+	cacheLock  sync.Mutex
```
trianglesphere, line 99: The LRU cache itself is thread safe, but this lock seems to be protecting against duplicate requests.

opnode/l1/downloader.go
```diff
@@ -0,0 +1,271 @@
...
+	// receipts slice is allocated in advance, one slot for each transaction, nil until downloaded
+	receipts []*types.Receipt
+
+	// stop fetching if this context is dead
+	ctx context.Context
+	// to free up resources (e.g. ongoing receipt tasks) if a downloadTask finishes prematurely
+	cancel context.CancelFunc
+
+	// feed to subscribe the requests to (de-duplicate work)
+	feed event.Feed
```
trianglesphere, line 60: If I'm reading the code correctly, this is currently just used to signal to the Fetch function that it has succeeded or failed. If so, it feels a little overkill.

trianglesphere (commented):
I was leaving drive by comments, so here's a full review.

I'm with norswap that a comment laying out the architecture would be helpful here. One comment on the architecture is that I think having a layer that does synchronous calls to get the block and receipts and then building the concurrency and de-duplication on top of that would make some of the steps easier to follow (they're currently intertwined).

opnode/l1/downloader.go
protolambda: Ofc you can still copy a bare struct if it's typed as interface, but it's `*downloader -> downloader` you can't accidentally do when it's wrapped with the interface.

norswap: I can see both sides of the argument here. I prefer the current way which is slightly clearer, while less idiomatic. I think what it boils down to is whether there will be other implementations of ` Downloader`. If we expect there will be, it makes sense to return the struct instead, making it clear it's just one possible implementation.

Having `Downloader` defined at the use site would be much less clear in this case imho.

opnode/l1/downloader.go
norswap: Do you suggest an alternative to signal completion?

opnode/l1/downloader.go
protolambda: Indeed, the lock makes sure we are not concurrently adding a new duplicate job.

Consider this race:
```
routine A and B get from the cache (in parallel, but safe)
routine A and B find nothing
routine A and B create and add a new job
```
The LRU interface does provide a `ContainsOrAdd`, but that doesn't update the frequency like `Get`, and forces us to allocate before knowing it's a duplicate or not.


I'll try add a comment about the atomic `Get / Add` to avoid duplicate jobs.

opnode/l1/downloader.go
protolambda: It's not, we can't use a regular channel since there may be multiple listeners waiting for the job (duplicates, as described in the comment).

opnode/l1/downloader.go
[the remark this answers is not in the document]
norswap: I was asking where the upstream timeout was set essentially, but I see this happens in `DriverStep`. Marking this as resolved.

opnode/l1/downloader.go
norswap: Good point, is that actually necessary?

protolambda:
Added more doc-comments, hope that explains better how the downloader works. This PR is blocking other reference implementation PRs, can I get a final review?

opnode/l1/downloader.go
trianglesphere: It makes sense to use the feed in this case, but I don't love a design that requires a queue to de-duplicate in the first place

norswap (approved):

PR #128, opened 2022-01-10
ref impl: input derivation and reading l1-info

protolambda:

This:
- Defines receipts to user deposits conversion
- Defines L1-info to deposit conversion
- Defines how to combine these deposits into payload attributes
- Test user-deposits
- Test l1-info deposits
- *(not tested, except in e2e): concatenating above deposits into payload attributes*
- Defines the inverse process of L1-info deposit: part of sync requires us to read which L1 info was last processed on L2, we simply parse the last L1-info transaction.

**Depends on #127**

Review: any team. No wild changes required, but naming is inconsistent, and we can try cross-referencing the functions. And should the inverse process (reading L1 info deposits) be specified as well, or is it implicitly defined already?

This took more dev iteration on staging, deriving inputs before using the engine-api for the outputs part can have many different typing approaches. This ended up working best for testing (minimize type transformations, simplify typing, interfaces / bare info instead of full blocks etc.).

Note: this also includes #123 (we can cherry-pick commits from review, or close that older PR)

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -35,7 +35,6 @@ require (
 	golang.org/x/sys v0.0.0-20210816183151-1e6c022a8912 // indirect
 	gopkg.in/natefinch/npipe.v2 v2.0.0-20160621034901-c1b8fa8bdcce // indirect
 	gopkg.in/yaml.v3 v3.0.0-20200313102051-9f266ea9e77c // indirect
-
 )
 
 replace github.com/ethereum/go-ethereum v1.10.13 => github.com/ethereum-optimism/reference-optimistic-geth v0.0.0-20220107224313-7f6d88bc156a
--- a/opnode/l2/input_derivation.go
+++ b/opnode/l2/input_derivation.go
@@ -1,0 +1,209 @@
+package l2
+
+import (
+	"encoding/binary"
+	"fmt"
+	"math/big"
+
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/core/types"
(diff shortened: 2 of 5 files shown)


protolambda:
@trianglesphere found a bug: we counted one 32 bytes too many in the minimum log unindexed data length. Fixed in 225955b

protolambda:
This reminds me that:
- We can gas-golf the unindexed data part of the deposit-log, to not use `6*32` bytes. (boolean, is unnecessary, data length is unneccessary because we only have a single dynamic length bytes, and data-offset is always 160). cc @maurelian 
- We should add a test that creates a Go EVM instance, and runs test transactions like @trianglesphere did manually, and then try derive deposits with the driver code (and maybe check invariants). One manual case would even catch this. After Go 1.18 we can use the Go fuzzing feature to try many more transactions easily.

protolambda:
9c3fd4e4d92a01d00c73d5c397ddc59c95f6b55d adds a fix to deal with padding of transaction data, thanks to @maurelian for finding this issue.

maurelian:
> We can gas-golf the unindexed data part of the deposit-log, to not use 6*32 bytes. (boolean, is unnecessary, data length is unneccessary because we only have a single dynamic length bytes, and data-offset is always 160). cc @maurelian

Nice observations! 👍

norswap (approved):
Looks good! Didn't go in too much depth since this already went through a few iterations.

protolambda:
Rebased onto `main` to resolve merge conflicts.

PR #129, opened 2022-01-10
ref impl: derive outputs and run driver step

protolambda:

This:
- Implements the retrieval of output full execution-payload when given the payload-attributes (inputs)
- Implements how a L2 chain extends with a block: L1 data is retrieved, the input-derivation is called, the output-derivation is called, the payload is executed, and the forkchoice is updated 

**Depends on #128**

Review: any team.

Regarding testing: this code mostly strings together consecutive API calls, but there is little branching (aside from API error handling). Any suggestions how to best test this? End-to-end for API is verified to work, but no neat unit-test yet.

### DIFF
--- a/opnode/l2/output_derivation.go
+++ b/opnode/l2/output_derivation.go
@@ -1,0 +1,35 @@
+package l2
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/ethereum/go-ethereum/common"
+)
+
+type BlockPreparer interface {
+	GetPayload(ctx context.Context, payloadId PayloadID) (*ExecutionPayload, error)
+	ForkchoiceUpdated(ctx context.Context, state *ForkchoiceState, attr *PayloadAttributes) (ForkchoiceUpdatedResult, error)
+}
+
+// DeriveBlockOutputs uses the engine API to derive a full L2 block from the block inputs.
+// The fcState does not affect the block production, but may inform the engine of finality and head changes to sync towards before block computation.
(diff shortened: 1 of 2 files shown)


opnode/l2/output_derivation.go
```diff
@@ -0,0 +1,35 @@
...
+	"context"
+	"fmt"
+
+	"github.com/ethereum/go-ethereum/common"
+)
+
+type BlockPreparer interface {
+	GetPayload(ctx context.Context, payloadId PayloadID) (*ExecutionPayload, error)
+	ForkchoiceUpdated(ctx context.Context, state *ForkchoiceState, attr *PayloadAttributes) (ForkchoiceUpdatedResult, error)
+}
```
norswap, line 13: What's the rationale for separating `BlockPreparer` from `EngineAPI`?

protolambda: Reducing the scope of things to mock in testing (as outlined in the PR description though, I'm not sure how best to test these consecutive interface calls. Maybe we just use testify mock, and catch the calls?

norswap: But you'll need to mock `ExecutePayload` anyway to test `DriverStep`, so not sure what the advantage is.
(Also, if you merge the two files, the tests can live in a single test file.)

Not sure of the nature of your testing question: can't we just have a mock implementation of `DriverAPI` (with failing implems for `EthBackend` since we don't need them here).

protolambda:
Rebased on main to resolve merge conflicts (`go.mod` was different, then fixed in previous base PR rebase, but then needs to be resolved during merge. Also needed the `7 -> 6 bytes32` and log padding fixes to be resolved).

PR #131, opened 2022-01-10
ref impl: driver agent: respond to head events and stay in sync

protolambda:

This:
- `EngineDriver` is like a binding to a remote engine, tracking to where it's synced.
- `Drive` is a sync process that can be started/shutdown **per remote engine**, and responds to head events from L1, and syncs (happy extend-case, or worst-case sync with reorg) the L2 chain by running a driver-step on the right starting-point. There is only a single active driving process at a time per engine.

**Depends on #130**

Review: any team, systems team preferred.

Spec: Do we want to go-indepth, like back-off strategy and failure-cases of sync in the spec? Or maybe just the "if out of sync, run a driver-step on with L1 block `X_n` and L2 block parent `Y_{n-1}`" summary of driver routine behavior? (we already more or less have the latter)

Testing: this is the most time/event/integration heavy of all of the reference implementation. Only a single event hub, but still challenging to unit-test. It works e2e though. Any suggestions/help how to best test this?

### DIFF
--- a/opnode/l2/driver.go
+++ b/opnode/l2/driver.go
@@ -177,7 +177,7 @@ func (e *EngineDriver) Drive(ctx context.Context, dl Downloader, l1Heads <-chan 
 					continue
 				}
 				e.Log.Debug("finding next sync step, engine syncing", "l2", e.l2Head, "l1", e.l1Head)
-				_, nextRefL1, refL2, err := FindSyncStart(ctx, e.SyncRef, &e.Genesis)
+				nextRefL1, refL2, err := FindSyncStart(ctx, e.SyncRef, &e.Genesis)
 				if err != nil {
 					e.Log.Error("Failed to find sync starting point", "err", err)
 					continue
--- a/opnode/l2/driver.go
+++ b/opnode/l2/driver.go
@@ -1,0 +1,207 @@
+package l2
+
+import (
+	"context"
+	"sync"
+	"time"
+
+	"github.com/ethereum/go-ethereum"
+	"github.com/ethereum/go-ethereum/event"
+
+	"github.com/ethereum-optimism/optimistic-specs/opnode/eth"
+
+	"github.com/ethereum/go-ethereum/log"
+)
+
+type EngineDriver struct {
+	Ctx context.Context
+	Log log.Logger
+	// API bindings to execution engine
+	RPC DriverAPI
+
+	SyncRef SyncReference
+
+	// The current driving force, to shutdown before closing the engine.
+	driveSub ethereum.Subscription
+	// There may only be 1 driver at a time
+	driveLock sync.Mutex
+
+	// Locks the L1 and L2 head changes, to keep a consistent view of the engine
+	headLock sync.RWMutex
+
+	// l1Head tracks the L1 block corresponding to the l2Head
+	l1Head eth.BlockID
+
+	// l2Head tracks the head-block of the engine
+	l2Head eth.BlockID
+
+	// l2Finalized tracks the block the engine can safely regard as irreversible
+	// (a week for disputes, or maybe shorter if we see L1 finalize and take the derived L2 chain up till there)
+	l2Finalized eth.BlockID
+
+	// The L1 block we are syncing towards, may be ahead of l1Head
+	l1Target eth.BlockID
+
+	// Genesis starting point
+	Genesis Genesis
+}
+
+// L1Head returns the block-id (hash and number) of the last L1 block that was derived into the L2 block
+func (e *EngineDriver) L1Head() eth.BlockID {
+	e.headLock.RLock()
+	defer e.headLock.RUnlock()
+	return e.l1Head
+}
+
+// L2Head returns the block-id (hash and number) of the L2 chain head
+func (e *EngineDriver) L2Head() eth.BlockID {
+	e.headLock.RLock()
+	defer e.headLock.RUnlock()
+	return e.l2Head
+}
+
+func (e *EngineDriver) Head() (l1Head eth.BlockID, l2Head eth.BlockID) {
+	e.headLock.RLock()
+	defer e.headLock.RUnlock()
+	return e.l1Head, e.l2Head
+}
+
+func (e *EngineDriver) UpdateHead(l1Head eth.BlockID, l2Head eth.BlockID) {
+	e.headLock.Lock()
+	defer e.headLock.Unlock()
+	e.l1Head = l1Head
+	e.l2Head = l2Head
+}
+
+func (e *EngineDriver) RequestHeadUpdate() error {
+	e.headLock.Lock()
+	defer e.headLock.Unlock()
+	refL1, refL2, _, err := e.SyncRef.RefByL2Num(e.Ctx, nil, &e.Genesis)
+	if err != nil {
+		return err
+	}
+	e.l1Head = refL1
+	e.l2Head = refL2
+	return nil
+}
+
+func (e *EngineDriver) Drive(dl Downloader, l1Heads <-chan eth.HeadSignal) ethereum.Subscription {
+	e.driveLock.Lock()
+	defer e.driveLock.Unlock()
+	if e.driveSub != nil {
+		return e.driveSub
+	}
+	e.driveSub = event.NewSubscription(func(quit <-chan struct{}) error {
+		// keep making many sync steps if we can make sync progress
+		hot := time.Millisecond * 30
+		// check on sync regularly, but prioritize sync triggers with head updates etc.
+		cold := time.Second * 8
+		// at least try every minute to sync, even if things are going well
+		max := time.Minute
+
+		syncTicker := time.NewTicker(cold)
+		defer syncTicker.Stop()
+
+		// backoff sync attempts if we are not making progress
+		backoff := cold
+		syncQuickly := func() {
+			syncTicker.Reset(hot)
+			backoff = cold
+		}
+		// exponential backoff, add 10% each step, up to max.
+		syncBackoff := func() {
+			backoff += backoff / 10
+			if backoff > max {
+				backoff = max
+			}
+			syncTicker.Reset(backoff)
+		}
+
+		l2HeadPollTicker := time.NewTicker(time.Second * 14)
+		defer l2HeadPollTicker.Stop()
+
+		onL2Update := func() {
+			// When we updated L2, we want to continue sync quickly
+			syncQuickly()
+			// And we want to slow down requesting the L2 engine for its head (we just changed it ourselves)
+			// Request head if we don't successfully change it in the next 14 seconds.
+			l2HeadPollTicker.Reset(time.Second * 14)
+		}
+
+		for {
+			select {
+			case <-e.Ctx.Done():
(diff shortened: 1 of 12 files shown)


opnode/l2/driver.go
```diff
@@ -0,0 +1,207 @@
...
+	"github.com/ethereum/go-ethereum"
+	"github.com/ethereum/go-ethereum/event"
+
+	"github.com/ethereum-optimism/optimistic-specs/opnode/eth"
+
+	"github.com/ethereum/go-ethereum/log"
+)
+
+type EngineDriver struct {
+	Ctx context.Context
```
trianglesphere, line 17: I need to understand how this is being used before making a final recommendation, but generally people try to avoid embedding contexts in structs (though there is some disagreement)
https://go.dev/blog/context-and-structs
https://github.com/golang/go/issues/22602

protolambda: The idea is that all background work, like calls to the engine API, can wrap this context, so they don't linger around when we shut down / replace an engine. When we shut down the engine we close this context. (maybe we should just make it a private field, and cancel the context when we `Close` the driver.

protolambda: Alternatively, we can remove it from the struct, and add it to the individual methods. Think I like that better now that you mention it

protolambda: Fixed in 153473aac2f602a0e5e7e524011d7f726ae84726

norswap:
Quick question: this is divergent from staging, I assume this is the canonical version?

opnode/l2/driver.go
```diff
@@ -0,0 +1,208 @@
...
+				return ctx.Err()
+			case <-quit:
+				return nil
+			case <-l2HeadPollTicker.C:
+				ctx, cancel := context.WithTimeout(ctx, time.Second*4)
+				if err := e.RequestHeadUpdate(ctx); err != nil {
+					e.Log.Error("failed to fetch L2 head info", "err", err)
+				}
+				cancel()
+				continue
```
norswap, line 142: Don't we want to verify that the L2 head of the execution engine is either:
- something we're already aware of
- congruent with the referenced L1 block (i.e. valid) (in the case this is a head that was synced by the execution engine that we didn't set ourselves)

opnode/l2/driver.go
```diff
@@ -0,0 +1,208 @@
...
+	if e.driveSub != nil {
+		return e.driveSub
+	}
+	e.driveSub = event.NewSubscription(func(quit <-chan struct{}) error {
+		// keep making many sync steps if we can make sync progress
+		hot := time.Millisecond * 30
+		// check on sync regularly, but prioritize sync triggers with head updates etc.
+		cold := time.Second * 8
+		// at least try every minute to sync, even if things are going well
+		max := time.Minute
```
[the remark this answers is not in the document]
norswap: Another thing that I have some doubts about: how big is the channel in a `Ticker`? If it's only one (meaning only one tick event is waiting at any one time) then it's perfect. If it's bigger, it means a queue of tick events can build up, which doesn't seem ideal given the potential time it takes to process a sync step.

The official doc is delightfully vague on the topic:
> NewTicker returns a new Ticker containing a channel that will send the time on the channel after each tick. The period of the ticks is specified by the duration argument. The ticker will adjust the time interval or drop ticks to make up for slow receivers.

opnode/l2/driver.go
```diff
@@ -0,0 +1,208 @@
...
+	Log log.Logger
+	// API bindings to execution engine
+	RPC DriverAPI
+
+	SyncRef SyncReference
+
+	// The current driving force, to shutdown before closing the engine.
+	driveSub ethereum.Subscription
+	// There may only be 1 driver at a time
+	driveLock sync.Mutex
```
norswap, line 26: Shouldn't the property of one driver per engine be enforced at the point where the `EngineDriver` gets instantiated rather than allowing multiple `EngineDriver` for the same execution engine then using a mutex?

opnode/l2/driver.go
```diff
@@ -0,0 +1,208 @@
...
+	e.headLock.Lock()
+	defer e.headLock.Unlock()
+	refL1, refL2, _, err := e.SyncRef.RefByL2Num(ctx, nil, &e.Genesis)
+	if err != nil {
+		return err
+	}
+	e.l1Head = refL1
+	e.l2Head = refL2
+	return nil
+}
```
norswap, line 85: Do these two updating functions need to be public?

norswap (approved):
Nice! I think the architecture that we conceptually have three concurrent loops for checking l1 head, l2 head and taking a sync step (with the sync loop slowing down each time it doesn't move closer to sync) is pretty neat.

Also something that could use a description. I'm thinking we probably need some markdown implementation notes.

opnode/l2/driver.go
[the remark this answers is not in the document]
trianglesphere: the channel is size 1. If you go to good, and click on the paragraph thing (it only works in chrome for me), it takes you to the source code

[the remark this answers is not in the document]
norswap: Nice trick 👍

opnode/l2/driver.go
```diff
@@ -0,0 +1,208 @@
...
+		}
+
+		for {
+			select {
+			case <-ctx.Done():
+				return ctx.Err()
+			case <-quit:
+				return nil
+			case <-l2HeadPollTicker.C:
+				ctx, cancel := context.WithTimeout(ctx, time.Second*4)
```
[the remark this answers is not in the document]
protolambda: Sounds good to me, agreed on the `ctx` strategy. Not critical for this PR though, merging this and the other work, and then reviewing a consistent set of timeouts/ctx would be faster :+1:

opnode/l2/driver.go
```diff
@@ -0,0 +1,208 @@
...
+					e.Log.Debug("Received L1 head signal, already synced to it, ignoring event", "l1_head", e.l1Head)
+					continue
+				}
+				if e.l1Head == l1HeadSig.Parent {
+					// Simple extend, a linear life is easy
+					if l2ID, err := DriverStep(ctx, e.Log, e.RPC, dl, l1HeadSig.Self, e.l2Head, e.l2Finalized.Hash); err != nil {
+						e.Log.Error("Failed to extend L2 chain with new L1 block", "l1", l1HeadSig.Self, "l2", e.l2Head, "err", err)
+						// Retry sync later
+						e.l1Target = l1HeadSig.Self
+						onL2Update()
```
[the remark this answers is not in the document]
protolambda: Good catch, will separate the L2 head poll adjustment from the sync. Tried to reduce calls earlier, but it got a bit messy.

opnode/l2/driver.go
protolambda: We only need to track what the engine says it has to determine if we can go through the happy sync path. We trust our engine to have L2 blocks with correct L1 information (since we inserted it and set it with forkchoice updates). FindSyncStart can handle any out-of-sync cases.

opnode/l2/driver.go
protolambda: Being able to manage the engine without actively running the driver process will be useful later, e.g.:
- Not running the default driver during state sync
- Stopping/starting the driver for integration tests to force nodes out-of-sync

opnode/l2/driver.go
protolambda: `RequestHeadUpdate` is used after initialization to try start in the happy sync without the driver having to figure it out.

`UpdateHead` is used to default to the genesis if the above `RequestHeadUpdate` fails.

protolambda:
Rebased on base branch

protolambda:
Went back and forth whether or not we need `refL1` in the `FindSyncStart` return params, but we just need a name update to nextRefL1 in the driver.

protolambda:
Rebased on the updated impl-sync-start PR (rebased that on main)

opnode/l2/driver.go
norswap: But if we're making the assumption that we are the one inserting the L2 blocks, what's the purpose of getting the execution engine to tell us about new heads (which we already know about & track).

I guess the point of this is being a stub for later, when we do have L2-level sync, at which point we will need to validate the block, right?

trianglesphere:
> Testing: this is the most time/event/integration heavy of all of the reference implementation. Only a single event hub, but still challenging to unit-test. It works e2e though. Any suggestions/help how to best test this?

My recommendation for this is to make the `Drive` loop as small and self contained as possible. With explicit inputs, it's easy too set up a unit test. Unfortunately having timers does make this harder, but not impossible.

protolambda:
@trianglesphere implemented your suggestion (large refactor, no functional changes, but should pay off in test-ability), can you review?

opnode/l2/driver_state.go
```diff
@@ -0,0 +1,130 @@
...
+		log.Debug("Engine is already synced, aborting sync", "l1_head", e.l1Head, "l2_head", e.l2Head)
+		return false
+	}
+	if l2ID, err := driver.driverStep(ctx, nextRefL1, refL2, e.l2Finalized); err != nil {
+		log.Error("Failed to sync L2 chain with new L1 block", "l1", nextRefL1, "onto_l2", refL2, "err", err)
+		return false
+	} else {
+		e.UpdateHead(nextRefL1, l2ID) // l2ID is derived from the nextRefL1
+	}
+	return e.l1Head != e.l1Target
```
trianglesphere, line 102: I'm a little worried that this doesn't catch up in the case of a sync. The flow as I understand it is

L1 & L2 head on block 10, say L1 Target (or tip) is a block 20
- New L1 Block -> NotifyL1Head fails b/c out of sync, returns false
- Sync ticker fires -> RequestSync advances by one, but still returns false
- Sync Ticker remains cold

Part of it is that I wonder if this sync function should try to advance to head. I think it may be slightly preferred to setting the `syncquickly` correctly b/c it avoids repeated calls to `syncStart` (only changes in `NotifyL1Head`), though maybe we want an interrupt into this function when we get a new L1 head.

Note: I think a very similar bug is present in the pre-refactor form because `syncbackoff` is always called during the sync ticker.

trianglesphere (approved):
Nice work on the refactor. I think pulling it apart like this teases out the fundamental structure and makes it a bit easier to understand.

One comment was for more comments, and the other is about what I think's a bug. Feel free to merge this once you think you've got it resolved.

opnode/l2/driver_state.go
protolambda: > New L1 Block -> NotifyL1Head fails b/c out of sync, returns false

But it does update the `l1Target`

> Sync ticker fires -> RequestSync advances by one, but still returns false

It returns false when the `l1Head == l1Target`: when it's already synced. If it advances by 1, but not synced afterwards, then it's an "update", and it will continue with `syncQuickly()` (faster sync) and `onL2Update` (slow down polling L2 because we are actively changing it).

It's the `!=` I think that's confusing, it returns true in your example, not false. I'll add a comment to clarify.

Always calling `syncBackoff` is fine, it will reset with `syncQuickly` if there's actually something to sync.

PR #132, opened 2022-01-10
ref impl: main node process (connect with RPC, start drivers, handle shutdown)

protolambda:

This:
- Defines the `run` subcommand, a.k.a. `go run ./opnode/cmd run`
- Connects with L1 and L2 RPC connections (or at least creates client binding, initial connection may be broken)
- Initializes everything
- Runs node process
- Handles `Close` of the node
- readme how to run the reference implementation, explaining the options

~~**Depends on #131**~~ (merged)

Review: systems team preferred. No spec affected.


protolambda:
TODO: rebase on `main` once other PRs are merged. And adapt to minor interface changes. E.g. `ctx` addition to `Drive` method and removal from `EngineDriver` struct (153473aac2f602a0e5e7e524011d7f726ae84726)

opnode/node/node.go
```diff
@@ -1,18 +1,236 @@
...
+	go c.RunNode()
+
 	return nil
 }
 
+func (c *OpNodeCmd) RunNode() {
+	c.log.Info("Starting OpNode")
+
+	var unsub []func()
+	mergeSub := func(sub ethereum.Subscription, errMsg string) {
```
trianglesphere, line 150: rename to handle unsubscribe? My first impression was that this was supposed to merge subscriptions into a single stream

opnode/node/node.go
```diff
@@ -1,18 +1,236 @@
...
+	c.l1Downloader = l1.NewDownloader(c.l1Source)
+
+	for i, addr := range c.L2EngineAddrs {
+		// L2 exec engine: updated by this OpNode (L2 consensus layer node)
+		backend, err := rpc.DialContext(ctx, addr)
+		if err != nil {
+			if backend == nil {
+				return fmt.Errorf("failed to dial L2 address %d (%s): %v", i, addr, err)
+			}
+			c.log.Warn("failed to dial L2 address, but may connect later", "i", i, "addr", addr, "err", err)
```
norswap, line 115: How does this work, do we need to call `DialContext` again or is this done implicitly in some other call?

opnode/node/node.go
```diff
@@ -1,18 +1,236 @@
...
+		engDriveSub := eng.Drive(c.l1Downloader, l1SubCh)
+		mergeSub(engDriveSub, "engine driver unexpectedly failed")
+	}
+
+	// Keep subscribed to the L1 heads, which keeps the L1 maintainer pointing to the best headers to sync
+	l1HeadsSub := event.ResubscribeErr(time.Second*10, func(ctx context.Context, err error) (event.Subscription, error) {
+		if err != nil {
+			c.log.Warn("resubscribing after failed L1 subscription", "err", err)
+		}
+		return eth.WatchHeadChanges(c.ctx, c.l1Source, eth.HeadSignalFn(func(sig eth.HeadSignal) {
```
norswap, line 192: The type conversion here is redundant.

norswap (approved):
LGTM!

opnode/README.md
```diff
@@ -0,0 +1,38 @@
...
+Specify genesis details:
+
+- L1 number / hash: starting-point of L2 chain inputs
+- L2 genesis hash: to confirm we are building on the correct L2 genesis
+
+```shell
+# websockets or IPC preferred for event notifications to improve sync, http RPC works with adaptive polling.
+op node run \
+  --l1=ws://localhost:8546 --l2=ws//localhost:9001 \
+  --genesis.l1-num=.... --genesis.l1-hash=..... --genesis.l2-hash=....
```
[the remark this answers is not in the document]
protolambda: Parsing is flexible, the 0x is optional

protolambda:
Rebased on top of the driver PR (all dependencies rebased on `main`)

opnode/node/node.go
protolambda: See comment above: "HTTP or WS RPC may create a disconnected client, RPC over IPC may fail directly"

E.g. if the HTTP URL doesn't parse, you may get a nil client + error. But if it simply fails to connect, then you get a non-nil client + error warning about the failed connection. The client will just re-attempt to connect when used later.

opnode/node/node.go
```diff
@@ -1,18 +1,237 @@
...
+
+	l1Downloader l1.Downloader
+
+	ctx   context.Context
+	close chan chan error
+}
+
+func (c *OpNodeCmd) Default() {
+	c.L1NodeAddrs = []string{"http://127.0.0.1:8545"}
+	c.L2EngineAddrs = []string{"http://127.0.0.1:8551"}
```
trianglesphere, line 66: Not sure what's standardized yet, but you used port `9000` in the rollup node experiment repo
```suggestion
	c.L2EngineAddrs = []string{"http://127.0.0.1:9000"}
```

protolambda: Good catch, but either works I think. 8551 is the standard engine-API port. It really depends where the engine is activated or not to avoid conflicts (L1 might also have the engine port open). Maybe we should just specify a standard port for Optimism engine API usage. Leaving this to the default port for now (assuming we're testing on non-merged L1 testnets)

protolambda:
Rebased on `main` + made minor changes to adapt to the separated state/actions of the driver in PR 131

PR #140, opened 2022-01-19
Fix Deposit Feed test

maurelian:
**Description**
1. Adds a return value check in the Dummy contract used for testing Deposits made by a contract.
2. This led me to identify a test that was passing incorrectly.
3. And I added two more test cases that were lacking.

### DIFF
--- a/packages/contracts/test/L1/DepositFeed.spec.ts
+++ b/packages/contracts/test/L1/DepositFeed.spec.ts
@@ -213,7 +213,6 @@ describe('DepositFeed', () => {
         // Deploy a dummy contract so we can impersonate it
         const dummy = await (await ethers.getContractFactory('Dummy')).deploy()
         await dummy.deployed()
-        // this is not emitting an event!
         await expect(
           dummy.forward(
             depositFeed.address,
--- a/packages/contracts/test/L1/DepositFeed.spec.ts
+++ b/packages/contracts/test/L1/DepositFeed.spec.ts
@@ -26,7 +26,6 @@ const decodeDepositEvent = async (
   const events = await depositFeed.queryFilter(
     depositFeed.filters.TransactionDeposited()
   )
-  console.log(events.length);
 
   const eventArgs = events[events.length - 1].args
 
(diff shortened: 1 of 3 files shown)


packages/contracts/test/L1/DepositFeed.spec.ts
```diff
@@ -164,18 +209,55 @@ describe('DepositFeed', () => {
         })
       })
 
+      it('when a contract deposits a transaction with an ETH value.', async () => {
```
protolambda, line 212: The dummy contract forwards ETH, as opposed to spending existing balance. During execution it's effectively the same (contract balance increases and decreases), but do we need to test this the other way?

packages/contracts/test/L1/DepositFeed.spec.ts
```diff
@@ -164,18 +209,55 @@ describe('DepositFeed', () => {
...
+        // Deploy a dummy contract so we can impersonate it
+        const dummy = await (await ethers.getContractFactory('Dummy')).deploy()
+        await dummy.deployed()
+        // this is not emitting an event!
```
protolambda, line 216: It does decode an awaited event below however, is this comment wrong?

protolambda (commented):
Wondering about one comment, and if there's a better way to test the contract paying the deposited ETH. Looks good otherwise :+1:

packages/contracts/test/L1/DepositFeed.spec.ts
maurelian: my bad, this was a debugging comment that got left in.

protolambda: Removed the old comment in 55bf74ba0b9a48d536b98bdfff8c4d23d8a10b1d

protolambda (approved):
LGTM!

packages/contracts/test/L1/DepositFeed.spec.ts
maurelian: I think it's fine. There's zero difference in the execution context of the DepositFeed right?

Issue #106, opened 2022-01-03
Rework execution engine specs to match the implementation


norswap:
Our execution engine implementation is at https://github.com/ethereum-optimism/reference-optimistic-geth/

(I would like to rename this repository. Possible to `execution-engine`. Is that okay? If so, I'll open another issue to track.)

We need to update the spec to reflect the implementation, and improve the presentation somewhat.

I also want to move the sync conversation to elsewhere. Synchronization does not happen in the execution engine but in the rollup node. However, it is a thorny subject, which is not ready for specification yet. It is, in reality, an implementation concern, but we should write about how it can work in practice. I think this deserves its own document.

protolambda:
> (I would like to rename this repository. Possible to execution-engine. Is that okay? If so, I'll open another issue to track.)

For now I would rather not change names, or at least preserve the "geth" part of the name. Other L1 clients are also execution-engines, so it's confusing. Erigon, also in Go, made progress with implementing the Merge during holidays, so maybe we can give it an attempt to port over the diff for additional engine support :)

> It is, in reality, an implementation concern, but we should write about how it can work in practice. I think this deserves its own document.

Resolving the difference between L1 and L2 chain, and syncing back to head, is indeed worth specifying more. Implementation showed some topics that deserve more focus: starting at L2 engine history, watching L1/L2 head changes to adapt sync, the genesis edge-case (first L2 block is not derived from L1 but part of chain-config), and dealing with engine errors or L1 errors.
Once the ref-impl is polished/tested more I can work on this doc, with pointers to the ref-impl.

norswap:
> For now I would rather not change names, or at least preserve the "geth" part of the name. Other L1 clients are also execution-engines, so it's confusing. Erigon, also in Go, made progress with implementing the Merge during holidays, so maybe we can give it an attempt to port over the diff for additional engine support :)

Does not have to be now, and agree that keeping "geth" in the name makes a lot of sense!

maurelian:
closed by #144

norswap:
I apologize, made I mistake in the description of #144 (now edited away). The PR does not actually close this one (this one is about the execution engine, while #144 is about the rollup node).

Issue #74, opened 2021-12-13
Doc: Block-derivation pseudo code for rollup node spec


maurelian:
Transferred from ENG-1737

EDIT(norswap): The goal is to add some pseudo-code to the rollup-node spec to give a more digestible overview of the changes.

maurelian:
Closed by #144

Issue #141, opened 2022-01-20
Implementation Notes: Sync Algorithm


norswap:
We should have some kind of writing markdown artifact that explains the sync algorithm in-practice.

This can be separate from "the specification" — to specify the correct behaviour we do not need to really specify how sync proceeds.
We currently have stuff about sync in the rollup driver spec that needs to be moved out.

I propose creating another top-level directory called `implem_notes` (or something else) to put markdown files that document the reference implementation. Thoughts?

PR #144, opened 2022-01-21
Rollup Node Specification Revamp

### LINKED ISSUE #87: Standardize language around deposits
We have a lot of things using the term "
deposit", which are similar but different.

- Deposit
- Deposit transaction
- Deposit transaction type
- L1 Transaction Deposits (Transaction)
- L1 Attributes Deposits (Transaction)

It's fine for now, but this will get confusing eventually.

norswap:
This started as a pull-request to address #87 (standardize languages around deposits), but due to a small communication problem, I was faced with a pretty big merge conflict, which also lead to some interesting discussions with @protolambda about how to structure this spec.

Since I was going to introduce a huge unreadable diff anyway, I've also used to opportunity to knock off #74 (derivation pseudocode) and #106 (sync spec with implementation, though proto had pretty much taken care of that already).

**The best way to review the changes to the rollup node specs is to use [this links](https://github.com/ethereum-optimism/optimistic-specs/blob/a9087ce53d6953057c36d5dec741cdff603d659a/specs/rollup-node.md) to visualize it rendered directly.**

So this now closes #87, #74 and #106.


norswap:
@protolambda I tried to synthesize our approaches to make the [rollup node spec](https://github.com/ethereum-optimism/optimistic-specs/blob/a9087ce53d6953057c36d5dec741cdff603d659a/specs/rollup-node.md) both more friendly to newcomers & more concise. Let me know if you have any feedback!

specs/rollup-node.md
```diff
@@ -175,58 +235,77 @@ The following scenarios are assimilated to errors:
...
+    - It is necessary to ensure that no L1 re-org occured during this lookup, i.e. that `nextRefL1.parent == refL1`.
+    - If the next L1 block does not exist yet, there is no re-org, and nothing new to derive, and we can abort the
+        process.
+- Otherwise, if `refL2` is the L2 genesis block, we have re-orged past the genesis block, which is an error that
+    requires a re-genesis of the L2 chain to fix (i.e. creating a new genesis configuration).
+- Otherwise, if either `currentL1` does not exist, or `currentL1 != refL1`, set `refL2` to `parentL2` and restart this
+    algorithm from step 2.
+  - Note: if `currentL1` does not exist, it means we are in a re-org to a shorter L1 chain.
+  - Note: as an optimization, we can cache `currentL1` and reuse it as the next value of `nextRefL1` to avoid an
+        extra lookup.
```
norswap, line 306: I've tried simplifying this as much as possible while preserve the semantics. @protolambda could you look it over to see if it seems alright?

specs/rollup-node.md
```diff
@@ -175,58 +235,77 @@ The following scenarios are assimilated to errors:
...
+3. Fetch `currentL1`, the canonical L1 block at the same height as `refL1`.
+
+- If `currentL1 == refL1`, then `refL2` was built on a canonical L1 block:
+  - Find the next L1 block (it may not exist yet) and return `(refL2, nextRefL1)` as the starting point of the
+    re-derivation.
+    - It is necessary to ensure that no L1 re-org occured during this lookup, i.e. that `nextRefL1.parent == refL1`.
+    - If the next L1 block does not exist yet, there is no re-org, and nothing new to derive, and we can abort the
+        process.
+- Otherwise, if `refL2` is the L2 genesis block, we have re-orged past the genesis block, which is an error that
+    requires a re-genesis of the L2 chain to fix (i.e. creating a new genesis configuration).
```
norswap, line 301: I'm not 100% sure about this, but this my best guess. What happens in practice if we re-org past genesis? Wouldn't it be better to specify that a re-org past genesis is allowed, in which case we just wait for the L2 chain inception L1 block height to restart the whole process?

specs/rollup-node.md
```diff
@@ -175,58 +235,77 @@ The following scenarios are assimilated to errors:
...
+previous section to each successive L1 block until we have caught up with the L1 head.
 
-As already alluded to in the section on [interacting with the execution engine][calling-exec-engine], an L2 block is
-considered *finalized* after a delay of `FINALIZATION_DELAY_BLOCKS == 50400` blocks after the L1 block that generated
-it. This is a duration of approximately 7 days worth of L1 blocks.
+> **TODO** specify genesis block
 
-L1 Ethereum [reaches finality approximately every 12 minutes][l1-finality], so these L2 blocks can safely be considered
-to be final: they will never disappear from the chain's history because of a re-org.
+# Handling L1 Re-Orgs
```
norswap, line 266: I want to move this to a implementation notes directory, cf. https://github.com/ethereum-optimism/optimistic-specs/issues/141

Any objections?

specs/rollup-node.md
```diff
@@ -175,58 +235,77 @@ The following scenarios are assimilated to errors:
 - [`engine_executePayloadV1`] returning a `status` of `"SYNCING"` or `"INVALID"` whenever passed an execution payload
   that was obtained by a previous call to [`engine_getPayloadV1`].
 
-## Handling L1 Re-Orgs
+### Finalization Guarantees
```
norswap, line 239: I added mentions to fault proofs here because this section makes little sense without it. I'm also okay to remove this this section altogether to preserve consistency (though doing so make the prescriptions on the `forkchoiceState` seemingly arbitrary).

specs/deposits.md
```diff
@@ -185,23 +205,21 @@ contract][deposit-feed-contract] on L1.
...
 
-The deposit feed handles two special cases:
+The deposit contract handles two special cases:
 
 1. A contract creation deposit, which is indicated by setting the `isCreation` flag to `true`.
    In the event that the `to` address is non-zero, the contract will revert.
 2. A call from a contract account, in which case the `from` value is transformed to its L2
    [alias][address-aliasing].
 
-> **TODO** Define if/how ETH withdrawals occur.
```
norswap, line 202: Removed since it's out of scope for the deposit rollup.

specs/deposits.md
```diff
@@ -44,41 +69,36 @@ Picking a high identifier minimizes the risk that the identifier will be used be
...
+   by
+2. an array of zero-or-more [user-deposited transactions][user-deposited] submitted to the deposit
+   feed contract on L1.
 
-The rationale for creating only one new transaction type is to minimize both
-modifications to L1 client software and complexity in general.
+We only define a single new transaction type in order to minimize modifications to L1 client
+software, and complexity in general.
 
-> **TODO** Specify and link to deposit blocks
```
norswap, line 59: This is defined in the rollup node, and already implicitly in this file (first tx is l1 attributes, then user-deposited), so I don't think it's necessary to add something here.

Also we link to the glossary.

specs/deposits.md
```diff
@@ -44,41 +69,36 @@ Picking a high identifier minimizes the risk that the identifier will be used be
...
-processed will only derive transactions with a `from` address attested to by the logs of the [L1
+As noted above, the deposited transaction type does not include a signature for validation. Rather,
+authorization is handled by the [L2 chain derivation][g-derivation] process, which when correctly
+applied will only derive transactions with a `from` address attested to by the logs of the [L1
 deposit feed contract][deposit-feed-contract].
 
-In the event a deposit transaction is included which is not derived by the [execution
-engine][execution-engine] using the correct derivation algorithm, the resulting state transition
-would be invalid.
-
```
norswap, line 73: Removed this part because it was erroneous (it's not about the execution engine but the rollup node) and also because we can't really speak to the consequences at this stage (since we don't submit roots on L1) - what does it mean to be invalid? A bad transaction being included can also not lead to any state transition (i.e. the only bad output is in the block hash, but the state root is correct). Also if you don't follow the prescribed rules, then the result is invalid, which is sort of tautological.

specs/deposits.md
```diff
@@ -100,46 +120,46 @@ tooling (such as wallets and block explorers).
...
 4. `value` is `0`
 5. `gasLimit` is set to the maximum available.
-6. `data` is an [ABI] encoded call to the [L1 attributes predeploy][predeploy] contract's
+6. `data` is an [ABI] encoded call to the [L1 attributes predeployed contract][predeploy]'s
    `setL1BlockValues()` function with correct values associated with the corresponding L1 block (cf.
-   [reference implementation][l1-attrib-ref-implem]).
+   [reference implementation][l1-attr-ref-implem]).
 
-   <!-- TODO: Define how this account pays gas on these transactions. -->
+No gas is paid for L1 attributes deposited transactions.
```
norswap, line 143: Maybe change this to say that "gas" is paid for on L1? Should be done when we transcribe the deposit fee design in the spec.

meta/markdown-style.md
```diff
@@ -0,0 +1,42 @@
...
+    - exceptions: where it fits neatly on a single line, in particular in lists of links
+- Excepted for internal and glossary links (see below), add the link reference definition directly
+  after the paragraph where the link first appears.
+
+### Glossary
+
+- Use links to the [glossary] liberally.
+- Include the references to all the glossary links at the top of the file, under the top-level
+  title.
+- A glossary link reference should start with the `g-` prefix. This enables to see what links to the
```
protolambda, line 30: good idea, I was having trouble with this before

specs/deposits.md
```diff
@@ -1,25 +1,50 @@
...
+
+[Deposited transactions][g-deposited], also known as [deposits][g-deposits] are transactions which
+are initiated on L1, and executed on L2. This document outlines a new [transaction
+type][g-transaction-type] for deposits. It also describes how deposits are initiated on L1, along
+with the authorization and validation conditions on L2.
+
+**Vocabulary note**: *deposited transaction* refers specifically to an L2 transaction, while
+*deposit* can refer to the transaction at various stages (for instance when it is deposited on L1).
+
+## Table of Contents
```
protolambda, line 21: Good to see the ToC update, I will automate this with doctoc in a new PR :+1:

specs/deposits.md
protolambda: The "valid"/"invalid" refers to following the verifier rules I think. The local rollup-node can insert any deposit into their execution-engine, but if it does not originate from the specified derivation, other nodes will regard it as invalid. You're right that while we don't have block propagation or state-root posting it's not applicable (as other nodes cannot consume the invalid data in the first place).

specs/deposits.md
protolambda: Technically it's not paid for on L1 either, it's the rollup itself as system inserting the deposit. The rate-limit is simply the L1 block production.

specs/rollup-node.md
protolambda: If the L1 reorgs past expected anchor point on L1 then we should just manually redeploy the L2 chain (new L2 genesis) and not build any blocks on top. Wiping the whole L2 chain with a reorg without intervention is pretty severe.

If we start a L2 chain we can probably pick a finalized/confirmed L1 anchor point in the past, to not risk a reorg, at the cost of just a few empty deposit-blocks at the start of the L2 chain.

specs/rollup-node.md
protolambda: New version looks good!

specs/rollup-node.md
```diff
@@ -1,167 +1,227 @@
...
 
-[`engine_getPayloadV1`]: exec-engine.md#engine_executepayloadv1
-[`ExecutionPayloadV1`]: https://github.com/ethereum/execution-apis/blob/main/src/engine/specification.md#executionpayloadv1
+- `refL2` be the (hash of) the current L2 chain head
+- `refL1` be the (hash of) the L1 block from which `refL2` was derived
+- `payloadAttributes` be some previously derived [payload attributes] for the L1 block with number `l1Number(refL1) + 1`
+
+Then we can apply the following pseudocode logic to update the state of both the rollup driver and execution engine:
+
+```pseudocode
```
protolambda, line 153: This is fine with me, but there's no syntax highlighting so it may be difficult to read. If Go is too verbose for snippets in the spec, we can try python.

protolambda (approved):
Looks great! Not sure about the pseudocode style/language, but ok to defer to when we have more code and can inform that style/language choice better.

specs/deposits.md
```diff
@@ -100,46 +120,46 @@ tooling (such as wallets and block explorers).
...
+2. The L1 attributes predeployed contract
 
 ### L1 Attributes Depositor Account
 
 [depositor-account]: #l1-attributes-depositor-account
 
-The depositor account is an EOA with no known private key. It has the address
+The depositor account is an [EOA][g-eoa] with no known private key. It has the address
 `0xdeaddeaddeaddeaddeaddeaddeaddeaddead0001`. Its value is returned by the `CALLER` and `ORIGIN`
 opcodes during execution of the L1 attributes deposit transaction.
```
maurelian, line 158:
```suggestion
opcodes during execution of the L1 attributes deposited transaction.
```

specs/rollup-node.md
maurelian: Agree with Python, it seems to be more or less the standard for pseudocode in ethereum

specs/rollup-node.md
```diff
@@ -1,167 +1,227 @@
...
 
-- `headBlockHash`: block hash of the last block of the L2 chain, according to the rollup driver.
-- `safeBlockHash`: same as `headBlockHash`.
-- `finalizedBlockHash`: the hash of the block whose number is `number(headBlockHash) - FINALIZATION_DELAY_BLOCKS` if
-  the number of that block is `>= L2_CHAIN_INCEPTION`, 0 otherwise (where `FINALIZATION_DELAY_BLOCKS == 50400`
-  (approximately 7 days worth of L1 blocks) and `L2_CHAIN_INCEPTION` is the [L2 chain inception] (the number of the
-  first L1 block for which an L2 block was produced). See the [Finalization Guarantees][finalization] section for more
-  details.
+Once the [payload attributes] for a given L1 block `B` have been built, and if we have already derived an L2 block from
+`B`'s parent block, then we can use the payload attributes to derive a new L2 block.
```
maurelian, line 141: This 'and if' really stands out to me. What if we haven't?

specs/rollup-node.md
norswap: I'll try using `javascript` since we use JSON notation at the top, I think it should highlight well then.

specs/rollup-node.md
norswap: Good point, I'll add a footnote saying just that.

maurelian (approved):
A few suggestions, but LGTM

specs/rollup-node.md
norswap: Then we can't use the payload attribute to derive a new L2 block :)
Feel free to suggest a clarification if needed!

specs/rollup-node.md
norswap: Wrote my answer before seeing maurelian's. Feel free to suggest a modification to the syntax if you feel like it's warranted. I personally don't really see the need to change anything here.

norswap:
Okay to merge on my end unless they are further suggestions!

PR #152, opened 2022-01-25
ref impl: reduce command running boilerplate

protolambda:
The start, command parsing, shutdown handling, etc. is a lot of boilerplate that is common between all usages of `ask`, so I moved it into the library as convenience method.


trianglesphere (approved):
Nice, I love to see the red.

norswap (approved):

PR #154, opened 2022-01-25
End to End Test

trianglesphere:
This PR adds two new packages
1. Add a `contracts` package - these are go binding to the two contracts that we have
2. Add a `test` package that performs an end to end test of the system.

The end to end test does the following
1. Setup a L1 Geth node (in memory) with the deposit feed contract, clique signing, a premine, and unlocked accounts.
2. Setup a L2 Geth node (in memory) with the L1Block Info contract
3. Configure the rollup node to point to the L1 and L2 Geth nodes (over http or ws).
4. Run the entire system.
5. Run a Deposit transaction against L1 Geth and verifies that the mint shows up in L2 Geth.

The test runs in about 4 seconds. (Could be lower if I can get the rollup node to handle 1 second block times.)


opnode/contracts/l1block/l1_block_info_raw.go
```diff
@@ -0,0 +1,356 @@
+// Code generated - DO NOT EDIT.
```
protolambda, line 1: Can you add a go-generate statement with the abigen command so we can keep the bindings up-to-date? (not here but in the package it is used in, or maybe the `contracts` package in a `doc.go` or something). As well as for the other contract bindings?

trianglesphere: I can add a script/go-generate to get this working.
My understanding is that the `go:generate` command would go in these files.

trianglesphere: Ok, I figured out how to do this (in `doc.go`)

opnode/contracts/doc.go
```diff
@@ -0,0 +1,14 @@
...
+
+
+The bindings are generated with `abigen`. `jq` is required to parse out the ABI section
+from the hardhat artifacts.
+
+*/
+
+package contracts
+
+//go:generate sh -c "jq .abi ../../packages/contracts/artifacts/contracts/L1/DepositFeed.sol/DepositFeed.json | abigen --pkg deposit --out deposit/deposit_feed_raw.go --abi -"
```
tynes, line 13: You can pass in the bytecode like this https://github.com/ethereum-optimism/optimism/blob/0cc660d3b2cf78827d008ea01b51f54fb6e5a708/go/gas-oracle/Makefile#L31 to avoid needing to hardcode in the bytecode in the other file. This will make the binary available on a global like so: https://github.com/ethereum-optimism/optimism/blob/0cc660d3b2cf78827d008ea01b51f54fb6e5a708/go/gas-oracle/bindings/gaspriceoracle.go#L34

opnode/test/setup.go
```diff
@@ -0,0 +1,161 @@
...
+	}
+
+	l1Alloc[common.HexToAddress(cfg.depositContractAddress)] = core.GenesisAccount{Code: common.FromHex(depositContractByteCode), Balance: common.Big0}
+	l2Alloc[common.HexToAddress(cfg.l1InforPredeployAddress)] = core.GenesisAccount{Code: common.FromHex(l1InfoContractByteCode), Balance: common.Big0}
+
+	genesisTimestamp := uint64(time.Now().Unix())
+
+	l1Genesis := &core.Genesis{
+		Config: &params.ChainConfig{
+			ChainID:             big.NewInt(int64(cfg.l1.ethConfig.NetworkId)),
```
tynes, line 110: The config for the chain id is based on the network id? Also can use `new(big.Int).SetUint64` to prevent the need to cast to `int64`

opnode/contracts/doc.go
trianglesphere: Nice, I copied the makefile here and it's way nicer.

trianglesphere:
I like the new ABI gen method better, but I need to change it so it also has the deployed byte code and place that directly into the genesis.

norswap (approved):
This is great!
All my comments are the biggest nits.
Was able to run the tests locally without a hitch 👍

opnode/test/geth.go
```diff
@@ -0,0 +1,77 @@
...
+	}
+
+	keydir := n.KeyStoreDir()
+	scryptN := keystore.LightScryptN
+	scryptP := keystore.LightScryptP
+	n.AccountManager().AddBackend(keystore.NewKeyStore(keydir, scryptN, scryptP))
+	ks := n.AccountManager().Backends(keystore.KeyStoreType)[0].(*keystore.KeyStore)
+
+	for _, pk := range privateKeys {
+		act, err := ks.ImportECDSA(pk, "foobar")
```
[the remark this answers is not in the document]
trianglesphere: It's the password associated with each account. I'll pull it out to a constant, but given we hardcode the mnemonic, I don't see a need to protect it.

PR #182, opened 2022-02-01
L2 output root construction spec

### LINKED ISSUE #168: Spec improved merkleized structure of data to commit to (instead of directly to the state-root)

protolambda:
Fixes #168

Defines the continued block derivation: the L2 output construction, primarily the merkle-structure for the L2 output root.

This does not define the "proposer" job of submitting the root to L1 yet, that overlaps with the smart-contract and fault-proof work and would be better for a separate PR.


specs/rollup-node.md
```diff
@@ -251,6 +253,46 @@ However, they can still be challenged by a fault proof until the end of the faul
 
 > **TODO** the spec doesn't encode the notion of fault proof yet, revisit this (and include links) when it does
 
+## From L2 Block to L2 Output Root
+
+After processing a block the resulting outputs will need to be synchronized with L1 for trustless execution of
+L2-to-L1 messaging, such as withdrawals. To synchronize outputs are merkleized: hashed in a structured form for minimal
+proof cost to any piece of data. The merkle-structure is defined with [SSZ], a type system for merkleization and
+serialization, used in L1 (beacon-chain). However, we replace `sha256` with `keccak256` to save gas costs in the EVM.
```
trianglesphere, line 261: How easy is it to maintain our own version of SSZ?

specs/rollup-node.md
```diff
@@ -251,6 +253,46 @@ However, they can still be challenged by a fault proof until the end of the faul
 
 > **TODO** the spec doesn't encode the notion of fault proof yet, revisit this (and include links) when it does
 
+## From L2 Block to L2 Output Root
+
+After processing a block the resulting outputs will need to be synchronized with L1 for trustless execution of
+L2-to-L1 messaging, such as withdrawals. To synchronize outputs are merkleized: hashed in a structured form for minimal
```
trianglesphere, line 259: "To synchronize outputs are merkleized" is a little odd phrasing.

specs/rollup-node.md
```diff
@@ -251,6 +253,46 @@ However, they can still be challenged by a fault proof until the end of the faul
...
+
+[ExecutionPayload]: https://github.com/ethereum/consensus-specs/blob/dev/specs/bellatrix/beacon-chain.md#executionpayload
+
+The `history_accumulator_root` is a reserved field, elevating a storage variable of the L2 chain that maintains
+the SSZ merkle root of an append-only `List[Bytes32, MAX_ITEM_COUNT]` (`keccak256` SSZ),
+where each item is defined as `keccak256(l2_block_hash ++ l2_state_root)`, one per block of the L2 chain.
+While reserved, a zeroed `Bytes32` is used instead.
+This is a work-in-progress, see [issue 181](https://github.com/ethereum-optimism/optimistic-specs/issues/181).
+
+The `extension` is a zeroed `Bytes32`, to be substituted with a SSZ container to extend merkleized information in future
```
maurelian, line 293: Does this limit us to a single additional 32 byte property in the future?

specs/rollup-node.md
```diff
@@ -251,6 +253,46 @@ However, they can still be challenged by a fault proof until the end of the faul
...
+one can directly access the MPT storage trie root, thus reducing the verification cost of withdrawals on L1.
+
+The `latest_block` is an execution-layer block of L2, represented as the [`ExecutionPayload`][ExecutionPayload] SSZ type
+defined in L1. There may be multiple blocks per L2 output root, only the latest is presented.
+
+[ExecutionPayload]: https://github.com/ethereum/consensus-specs/blob/dev/specs/bellatrix/beacon-chain.md#executionpayload
+
+The `history_accumulator_root` is a reserved field, elevating a storage variable of the L2 chain that maintains
+the SSZ merkle root of an append-only `List[Bytes32, MAX_ITEM_COUNT]` (`keccak256` SSZ),
+where each item is defined as `keccak256(l2_block_hash ++ l2_state_root)`, one per block of the L2 chain.
```
maurelian, line 289: `++` means concatenation?

specs/rollup-node.md
```diff
@@ -251,6 +253,46 @@ However, they can still be challenged by a fault proof until the end of the faul
...
+  extension: Bytes32
+```
+
+The `state_root` is the Merkle-Patricia-Trie ([MPT][g-mpt]) root of all execution-layer accounts,
+also found in `latest_block.state_root`: this field is frequently used and thus elevated closer to the L2 output root,
+reducing the merkle proof depth and thus the cost of usage.
+
+The `withdrawal_storage_root` elevates the Merkle-Patricia-Trie ([MPT][g-mpt]) root of L2 Withdrawal contract storage.
+Instead of a MPT proof to the Withdrawal contract account in the account trie,
+one can directly access the MPT storage trie root, thus reducing the verification cost of withdrawals on L1.
```
maurelian, line 280: I don't think we've 100% committed to doing this work yet.

protolambda:
Here is some napkin math to clarify the reasoning for the elevated withdrawal storage trie root:

Number of accounts on L1: 186000000 (see https://etherscan.io/chart/address)
Number of accounts on L2: 252000 (see https://optimistic.etherscan.io/chart/address)

Number of cross domain messages: ~ 8500
https://etherscan.io/address/0x25ace71c97B33Cc4729CF772ae268934F7ab5fA1

Number of standard bridge transactions: ~ 155000
Number of standard bridge internal transactions: 5525 (one per finalized withdrawal by cross domain messenger AFAIK)
https://etherscan.io/address/0x99C9fc46f92E8a1c0deC1b1747d010903E884bE1#internaltx


Regarding the withdrawal trie cost:
```
# Approx L1 account trie layers:
log(account_trie_size) / log(16) = log(186000000) / log(16) = 6.867 = ~ 7

# Approx L2 account trie layers:
log(252000) / log(16) = 4.485 = ~ 5

# Approx withdrawal storage trie layers (assuming 1 slot per withdrawal)
log(5525) / log(16) = 3.108 = ~ 4

# withdrawal storage trie layers with 100x as many withdrawals
log(552500) / log(16) = 4.769 = ~ 5

# Approx withdrawal storage trie layers, if there were actually as many withdrawals as deposits:
log(155000) / log(16) = 4.310 = ~ 5


# cost per binary tree layer
EVM.codes sha3 with 64 byte input: 48 gas
For 4 layers (to match MPT layer): 4 * 48 = 192 gas

# cost per MPT layer
EVM.codes sha3 with 17 * 32 = 544 byte input: 183 gas, but with more extra complexity, so likely significantly higher
```

Reduction of removing the account trie work currently: `5 / (5 + 4) = 56%`

Reduction of removing the account trie, if we had as many accounts as L1, and 100 times as many withdrawals: `7 / (7 + 5) = 58%`

If MPT is really that much worse than binary trie, 30% for complexity, the reduction to swap withdrawal storage to binary trie would be: `(1 - (192 / (183*1.3))) * (4 / (5 + 4)) = 8.6 %`

I'm going for the ~ 57 % cost reduction, and leave the remaining 0 - 10% for later, after sequencing work (or someone else can gas golf it and swap for a binary trie). This also means we can short-cut the first part of the withdrawal proof to state-root, and leave the remaining part all the same: low complexity to implement.

specs/rollup-node.md
protolambda: yes

specs/rollup-node.md
protolambda: To swap the hash-function, you need to change two things in most implementations:
- the `hasher` object you pass as argument (it's an argument, and not a function call, since it is re-used a lot during tree hashing, like a scratch-pad)
- the `zero_hashes` (`H_1 = H(0, 0), H_2 = H(H_1, H_1), ...`) need to be initialized with the desired hash-function

These are pretty simple changes, and they seem worth the gas-cost savings to me.

trianglesphere:
If you want to go one step more accurate (doesn't change too much), the MPT average depth is `log_16(keys) + ~1.7`

norswap:
Flagging that I'd like to give this a review, but it might only happen early next week.

specs/rollup-node.md
protolambda: No, we can replace it with a container like:
```python
class Extension(Container):
  new_thing: foobar
  more_new_things: List[example, 123]
  extension_again: Bytes32
```
This adds new fields as well as a deeper extension slot.
The upper tree stays the same, but the sub-tree of the extension expands to contain the new things.

specs/rollup-node.md
protolambda: https://github.com/ethereum-optimism/optimistic-specs/pull/182#issuecomment-1029296578 explains the reasoning, we can further optimize later for the last 10%, and the current change is fully backwards compatible (with pre 1.0 contracts).

maurelian (approved):
🚢

maurelian:
FWIW, although I've approved, I think we should respect @norswap's request

> Flagging that I'd like to give this a review, but it might only happen early next week.

specs/glossary.md
```diff
@@ -72,12 +72,12 @@ block, and output block properties, which are derived after executing the block'
 
 "Externally Owned Account", an Ethereum term to designate addresses operated by users, as opposed to contract addresses.
 
-## Merkle Root
+## Merkle Patricia Trie
 
-[Merkle root]: glossary.md#merkle-roots
+[mpt]: glossary.md#merkle-patricia-trie
```
norswap, line 77: There are two links to this in this file that haven't been updated.

specs/glossary.md
```diff
@@ -72,12 +72,12 @@ block, and output block properties, which are derived after executing the block'
...
 
-The Merkle root is the root hash of a [Merkle Patricia tree] (MPT). A MPT is a sparse [trie], which is a tree-like
-structure that maps keys to values. The root hash of a MPT is a commitment to the contents of the tree, which allows a
+A [Merkle Patricia tree] (MPT) is a sparse trie, which is a tree-like structure that maps keys to values.
```
norswap, line 79: Since the heading is now "Merkle Patricia Trie", do you want to change this link say the same? (don't forget the anchor)

specs/rollup-node.md
```diff
@@ -251,6 +253,46 @@ However, they can still be challenged by a fault proof until the end of the faul
 
 > **TODO** the spec doesn't encode the notion of fault proof yet, revisit this (and include links) when it does
 
+## From L2 Block to L2 Output Root
+
+After processing a block the resulting outputs will need to be synchronized with L1 for trustless execution of
+L2-to-L1 messaging, such as withdrawals. Outputs are hashed in a tree structured form for minimal proof cost to any
+piece of data. This merkle-structure is defined with [SSZ], a type system for merkleization and serialization, used in
+L1 (beacon-chain). However, we replace `sha256` with `keccak256` to save gas costs in the EVM.
```
norswap, line 261: s/tree structured/tree-structured

norswap: s/for minimal proof cost to any piece of data/which minimizes the cost of proving any piece of data captured by the outputs/

We probably need a glossary entry for the concept of proof, I'll make an issue for it.

norswap: This is the first time the word "output" appears in this document, and so we need to define what it means.
Since the `L2Output` structure is defined right below, I think just linking to the `Block` entry of the glossary (which defines block outputs) should be enough.

specs/rollup-node.md
```diff
@@ -251,6 +253,46 @@ However, they can still be challenged by a fault proof until the end of the faul
...
+  state_root: Bytes32
+  withdrawal_storage_root: Bytes32
+  latest_block: ExecutionPayload  # includes block-hash
+  history_accumulator_root: Bytes32  # Not functional yet
+  extension: Bytes32
+```
+
+The `state_root` is the Merkle-Patricia-Trie ([MPT][g-mpt]) root of all execution-layer accounts,
+also found in `latest_block.state_root`: this field is frequently used and thus elevated closer to the L2 output root,
+reducing the merkle proof depth and thus the cost of usage.
```
norswap, line 276: Probably should explain that it is "elevated" because it could otherwise be proved against the block hash stored in the execution payload.

specs/rollup-node.md
norswap: add a TODO for adding links to "withdrawals" once we introduce them in the spec

specs/rollup-node.md
```diff
@@ -251,6 +253,46 @@ However, they can still be challenged by a fault proof until the end of the faul
...
+
+The `withdrawal_storage_root` elevates the Merkle-Patricia-Trie ([MPT][g-mpt]) root of L2 Withdrawal contract storage.
+Instead of a MPT proof to the Withdrawal contract account in the account trie,
+one can directly access the MPT storage trie root, thus reducing the verification cost of withdrawals on L1.
+
+The `latest_block` is an execution-layer block of L2, represented as the [`ExecutionPayload`][ExecutionPayload] SSZ type
+defined in L1. There may be multiple blocks per L2 output root, only the latest is presented.
```
norswap, line 283: This is a surprising comment at this point of the conversation. Under now, we're under the impression we need to derive outputs for every block, and this is done in the flow that goes from an L1 block to the L2 block.

If that's not the case, it should be precised towards the top of the section.

norswap: I'm wondering if this should be split up to a different document? The existing content would be the derivation spec, and this could be distinct. (In a subsequent PR though.)

specs/rollup-node.md
```diff
@@ -251,6 +253,46 @@ However, they can still be challenged by a fault proof until the end of the faul
...
+
+[ExecutionPayload]: https://github.com/ethereum/consensus-specs/blob/dev/specs/bellatrix/beacon-chain.md#executionpayload
+
+The `history_accumulator_root` is a reserved field, elevating a storage variable of the L2 chain that maintains
+the SSZ merkle root of an append-only `List[Bytes32, MAX_ITEM_COUNT]` (`keccak256` SSZ),
+where each item is defined as `keccak256(l2_block_hash ++ l2_state_root)`, one per block of the L2 chain.
+While reserved, a zeroed `Bytes32` is used instead.
+This is a work-in-progress, see [issue 181](https://github.com/ethereum-optimism/optimistic-specs/issues/181).
+
```
norswap, line 292: For now, this needs
- a link to the [SSZ spec](https://github.com/ethereum/consensus-specs/blob/dev/ssz/simple-serialize.md)
- a definition of `MAX_ITEM_COUNT`

norswap: In the future, we probably want to specify exactly where this storage slot comes from, and why it is a storage slot (i.e. we want to make the L2 history available on L2 too).

specs/rollup-node.md
```diff
@@ -251,6 +253,46 @@ However, they can still be challenged by a fault proof until the end of the faul
...
+The `extension` is a zeroed `Bytes32`, to be substituted with a SSZ container to extend merkleized information in future
+upgrades. This keeps the static merkle structure forwards-compatible.
+
 ## Whole L2 Chain Derivation
 
 The previous two sections presents an inductive process: given that we know the "current" L2 block, well as the next L1
```
norswap, line 298: No longer accurate :) Could temporarily change to the previous 3 section + see my comment about potentially splitting this spec.

specs/rollup-node.md
```diff
@@ -251,6 +253,46 @@ However, they can still be challenged by a fault proof until the end of the faul
...
+
+The `history_accumulator_root` is a reserved field, elevating a storage variable of the L2 chain that maintains
+the SSZ merkle root of an append-only `List[Bytes32, MAX_ITEM_COUNT]` (`keccak256` SSZ),
+where each item is defined as `keccak256(l2_block_hash ++ l2_state_root)`, one per block of the L2 chain.
+While reserved, a zeroed `Bytes32` is used instead.
+This is a work-in-progress, see [issue 181](https://github.com/ethereum-optimism/optimistic-specs/issues/181).
+
+The `extension` is a zeroed `Bytes32`, to be substituted with a SSZ container to extend merkleized information in future
+upgrades. This keeps the static merkle structure forwards-compatible.
+
```
norswap, line 295: Since each field is described in a single paragraph, would it be a good idea (i.e. not too ugly) to make them all list items?

norswap (changes requested):
Pretty good!

My major comment is to clarify when we need these outputs if not every block.

Also (cf. my comment), It hink we might want to move this to another file after it's merged!

specs/rollup-node.md
protolambda: The section title "From L2 Block to L2 Output Root" is definitely confusing here. I'll move it to a separate document

specs/rollup-node.md
protolambda: I think it's too much text for a list, and the whitespace between the items helps readability.

specs/rollup-node.md
protolambda: I am not sure about the exact output wording / glossary usage, feel free to change it

protolambda:
- Addressed review suggestions from @norswap 
- Separated into a proposals doc
- Rebased on main, integrate generated TOC etc.

norswap (approved):

PR #195, opened 2022-02-03
Package Restructure

### LINKED ISSUE #148: Re-organize rollup node packages
Currently the bulk of the rollup node is in the L2 package. Suggest at high level something like

```
L1
    - Interact with L1 client
L2
    - Interact with L2 Client
Rollup
    - Driver
    - Sync
    - Input Derivation (stateless pure function, keep separate for use in cannon)
    - Input Derivation in reverse (for L1 info transaction)
    - Output Derivation (to L2?)
```

I'll play with this to see how it feels, but I may not split up the rollup package too much until it grows larger.

trianglesphere:
**Description**
This PR organizes the former L2 package. Rollup specific components were subdivided into one of three sub packages in the rollup directory:
- derive: Input/Output/L2->L1 info steps. Now without a godocs
- driver: Runs the rollup
- sync: Determines the starting point for the sync.

Several functions in the `derive` package have been renamed and some comments have been updated. No functionality was changed.

**Metadata**
- Fixes #148


norswap:
While we're on refactoring, thoughts on eschewing the `input/output` terminology and simply use `payloadAttributes/executionPayload?. It feels redundant to have those synonyms and they're too generic to really help understanding.

opnode/l2/api.go
```diff
@@ -1,3 +1,4 @@
+// Package l2 connects to the L2 execution engine over the Catalyst API.
```
protolambda, line 1:
```suggestion
// Package l2 connects to the L2 execution engine over the Engine API.
```
"catalyst" is a legacy name used inside of geth for the Merge related parts (the first Geth merge prototype by Guillaume had this name), "Engine API" is better

protolambda (dismissed):
LGTM. One comment to fix, happy to merge after

Edit: renaming the input/output first would be good to avoid confusion with the output root, good suggestion from @norswap

maurelian:
> While we're on refactoring, thoughts on eschewing the `input/output` terminology and simply use `payloadAttributes/executionPayload?. It feels redundant to have those synonyms and they're too generic to really help understanding.

We concluded that 'Yes', we should do this. Esp. to avoid collision with the L2 Output Oracle/Submitter.

PR #196, opened 2022-02-04
Add sequenceNumber to L1Block.sol

### LINKED ISSUE #192: Update L1-info contract (the L2 predeploy) to store sequence number

maurelian:
**Description**

Adds `sequenceNumber` to data written in the L1 Attributed deposited tx

**Metadata**
- Fixes #192 
- Fixes ENG-1976

### DIFF
--- a/packages/contracts/contracts/L2/L1Block.sol
+++ b/packages/contracts/contracts/L2/L1Block.sol
@@ -15,12 +15,14 @@ contract L1Block {
     uint256 public number;
     uint256 public timestamp;
     uint256 public basefee;
+    uint256 public sequenceNumber;
     bytes32 public hash;
 
     function setL1BlockValues(
         uint256 _number,
         uint256 _timestamp,
         uint256 _basefee,
+        uint256 _sequenceNumber,
         bytes32 _hash
     ) external {
         if (msg.sender != DEPOSITOR_ACCOUNT) {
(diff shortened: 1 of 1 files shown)


packages/contracts/contracts/L2/L1Block.sol
```diff
@@ -30,6 +32,7 @@ contract L1Block {
         number = _number;
         timestamp = _timestamp;
         basefee = _basefee;
+        basefee = _sequenceNumber;
```
trianglesphere, line 35: ?, waiting on something else?

maurelian: typo, fixed.

maurelian:
Leaving in draft until the specs and ref-impl land. #182

norswap:
I'm sorry for the silly question, but what is a sequence number? Is it an epoch number? (And if so, is that what we should call it?)

Sequence number sounds like L2 block number to me, which is something that couldn't possibly be written here.

protolambda:
@norswap it's from an earlier idea to track how many L2 blocks have past since the last L1-info-deposit was included, so users/contracts can find it easily, and we might end up using it to try find the sync starting point faster. It's not really blocking anything, but was interesting enough to list in the epic. Focusing on the other sequencer work for now.

tynes:
Should this contract be behind a delegatecall proxy? In case it needs to be upgraded to include new properties like the value of `RANDOM`

norswap:
@tynes Just saw this. That's one option, with the other one being to make the change part of the hard fork, which would mean increasing the geth diff.

How would you go about the hard fork stuff seems like a bit of a headache however. Do we just stuff the contract in geth's memory? (That's the easiest way to do it with a hard fork.) Proxy definitely feels much easier.

PR #197, opened 2022-02-04
opnode: Add Dockerfile/Makefile

mslipper:
- Adds a Dockerfile so that we can build opnode images
- Adds a Makefile to make building the opnode easier

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,7 @@
+opnode:
+	go build -o ./bin/op ./opnode/cmd
+.PHONY: opnode
+
+clean:
+	rm -rf ./bin
+.PHONY: clean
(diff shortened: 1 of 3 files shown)


trianglesphere:
One longer term Q is that given Conner added a makefile in `bss` in #179 (and there's another one in `opnode/contracts`) do we want a makefile at the root of the repo or in the root of each project?

maurelian:
However we manage make files, a single command to build everything would be nice. That can be a separate PR tho.

Makefile
```diff
@@ -0,0 +1,7 @@
+opnode:
+	go build -o ./bin/op ./opnode/cmd
+.PHONY: opnode
+
+clean:
+	rm -rf ./bin
+.PHONY: clean
```
maurelian, line 7: should add the blank line, then LGTM.

maurelian (approved):

norswap:
Agree with @trianglesphere we should consolidate the Makefiles in an upcoming PR.

PR #200, opened 2022-02-08
opnode: Switch to urfave

trianglesphere:
**Description**
This removes OpNodeCmd in favor of a separate configuration and
node struct. Command line flag parsing is also done with urfave/cli.

I did this originally to more easily separate `OpNode` configuration and node management.

Some things that might need to be sorted out:
- Where to place `flags` (was a seperate package in the `l2outputsubmitter`)
- Geth's versioning note relies on upstream, not our versioning.
- Maybe some command line info cleanup? I just went with simple things
- [x] Update docs about running the node

**User Facing Changes**
- Removed the `run` sub-command. That is `opnode run --FLAGS` is now `opnode --FLAGS`
- No longer defaults L1 and L2 engine addresses (hard to do with slices and urfave/cli)


norswap:
Are there any changes in behaviour from a user perspective?

trianglesphere:
> Are there any changes in behaviour from a user perspective?

Yep. I initially put them in the section called `Misc. Changes`, but I've updated it to ``User Facing Changes`

opnode/cmd/main.go
```diff
@@ -1,31 +1,211 @@
...
+var (
+	Version   = ""
+	GitCommit = ""
+	GitDate   = ""
+)
+
+func main() {
+	// Set up logger with a default INFO level in case we fail to parse flags,
+	// otherwise the final critical log won't show what the parsing error was.
+	log.Root().SetHandler(
```
protolambda, line 29: Setting up the default global logger is ok here in the `main`, but please avoid using it as much as possible elsewhere in the implementation. Testing with multiple nodes that all use the same logger gets messy.

opnode/node/node.go
```diff
@@ -16,139 +13,126 @@ import (
...
 	}
+}
 
-	l1Sources := make([]eth.L1Source, 0, len(c.L1NodeAddrs))
-	for i, addr := range c.L1NodeAddrs {
+func New(ctx context.Context, cfg *Config) (*OpNode, error) {
+	if err := cfg.Check(); err != nil {
+		return nil, err
+	}
+	log := cfg.LogCfg.NewLogger()
```
protolambda, line 66: It would be neat if we can attach a test-logger instance to the node during testing, instead of creating a logger from the log-config.

opnode/cmd/main.go
```diff
@@ -1,31 +1,211 @@
...
+}
+
+func RollupNodeMain(ctx *cli.Context) error {
+	log.Info("Initializing Rollup Node")
+	cfg, err := NewConfig(ctx)
+	if err != nil {
+		log.Error("Unable to create the rollup node config", "error", err)
+		return err
+	}
+	n, err := node.New(context.Background(), cfg)
```
protolambda, line 57: I think we should create the logger from the log-config outside of the `node.New`, and insert it as argument. That way we can pass test-loggers, and tag node instances with different context params to differentiate between the logs in a multi-opnode test.

opnode/cmd/main.go
```diff
@@ -1,31 +1,211 @@
...
+	log.Root().SetHandler(
+		log.LvlFilterHandler(
+			log.LvlInfo,
+			log.StreamHandler(os.Stdout, log.TerminalFormat(true)),
+		),
+	)
+
+	app := cli.NewApp()
+	app.Flags = Flags
+	app.Version = fmt.Sprintf("%s-%s", Version, params.VersionWithCommit(GitCommit, GitDate)) // TODO: Don't use Geth's hardcoded versions
```
protolambda, line 38: Can we just re-implement `VersionWithCommit` (and maybe simplify it) to not default to the geth hardcoded version? I would be fine with a `dev` placeholder version string for now.

protolambda (changes requested):
Looks good, but let's separate the logger setup for testing, and I think we shouldn't default to the geth version.

opnode/cmd/main.go
trianglesphere: This is following from the bss (l2output submitter). I think this is fine as pretty much everywhere else it's using the created logger. The other option here is to use golang's log package just for the entrypoint (but that won't respect verbosity flags, but IDK if we respect verbosity flags here anyways).

opnode/cmd/main.go
trianglesphere: I went with a placeholder. This is another thing from the l2os (I left a comment there about it)

opnode/cmd/main.go
```diff
@@ -1,31 +1,223 @@
...
+// has0xPrefix validates str begins with '0x' or '0X'.
+// Copied from geth
+func has0xPrefix(str string) bool {
+	return len(str) >= 2 && str[0] == '0' && (str[1] == 'x' || str[1] == 'X')
 }
 
-func (c *MainCmd) Help() string {
-	return "Run Optimism rollup node."
+// HexToHash is copied from Geth, but does not supress the error
+func HexToHash(s string) (common.Hash, error) {
```
protolambda, line 107:
```
	var x common.Hash
	err := x.UnmarshalText([]byte(stringhere))
```
May be a shorter alternative

protolambda (approved):
LGTM, maybe change the hash parsing

opnode/cmd/main.go
trianglesphere: Yep, that's the what I want. The rest of the functions just seem to suppress errors.

PR #202, opened 2022-02-08
rollup: implement sequencing-window based derivation

### LINKED ISSUE #185: Update derivation with sequencer block construction from L1 batch data

protolambda:
This updates the block-derivation to:
- Take a sequencing window as input
- Check the size and consistency of the input
- Derive L2 payloads from possibly out-of-order batches (ignore past/future epochs with overlapping sequencing window)
- Build L2 blocks with fixed block time (fill gaps with empty blocks)
- Limit L2 sequencer-produced blocks to a time-window, do not fill ridiculous out-of-window gaps with empty L2 blocks (by accident or not)
- Force deposits of the epoch (first L1 block of window) into the batch of the first L2 time slot (created if none exists)

This is a *work in progress*. this is output from discussion with Karl/Norswap about the many options.

Fixes #185
Work in progress #183

TODOs (may split into more PRs):
- Spec for sequencing window, and outline why this seems to work best.
- Load new rollup configuration (block time and special addresses etc.)
- Update L1 downloader to fetch receipts lazily, and cache the block data (repeated usage due to overlapping sequencing windows).
- Implement batch encoding/parsing

### DIFF
--- a/opnode/rollup/derive/payload_attributes.go
+++ b/opnode/rollup/derive/payload_attributes.go
@@ -6,7 +6,6 @@ import (
 	"fmt"
 	"math/big"
 
-	"github.com/ethereum-optimism/optimistic-specs/opnode/eth"
 	"github.com/ethereum-optimism/optimistic-specs/opnode/rollup"
 
 	"github.com/ethereum-optimism/optimistic-specs/opnode/l2"
@@ -119,6 +118,8 @@ type L1Info interface {
 	Time() uint64
 	Hash() common.Hash
 	BaseFee() *big.Int
+	// MixDigest field, reused for randomness after The Merge (Bellatrix hardfork)
+	MixDigest() [32]byte
 }
 
 // L1InfoDeposit creats a L1 Info deposit transaction based on the L1 block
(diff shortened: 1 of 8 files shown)


trianglesphere:
For my own understanding, an overview or brief spec would be great.

If I'm reading it correctly, the batch submitter is separately submitting batches and this component is still only reading L1 data and then driving L2. I guess it's a little out of scope, but I'm super curious to see how sequenced (but not batched) transactions are handled.

norswap:
Summary of the derivation deisgn:

- no more separation between deposit blocks and sequencer blocks
    - but the first block of each epoch has the deposit transactions from the L1 block  associated with the epoch 
- epoch numbers are equal to the L1 block number that has the deposit
- the sequencer is free to pick the L2 block timestamp (expected to be the wallclock time when the block is "mined")
    - constraint: block timestamps are multiple of the block time (probably 2s)
    - constraint: successive blocks must have timestamps two second apart
- if no batches for the epoch associated with an L1 block are made within the sequencing window, then the rollup driver automatically derives blocks
    - it will give the first block of the epoch a timestamp that is `newL2Timestamp = max(blockTimeRounded(l1Timestamp), lastL2Timestamp + blockTime)`
    - it will fill in the missing time period (between `lastL2Timestamp` and `newL2Timestamp`) with empty blocks

opnode/rollup/derive/payload_attributes.go
```diff
@@ -178,43 +181,166 @@ func UserDeposits(height uint64, receipts []*types.Receipt) ([]*types.DepositTx,
...
+	// Retrieve the deposits of this epoch (all deposits from the first block)
+	deposits, err := DeriveDeposits(blocks[0])
+	if err != nil {
+		return nil, fmt.Errorf("failed to derive deposits: %v", err)
+	}
+
+	// All block times in L2 are config.BlockTime seconds apart. If L2 time moves faster than L1, use that.
+	// L1 is expected to have larger increments per block, L2 time will get back in sync eventually.
+	firstTimestamp := blocks[0].Time()
+	if firstTimestamp >= config.Genesis.L2Time { // if not, then lastL2Time will be larger
```
norswap, line 219: If `lastL2Time` will be larger, why do we even need this condition?

Why not:
```go
if lastL2Time+config.BlockTime > firstTimestamp {
	firstTimestamp = lastL2Time + config.BlockTime
} else {
	firstTimestamp = firstTimestamp - (firstTimestamp-config.Genesis.L2Time)%config.BlockTime
}
```

opnode/rollup/derive/payload_attributes.go
```diff
@@ -178,43 +181,166 @@ func UserDeposits(height uint64, receipts []*types.Receipt) ([]*types.DepositTx,
...
+		// round down, only create a trailing empty block if the L1 time is at least a full block span ahead of L2
+		firstTimestamp = firstTimestamp - (firstTimestamp-config.Genesis.L2Time)%config.BlockTime
 	}
+	if lastL2Time+config.BlockTime > firstTimestamp {
+		firstTimestamp = lastL2Time + config.BlockTime
+	}
+
+	// Sequencers may not use timestamps beyond the end of the sequencing window (with some configurable margin)
+	maxTimestamp := blocks[len(blocks)-1].Time() + config.MaxSequencerTimeDiff
```
norswap, line 228: While it does seem sane, is there a particular reason why do we have this constraint? i.e. what's the bad thing that can happen if this condition is violated?

opnode/rollup/derive/payload_attributes.go
```diff
@@ -176,45 +181,150 @@ func UserDeposits(height uint64, receipts []*types.Receipt) ([]*types.DepositTx,
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to derive deposits: %v", err)
+	}
+
+	// copy L1 randomness (mix-digest becomes randao field post-merge)
+	// TODO: we don't have a randomness oracle on L2, what should sequencing randomness look like.
+	// Repeating the latest randomness of L1 might not be ideal.
+	randomnessSeed := l2.Bytes32(l1Info.MixDigest())
+
+	// Collect all L2 batches, the bathes may be out-of-order, or possibly missing.
```
trianglesphere, line 259:
```suggestion
	// Collect all L2 batches, the batches may be out-of-order, or possibly missing.
```

PR #204, opened 2022-02-08
Doctoc: TOC automation and CI

protolambda:
This automates TOC building with `yarn lint:toc`, using [doctoc](https://github.com/thlorenz/doctoc). Enforced in CI by checking if the TOC update resulted in any git changes.

Note: I will mess up a TOC intentionally to check if CI works as intended, then roll back the change.

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,13 +6,15 @@
   "devDependencies": {
     "lint-staged": "11.0.0",
     "markdownlint": "^0.24.0",
-    "markdownlint-cli2": "^0.3.2"
+    "markdownlint-cli2": "^0.3.2",
+    "doctoc": "2.1.0"
   },
   "scripts": {
     "setup": "yarn install && cargo install lychee",
     "lint": "yarn lint:fix && yarn lint:check",
     "lint:fix": "markdownlint-cli2-fix \"./opnode/README.md\" \"./specs/**/*.md\" \"#node_modules\"",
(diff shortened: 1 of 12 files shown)


package.json
```diff
@@ -6,13 +6,15 @@
...
+    "doctoc": "2.1.0"
   },
   "scripts": {
     "setup": "yarn install && cargo install lychee",
     "lint": "yarn lint:fix && yarn lint:check",
     "lint:fix": "markdownlint-cli2-fix \"./opnode/README.md\" \"./specs/**/*.md\" \"#node_modules\"",
     "lint:check": "markdownlint-cli2  \"./opnode/README.md\" \"./specs/**/*.md\" \"#node_modules\"",
-    "lint:links": "lychee --exclude twitter.com --exclude-mail README.md \"./opnode/README.md\" \"./specs/**/*.md\" \"./meta/**/*.md\" \"./opnode/**/*.md\""
+    "lint:links": "lychee --exclude twitter.com --exclude-mail README.md \"./opnode/README.md\" \"./specs/**/*.md\" \"./meta/**/*.md\" \"./opnode/**/*.md\"",
+    "lint:toc": "yarn doctoc --title=\"**Table of Contents**\" ./specs ./meta"
```
maurelian, line 18: Nit: I always get confused about the semantics of prefixing these scripts with `yarn` or not. Does it matter or can we remove it for consistency?

maurelian (approved):
One nitty question, but LGTM, and will be happy to use this!

package.json
protolambda: I think we can remove it, good catch

PR #210, opened 2022-02-14
ref impl: Source and Driver Cleanup

trianglesphere:
**Description**
Hacking away code & restructuring it.

- Moves some L2 downloading logic from the rollup package to the L2 package
- Makes most of the driver API private
- Add a lifecycle to the driver API (`new`, `start`, `stop`, opaque struct object)


opnode/rollup/driver/driver.go
```diff
@@ -2,71 +2,136 @@ package driver
...
+	d.done <- ec
+	err := <-ec
+	return err
 }
 
-func (e *EngineDriver) requestEngineHead(ctx context.Context) (refL1 eth.BlockID, refL2 eth.BlockID, err error) {
-	refL1, refL2, _, err = e.SyncRef.RefByL2Num(ctx, nil, &e.Genesis)
-	return
+func (d *Driver) loop() {
+	ctx := context.Background()
```
protolambda, line 69: We should cancel this ctx when the driver close signal is received.

opnode/rollup/driver/driver.go
```diff
@@ -2,71 +2,136 @@ package driver
...
+		// And we want to slow down requesting the L2 engine for its head (we just changed it ourselves)
+		// Request head if we don't successfully change it in the next 14 seconds.
+		l2HeadPoll.Reset(time.Second * 14)
+	}
+	defer syncTicker.Stop()
+	defer l2HeadPoll.Stop()
+
+	for {
+		select {
+
```
protolambda, line 96: See other comment. Now that we're not passing the ctx as argument into the loop, and waiting solely for the driver close signal, we should cancel the ctx ourselves (and remove the empty line where it used to be)

opnode/rollup/driver/driver.go
```diff
@@ -2,71 +2,136 @@ package driver
...
+	ec := make(chan error)
+	d.done <- ec
+	err := <-ec
+	return err
 }
 
-func (e *EngineDriver) requestEngineHead(ctx context.Context) (refL1 eth.BlockID, refL2 eth.BlockID, err error) {
-	refL1, refL2, _, err = e.SyncRef.RefByL2Num(ctx, nil, &e.Genesis)
-	return
+func (d *Driver) loop() {
```
protolambda, line 68: Tbh I liked how the loop was separated because it makes encapsulated testing a bit easier. It had a lot of similar function arguments though, so I'm ok with attaching it to the Driver.

protolambda (changes requested):
need to cancel the ctx on close, restructure looks ok otherwise

opnode/rollup/driver/driver.go
trianglesphere: Background contexts aren't cancelable. We cancel the timeout context after the result of each operation.

opnode/rollup/driver/driver.go
trianglesphere: See below that each cancelable context is cancelled right after being created. Also FYI I added more timeouts.

opnode/rollup/driver/driver.go
trianglesphere: It was enough overlap to where the driver doesn't have much that the loop doesn't depend on, but if the driver keeps growing, we can split the loop out later.

opnode/rollup/driver/driver.go
protolambda: We can wrap thix `ctx` with `context.WithCancel`, and call `cancel` when we receive the close signal. There's no point in passing in the background context if we don't cancel it on close.

protolambda: Hmm, timeouts and direct `cancel` is actually fine, my bad

opnode/node/node.go
```diff
@@ -138,27 +126,28 @@ func (c *OpNode) Start() error {
...
 
 		// driver subscribes to L1 head changes
 		l1SubCh := make(chan eth.HeadSignal, 10)
 		l1HeadsFeed.Subscribe(l1SubCh)
 		// start driving engine: sync blocks by deriving them from L1 and driving them into the engine
-		engDriveSub := eng.Drive(c.ctx, l1SubCh)
-		handleUnsubscribe(engDriveSub, "engine driver unexpectedly failed")
+		err := eng.Start(reqCtx, l1SubCh)
+		// engDriveSub := eng.Drive(c.ctx, l1SubCh)
+		// handleUnsubscribe(engDriveSub, "engine driver unexpectedly failed")
```
protolambda, line 137: can we remove these?

protolambda (dismissed):
LGTM, just two old commented code lines that we should remove I think

opnode/node/node.go
```diff
@@ -138,27 +126,28 @@ func (c *OpNode) Start() error {
...
+		if err != nil {
+			c.log.Error("Could not start a rollup node", "err", err)
+			return err
+		}
 	}
 
 	// Keep subscribed to the L1 heads, which keeps the L1 maintainer pointing to the best headers to sync
 	l1HeadsSub := event.ResubscribeErr(time.Second*10, func(ctx context.Context, err error) (event.Subscription, error) {
 		if err != nil {
 			c.log.Warn("resubscribing after failed L1 subscription", "err", err)
```
protolambda, line 148: on node close this starts a resubscription loop

protolambda, line 148:
```suggestion
		if err != nil {
			if err == ctx.Err() {
				return nil, err
			}
			c.log.Warn("resubscribing after failed L1 subscription", "err", err)
```
If the outer ctx fails, we shouldn't create the new subscription

opnode/rollup/driver/driver.go
```diff
@@ -2,71 +2,139 @@ package driver
...
+		// And we want to slow down requesting the L2 engine for its head (we just changed it ourselves)
+		// Request head if we don't successfully change it in the next 14 seconds.
+		l2HeadPoll.Reset(time.Second * 14)
+	}
+	defer syncTicker.Stop()
+	defer l2HeadPoll.Stop()
+
+	for {
+		select {
+		case <-d.done:
```
protolambda, line 96: are we not sending anything on the error channel that is passed through the done channel?

trianglesphere: Nope. I've switched some of the done channel to not return an error and I'll probably do that here. It doesn't feel like there is a well defined way to return an error only at close time.

opnode/node/node.go
trianglesphere: tests still broken, will fix

protolambda: Looks like it is still resubscribing. It doesn't retry creating the inner RPC watcher anymore, but the `ResubscribeErr` still takes the error and does not stop because of it. Maybe we unsub from the `ResubscribeErr`? Does the `unsub` list not already run on close?

protolambda (approved):
LGTM

PR #213, opened 2022-02-16
ref impl: Remove code

trianglesphere:
**Description**
Remove function adapter for CanonicalChain and other methods.
This places the L1 Source Interface in the L1 package.


opnode/node/node.go
```diff
@@ -37,12 +37,11 @@ func (cfg *Config) Check() error {
...
-	log          log.Logger
-	l1Source     eth.L1Source           // (combined) source to fetch data from
-	l1Downloader l1.Downloader          // actual downloader
-	l2Engines    []*driver.EngineDriver // engines to keep synced
-	ctx          context.Context        // Embeded CTX to be removed
-	done         chan struct{}
+	log       log.Logger
+	l1Source  l1.Source              // Source to fetch data from (also implements the Downloader interface)
+	l2Engines []*driver.EngineDriver // engines to keep synced
+	ctx       context.Context        // Embeded CTX to be removed
```
protolambda, line 43: typo in comment, and maybe we should just remove the ctx at this point.

protolambda (commented):
LGTM, typo though, and maybe worth to remove the `ctx` from opnode while we are changing things here.

opnode/node/node.go
trianglesphere: Fixed in next PR

protolambda (approved):

PR #135, opened 2022-01-14
Add code coverage reporting for go in CI

maurelian:
Adds test coverage reports using CodeCov. 

They will appear here: https://app.codecov.io/gh/ethereum-optimism/optimistic-specs

PR #214, opened 2022-02-22
ref impl: Track code coverage across packages

trianglesphere:
**Description**

This enables tracking code coverage across packages to see an increase
in code coverage due to the e2e test.

**Metadata**
- Fixes #135

### DIFF
--- a/.github/workflows/go-ci.yml
+++ b/.github/workflows/go-ci.yml
@@ -26,6 +26,6 @@ jobs:
           # rules: https://golangci-lint.run/usage/quick-start/
           args: -E asciicheck,goimports
       - name: Test and generate coverage
-        run: go test -coverprofile=coverage.out -covermode=atomic ./...
+        run: go test -coverpkg=all -coverprofile=coverage.out -covermode=atomic ./...
       - name: Upload coverage output
         uses: codecov/codecov-action@v2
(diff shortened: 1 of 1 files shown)


.github/workflows/go-ci.yml
```diff
@@ -26,6 +26,6 @@ jobs:
           # rules: https://golangci-lint.run/usage/quick-start/
           args: -E asciicheck,goimports
       - name: Test and generate coverage
-        run: go test -coverprofile=coverage.out -covermode=atomic ./...
+        run: go test -coverpkg=all -coverprofile=coverage.out -covermode=atomic ./...
```
protolambda, line 29: I see all examples/docs use `coverpkg=./...`, but `all` was used in the original go changelog of `coverpkg`. And `go help packages` describes them somewhat differently. Shall we change it to `./...`? Kind of curious if we get different results, and it matches the package selector nicely.

protolambda (commented):
Looks good, but maybe use `./...` instead

.github/workflows/go-ci.yml
trianglesphere: It seems like using `./...` covers a smaller subset, but also seems like what we want - apparently `all` includes stuff like even the standard library.

PR #219, opened 2022-03-03
ref impl: Verify L2 from batches

### LINKED ISSUE #149: Look into syncing more blocks at once in the sync algorithm
As it stands the main loop finds the sync starting point and then syncs forwards one block, and then repeats that loop again.

The goal is to reduce the number of calls to sync_start that have to be performed to the L1 execution client.
1. Cache the result. This has some issues with stale results.
2. When you sync, sync until head. This also has cancellation issues.
3. When you sync, sync a batch of blocks (say up to 100). This has smaller cancellation issues, but can help improve the speed of a deep sync.

### LINKED ISSUE #150: Tighten up the driver and drive_step code
This was code that proto refactored just before creating the initial rollup node.
That refactor helped a bunch, but the code could probably be cleaned up just a bit more.
One thing to think about is cleaning up names (if there are better ones).

### LINKED ISSUE #185: Update derivation with sequencer block construction from L1 batch data

### LINKED ISSUE #189: Update derivation to build sequencer blocks from L1 batch data

trianglesphere:
**Description**

This PR enables verifiers to use submitted batches to construct the L2 chain. It updates two areas: block derivation and the rollup driver.

This updates the block-derivation to:
* Take a sequencing window as input
* Check the size and consistency of the input
* Derive L2 payloads from possibly out-of-order batches (ignore past/future epochs with overlapping sequencing window)
* Build L2 blocks with fixed block time (fill gaps with empty blocks)
* Limit L2 sequencer-produced blocks to a time-window, do not fill ridiculous out-of-window gaps with empty L2 blocks (by accident or not)
* Force deposits of the epoch (first L1 block of window) into the batch of the first L2 time slot (created if none exists)

This updates the rollup driver:
* Split the rollup driver into `state` and `step`.
* The `state` object maintains a view of the L1 and L2 chains and is responsible for producing sequencing windows based on new L1 blocks or L1/L2 re-orgs.
* The `step` function takes the sequence window (list of block number / hashes) then fetches data from L1, turns that into L2 blocks, and then executes L2 blocks.
* Adds a new `fakeChainSource` to test the rollup driver
* Refactors the rollup driver loop and handling code to look at sequencing windows instead of a single block at a time.

**Metadata**
- Fixes #150 
- Fixes #149 
- Fixes #189
- Fixes #185
- Fixes ENG-2013
- Fixes ENG-2014

### DIFF
--- a/opnode/rollup/derive/payload_attributes.go
+++ b/opnode/rollup/derive/payload_attributes.go
@@ -239,7 +239,7 @@ type L2Info interface {
 //  - The L2 information of the block the new derived blocks build on
 //
 // This is a pure function.
-func PayloadAttributesV2(config *rollup.Config, l1Info L1Info, receipts []*types.Receipt, seqWindow []BatchData, l2Info L2Info) ([]*l2.PayloadAttributes, error) {
+func PayloadAttributes(config *rollup.Config, l1Info L1Info, receipts []*types.Receipt, seqWindow []BatchData, l2Info L2Info) ([]*l2.PayloadAttributes, error) {
 	// check if we have the full sequencing window
 	if len(seqWindow) == 0 {
 		return nil, errors.New("cannot derive payload attributes from empty sequencing window")
@@ -294,43 +294,6 @@ func PayloadAttributesV2(config *rollup.Config, l1Info L1Info, receipts []*types
 	out[0].Transactions = append(append(make([]l2.Data, 0), deposits...), out[0].Transactions...)
 
 	return out, nil
-}
-
-// PayloadAttributes derives the pre-execution payload from the L1 block info and deposit receipts.
-// This is a pure function.
-func PayloadAttributes(block L1Info, receipts []*types.Receipt) (*l2.PayloadAttributes, error) {
-	// if !CheckReceipts(block, receipts) {
-	// 	return nil, fmt.Errorf("receipts are not consistent with the block's receipts root: %s", block.ReceiptHash())
-	// }
-
-	l1Tx := types.NewTx(L1InfoDeposit(block))
-	opaqueL1Tx, err := l1Tx.MarshalBinary()
-	if err != nil {
-		return nil, fmt.Errorf("failed to encode L1 info tx")
-	}
-
-	userDeposits, err := UserDeposits(block.NumberU64(), receipts)
-	if err != nil {
-		return nil, fmt.Errorf("failed to derive user deposits: %v", err)
-	}
-
-	encodedTxs := make([]l2.Data, 0, len(userDeposits)+1)
-	encodedTxs = append(encodedTxs, opaqueL1Tx)
-
-	for i, tx := range userDeposits {
-		opaqueTx, err := types.NewTx(tx).MarshalBinary()
-		if err != nil {
-			return nil, fmt.Errorf("failed to encode user tx %d", i)
-		}
-		encodedTxs = append(encodedTxs, opaqueTx)
-	}
-
-	return &l2.PayloadAttributes{
-		Timestamp:             l2.Uint64Quantity(block.Time()),
-		Random:                l2.Bytes32(block.MixDigest()),
-		SuggestedFeeRecipient: common.Address{}, // nobody gets tx fees for deposits
-		Transactions:          encodedTxs,
-	}, nil
 }
 
 func DeriveDeposits(l1Info L1Info, receipts []*types.Receipt) ([]l2.Data, error) {
--- a/opnode/rollup/derive/payload_attributes.go
+++ b/opnode/rollup/derive/payload_attributes.go
@@ -2,8 +2,11 @@ package derive
 
 import (
 	"encoding/binary"
+	"errors"
 	"fmt"
 	"math/big"
+
+	"github.com/ethereum-optimism/optimistic-specs/opnode/rollup"
 
 	"github.com/ethereum-optimism/optimistic-specs/opnode/l2"
 	"github.com/ethereum/go-ethereum/common"
@@ -115,6 +118,8 @@ type L1Info interface {
 	Time() uint64
 	Hash() common.Hash
 	BaseFee() *big.Int
+	// MixDigest field, reused for randomness after The Merge (Bellatrix hardfork)
+	MixDigest() common.Hash
 }
 
 // L1InfoDeposit creats a L1 Info deposit transaction based on the L1 block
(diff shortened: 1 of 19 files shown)


opnode/rollup/driver/state_test.go
```diff
@@ -37,76 +37,183 @@ func (id testID) ID() eth.BlockID {
...
+	args := <-outputIn
+	assert.Equal(t, int(s.Config.SeqWindowSize), len(args.l1Window), "Invalid L1 window size")
+	assert.Equal(t, len(expectedWindow), len(args.l1Window), "L1 Window size does not match expectedWindow")
+	for i := range expectedWindow {
+		assert.Equal(t, expectedWindow[i].ID(), args.l1Window[i], "Window elements must match")
+	}
+	outputReturn <- outputReturnArgs{l2Head: src.setL2Head(int(args.l2Head.Number) + 1).Self, err: nil}
+}
+
+func reorg__L2(t *testing.T, expectedWindow []testID, s *state, src *fakeChainSource, outputIn chan outputArgs, outputReturn chan outputReturnArgs) {
```
protolambda, line 103: maybe remove the underscores?

protolambda (approved):
LGTM. I'll open up an issue to implement configuration loading & more polished defaults, not a priority right now. Edit: opened #222

opnode/rollup/driver/state_test.go
trianglesphere: this was to make the test cases below line up.

opnode/rollup/driver/fake_chain.go
```diff
@@ -0,0 +1,187 @@
...
+	}
+	return m.l1s[m.l1reorg][m.l1head]
+}
+
+func (m *fakeChainSource) l1Head() eth.L1BlockRef {
+	m.log.Trace("L1 Head", "head", m.l1head)
+	return m.l1s[m.l1reorg][m.l1head]
+}
+
+// Unused functions
```
tynes, line 175: Delete?

opnode/rollup/driver/step.go
```diff
@@ -7,53 +7,119 @@ import (
...
+	logger.Trace("Got receipts")
+	// TODO: with sharding the blobs may be identified in more detail than L1 block hashes
+	batches, err := d.dl.FetchBatches(fetchCtx, l1Input)
+	if err != nil {
+		return l2Head, fmt.Errorf("failed to fetch batches from %s: %v", l1Input, err)
+	}
+	logger.Trace("Got batches")
 
-	attrs, err := derive.PayloadAttributes(bl, receipts)
+	attrsList, err := derive.PayloadAttributes(&d.Config, l1Info, receipts, batches, l2Info)
```
protolambda, line 81: Batch parsing and filtering needs to happen somewhere between the downloader and the continued derivation of L2 blocks. We should probably fetch the list of transactions per L1 block instead, and then call the parsing/filtering here, to not leak that into the downloader logic.

When we introduce blob-transactions we'll end up needing an additional `FetchBlobs(ctx, versionedHashesList)` method. We still need to parse/filter transactions in that case, but instead of immediately parsing calldata we'll need to fetch the blobs referenced by the transaction, and then parse those into batches.

So now we change:
- `FetchBatches` -> `FetchTransactions`
- Call `BatchesFromEVMTransactions` after fetching
- Call `FilterBatches` after parsing

And later with blobs that will be:
- `FetchTransactions`
- `BatchBlobsFromEVMTransactions`
- `BatchesFromBlobs`
- `FilterBatches`

And depending on cost/performance trade-offs we might still encode e.g. the batch epoch and timestamp in calldata, but put all the other data in the blobs. That way we can skip out-of-range data easily

opnode/rollup/derive/payload_attributes.go
```diff
@@ -176,45 +180,151 @@ func UserDeposits(height uint64, receipts []*types.Receipt) ([]*types.DepositTx,
...
+//  - The data batches the L2 sequencer work is derived from
+//  - The L2 information of the block the new derived blocks build on
+//
 // This is a pure function.
-func PayloadAttributes(block BlockInput, receipts []*types.Receipt) (*l2.PayloadAttributes, error) {
-	if !CheckReceipts(block, receipts) {
-		return nil, fmt.Errorf("receipts are not consistent with the block's receipts root: %s", block.ReceiptHash())
+func PayloadAttributes(config *rollup.Config, l1Info L1Info, receipts []*types.Receipt, seqWindow []BatchData, l2Info L2Info) ([]*l2.PayloadAttributes, error) {
+	// Retrieve the deposits of this epoch (all deposits from the first block)
+	deposits, err := DeriveDeposits(l1Info, receipts)
```
protolambda, line 243: TODO: separate system and user deposits

opnode/rollup/derive/payload_attributes.go
```diff
@@ -176,45 +180,151 @@ func UserDeposits(height uint64, receipts []*types.Receipt) ([]*types.DepositTx,
...
+				Timestamp:             l2.Uint64Quantity(t),
+				Random:                randomnessSeed,
+				SuggestedFeeRecipient: config.FeeRecipientAddress,
+				Transactions:          nil,
+			})
+		}
+	}
+
+	// Force deposits into the first block
+	out[0].Transactions = append(append(make([]l2.Data, 0), deposits...), out[0].Transactions...)
```
protolambda, line 294: TODO: when inserting the user deposited transactions, they need to be placed between the system deposit tx and the batch transactions

opnode/rollup/driver/step.go
trianglesphere: Updated to follow specs. please double check.

opnode/rollup/derive/payload_attributes.go
```diff
@@ -176,45 +180,151 @@ func UserDeposits(height uint64, receipts []*types.Receipt) ([]*types.DepositTx,
...
+	// If there are no submitted batches, at least derive the deposit block.
+	if len(seqWindow) == 0 {
+		highestSeenTimestamp += config.BlockTime
+	}
+
+	// fill the gaps and always ensure at least one L2 block
+	// TODO: What happens if `highestSeenTimestamp` start rolling into the next epoch (higher than the next L1 block timestamp)
+	var out []*l2.PayloadAttributes
+	for t := l1Info.Time() + config.BlockTime; t <= highestSeenTimestamp; t += config.BlockTime {
+		if bl, ok := l2Blocks[t]; ok {
```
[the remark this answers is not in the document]
trianglesphere: Done

opnode/rollup/derive/payload_attributes.go
trianglesphere: Do you mean pulling out the l1InfoTx deposit?

opnode/rollup/derive/payload_attributes.go
trianglesphere: Done, but a little hacky.

opnode/rollup/derive/payload_attributes.go
trianglesphere: Without adding unique info the the L1 info tx, this working fine but hacky. I think it should probably be cleaned up, but idc about now or in a follow on.

opnode/rollup/driver/state.go
```diff
@@ -2,118 +2,200 @@ package driver
...
-	log.Debug("finding next sync step, engine syncing", "l2", e.l2Head, "l1", e.l1Head)
-	nextRefL1, refL2, err := driver.findSyncStart(ctx)
+	return s.l1Window[len(s.l1Window)-1]
+}
+
+// extendL1Window extends the cached L1 window by pulling blocks from L1.
+// It starts just after `s.l1WindowEnd()`.
+func (s *state) extendL1Window(ctx context.Context) error {
+	s.log.Trace("Extending the cached window from L1", "cached_size", len(s.l1Window), "window_end", s.l1WindowEnd())
+	nexts, err := s.input.L1ChainWindow(ctx, s.l1WindowEnd())
```
norswap, line 91: It probably doesn't matter *too* much if there's download caching, but shouldn't this be bounded to only download at most the number of blocks we're actually going to need to fill in the remainder of the window?

norswap (commented):
note: didn't review tests

opnode/rollup/driver/step.go
```diff
@@ -7,53 +7,116 @@ import (
...
+// An incomplete sequencing window will result in an incomplete L2 chain if so.
+//
+// After the step completes it returns the block ID of the last processed L2 block, even if an error occurs.
+func (d *outputImpl) step(ctx context.Context, l2Head eth.BlockID, l2Finalized eth.BlockID, l1Input []eth.BlockID) (out eth.BlockID, err error) {
+	if len(l1Input) == 0 {
+		return l2Head, fmt.Errorf("empty L1 sequencing window on L2 %s", l2Head)
+	}
+
+	logger := d.log.New("input_l1_first", l1Input[0], "input_l1_last", l1Input[len(l1Input)-1],
+		"input_l2_parent", l2Head, "finalized_l2", l2Finalized)
```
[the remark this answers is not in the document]
trianglesphere: Yes. when logged these are treated as key-value items. When directly using the logging methods, the first argument is the log string and then the rest are the same key/value info bits.

opnode/rollup/driver/step.go
```diff
@@ -7,53 +7,116 @@ import (
...
-	logger.Debug("fetched L1 data for driver")
+	receipts, err := d.dl.FetchReceipts(fetchCtx, l1Input[0])
+	if err != nil {
+		return l2Head, fmt.Errorf("failed to fetch receipts of %s: %v", l1Input[0], err)
+	}
+	// TODO: with sharding the blobs may be identified in more detail than L1 block hashes
+	transactions, err := d.dl.FetchTransactions(fetchCtx, l1Input)
+	if err != nil {
+		return l2Head, fmt.Errorf("failed to fetch transactions from %s: %v", l1Input, err)
+	}
```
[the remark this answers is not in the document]
trianglesphere: We could. I'm not sure if it's worth it yet.

opnode/rollup/types.go
```diff
@@ -1,8 +1,48 @@
...
+	// Genesis anchor point of the rollup
+	Genesis Genesis
+	// Seconds per L2 block
+	BlockTime uint64
+	// Sequencer batches may not be more than MaxSequencerTimeDiff seconds after
+	// the L1 timestamp of the sequencing window end.
+	//
+	// Note: When L1 has many 1 second consecutive blocks, and L2 grows at fixed 2 seconds,
+	// the L2 time may still grow beyond this difference.
+	MaxSequencerTimeDiff uint64
```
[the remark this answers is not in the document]
trianglesphere: I'm not to sure here, this was originally @protolambda and I didn't really touch this

opnode/rollup/driver/state.go
trianglesphere: Perf wise we may want to only download the minimal amount, but I do like amortizing the cost and having a local cache that is larger than then window for the syncing case. In the case where it is caught up to head, it shouldn't hit this codepath.

opnode/rollup/driver/state.go
```diff
@@ -2,118 +2,200 @@ package driver
...
+type outputInterface interface {
+	step(ctx context.Context, l2Head eth.BlockID, l2Finalized eth.BlockID, l1Window []eth.BlockID) (eth.BlockID, error)
+}
 
-	// l2Head tracks the head-block of the engine
-	l2Head eth.BlockID
+type state struct {
+	// Chain State
+	l1Head      eth.BlockID   // Latest recorded head of the L1 Chain
+	l1Base      eth.BlockID   // L1 Parent of L2 Head block
```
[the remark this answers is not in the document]
trianglesphere: So I renamed `L1Parent` -> `L1origin`. I don't love `L1Origin` for this because it is not any generic `L1Origin`, it is the `L1Origin` of the `L2SafeHead`. Maybe origin is a little better, but I would not name this `L1Parent`

opnode/rollup/driver/driver.go
```diff
@@ -2,142 +2,59 @@ package driver
...
-	return
+func (i *inputImpl) L1ChainWindow(ctx context.Context, base eth.BlockID) ([]eth.BlockID, error) {
+	return sync.FindL1Range(ctx, i.chainSource, base)
 }
 
-func (e *Driver) driverStep(ctx context.Context, nextRefL1 eth.BlockID, refL2 eth.BlockID, finalized eth.BlockID) (l2ID eth.BlockID, err error) {
-	return e.step(ctx, nextRefL1, refL2, finalized.Hash)
+func (i *inputImpl) SafeL2Head(ctx context.Context) (eth.L2BlockRef, error) {
+	return sync.FindSafeL2Head(ctx, i.chainSource, i.genesis)
 }
```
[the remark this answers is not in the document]
trianglesphere: Initially the state was more isolated from the driver. I do want to take a look at cleaning this up later, but it requires cleaning up the source interfaces (`SafeL2Head`, etc).

opnode/rollup/sync/reference.go
```diff
@@ -26,10 +26,10 @@ type L2Client interface {
...
 // ChainSource provides access to the L1 and L2 block graph
 type ChainSource interface {
-	L1NodeByNumber(ctx context.Context, l1Num uint64) (eth.L1Node, error)
-	L1HeadNode(ctx context.Context) (eth.L1Node, error)
-	L2NodeByNumber(ctx context.Context, l2Num *big.Int) (eth.L2Node, error)
-	L2NodeByHash(ctx context.Context, l2Hash common.Hash) (eth.L2Node, error)
+	L1NodeByNumber(ctx context.Context, l1Num uint64) (eth.L1BlockRef, error)
+	L1HeadNode(ctx context.Context) (eth.L1BlockRef, error)
+	L2NodeByNumber(ctx context.Context, l2Num *big.Int) (eth.L2BlockRef, error)
+	L2NodeByHash(ctx context.Context, l2Hash common.Hash) (eth.L2BlockRef, error)
```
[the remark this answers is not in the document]
trianglesphere: The single PR would be nice. It would have been a pain to do the rename before this PR though, so doing it after requires people to review with the slightly confusing names.

opnode/rollup/derive/payload_attributes.go
```diff
@@ -176,45 +180,163 @@ func UserDeposits(height uint64, receipts []*types.Receipt) ([]*types.DepositTx,
...
+		// Check if we have already seen a batch for this L2 block
+		if _, ok := uniqueTime[batch.Timestamp]; ok {
+			// block already exists, batch is duplicate (first batch persists, others are ignored)
+			continue
+		}
+		uniqueTime[batch.Timestamp] = struct{}{}
+		out = append(out, batch)
+	}
+	return
+}
```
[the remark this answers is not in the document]
trianglesphere: They could be, or maybe switch `FilterBatch` to `ValidBatch` and then keep the check outside `BatchesFromEVMTransactions`. It doesn't really bother me to keep it as is though.

opnode/rollup/sync/reference.go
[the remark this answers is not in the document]
trianglesphere: Also, I updated the names

opnode/rollup/derive/payload_attributes.go
```diff
@@ -176,45 +180,163 @@ func UserDeposits(height uint64, receipts []*types.Receipt) ([]*types.DepositTx,
...
+		}
+	}
+
+	// If there are no submitted batches, at least derive the deposit block.
+	if len(seqWindow) == 0 {
+		highestSeenTimestamp += config.BlockTime
+	}
+
+	// fill the gaps and always ensure at least one L2 block
+	// TODO: What happens if `highestSeenTimestamp` start rolling into the next epoch (higher than the next L1 block timestamp)
```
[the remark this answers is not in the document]
trianglesphere: hmm, I think that B would be disallowed via minL2Time.

I wrote this comment prior to actually using the `FilterBatches` function which enforces the min and max l2 times (from one block after the last l2 head to the `l1 info block + block time`). With this check I think there might be some funky things with where the blocks line up relative to L1, but I think it sufficiently constrains the timestamps such that this is no longer a problem.

PR #220, opened 2022-03-03
Fix deposit specs to account for contract creation

norswap:
The spec didn't reflect the `isCreation` field of the `TransactionDeposited` event.

Similarly, we failed to mention that in deposited transactions, the `to` field must be the null address in case of a contract creation.


protolambda:
Spec question: should we allow deposits to deploy contracts at specific addresses, instead of deriving it from combining `from`/`nonce` etc.?

norswap:
@protolambda can you speak to the use cases? (and maybe perceived risks)

The big I see this allowing which is not possible on mainnet, is deploying a contract to a EOA. I guess that's alright, but I'm uneasy about the implications of diverging from mainnet (also potentially far down the line e.g. w/ address extension which has some really non-trivial implications).

norswap:
Merging this tomorrow morning if there's no further comments today.

Issue #187, opened 2022-02-01
Harden Downloader w/ batching, concurrency limits and memoization


protolambda:
PR #221 simplified the downloader to continue with sequencing work, but removed some features also. For production we should improve the downloader:
- Sanity limit on concurrency
- Improved retries and timeouts
- Implement memoization (that other code already assumes)
- Provide more optionality in what to download (we don't need receipts of all blocks right away, only for the start of a window)

mslipper:
I think we can should this to use batched RPCs rather than multiple concurrent requests.

PR #221, opened 2022-03-04
ref impl: Simpler downloader

trianglesphere:
**Description**
This was done because when the sequencer was using the old downloader, it was routinely taking 20-30 seconds (or more) to fetch receipts from L1. I'm not sure what the root cause was, but I was not able to find an easy fix.

This downloader is radically simpler, but does not provide as much as the previous downloader. It does not cache the result of the fetches (that should be a layer on top), nor does it limit total concurrency (but could be easily extended to do so).

It concurrently fetches receipts and performs retries (with constant rather than exponential backoff). It will panic in the case that it is unable to complete a fetch even with the retires.


protolambda:
> It will panic in the case that it is unable to complete a fetch even with the retires.

Panics should never be part of control-flow. If it fails with retries then it might be because the L1 source is offline, that shouldn't make the rollup-node also go offline.

Also, I'm hesitant to not limit concurrency and unsure about the best backoff strategy. Simplification is ok for now, but we may want to add some of that back, to make sure the rollup-node is not vulnerable to any kind of DoS with that resource consumption.

trianglesphere:
> Panics should never be part of control-flow. If it fails with retries then it might be because the L1 source is offline, that shouldn't make the rollup-node also go offline.

Panic removed, now it returns an error.

opnode/l1/downloader.go
```diff
@@ -2,291 +2,69 @@ package l1
...
-	//
-	// First field, aligned atomic changes.
-	finished uint32
-
-	// Count already downloaded receipts to track status
-	//
-	// Aligned after above field, atomic changes.
-	downloadedReceipts uint32
+var MaxConcurrentFetchesPerCall = 10
+var MaxReceiptRetry = 3
```
mslipper, line 15: Could be `const`:

```suggestion
const MaxConcurrentFetchesPerCall = 10
const MaxReceiptRetry = 3
```

mslipper: We should parameterize these, but for now this is fine IMO.

opnode/l1/downloader.go
```diff
@@ -2,291 +2,69 @@ package l1
...
-	for _, w := range dl.receiptWorkers {
-		w.Unsubscribe()
-	}
-	dl.receiptWorkers = dl.receiptWorkers[:0]
-	close(dl.receiptTasks)
-	for task := range dl.receiptTasks {
-		task.dest.Finish(wrappedErr{DownloadClosedErr})
-	}
+	wg.Wait()
+	return block, receipts, retErr
```
mslipper, line 69: To make reasoning about this function easier, I think we should return `nil, nil, retErr` if `retErr != nil`. This way either all of the receipt fetches succeed, or the function returns an error. I doubt callers will use the value of `receipts` when `retErr != nil` anyway since they could contain unexpected nil values.

opnode/l1/downloader.go
```diff
@@ -2,291 +2,69 @@ package l1
...
-
-			// no receipts to fetch? Then we are done!
-			if len(txs) == 0 {
-				dlTask.Finish(wrappedErr{nil})
+			semaphoreChan <- struct{}{}
+			for j := 0; j < MaxReceiptRetry; j++ {
+				receipt, err := dl.client.TransactionReceipt(ctx, hash)
+				if err != nil && j == MaxReceiptRetry-1 {
+					// dl.log.Error("Got error in final retry of fetch", "err", err)
+					errMu.Lock()
```
mslipper, line 54: Nit: Could use `atomic.Value` rather than a mutex, but it doesn't really make that much of a differentce.

mslipper (approved):

mslipper:
Made a couple of comments, none of which are blocking.

protolambda (approved):
Nice, this should be good enough and works around the old lock issue. That said, I do want a sanity limit on concurrency here, some improved retries/timeouts, the memoization that other code assumes, and some optionality in what to download (we don't need receipts of all blocks right away, only for the start of a window). I'll update #187 with details, and maybe open a PR tomorrow.

PR #224, opened 2022-03-09
Add codeowners file

trianglesphere:
**Description**
Adds a codeowners file to enable auto review assignment.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,0 +1,7 @@
+# Lines starting with '#' are comments.
+# Each line is a file pattern followed by one or more owners.
+
+*                    @protolambda @norswap @trianglesphere @maurelian @karlfloersch @ben-chain @cfromknecht @mslipper @tynes
+/l2os/               @protolambda @norswap @trianglesphere @cfromknecht @mslipper @tynes
+/opnode/             @protolambda @norswap @trianglesphere @cfromknecht @mslipper @tynes
+/packages/contracts/ @protolambda @norswap @maurelian @karlfloersch @smartcontracts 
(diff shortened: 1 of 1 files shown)


.github/CODEOWNERS
```diff
@@ -0,0 +1,7 @@
+# Lines starting with '#' are comments.
+# Each line is a file pattern followed by one or more owners.
+
+*                    @protolambda @norswap @trianglesphere @maurelian @karlfloersch @ben-chain @cfromknecht @mslipper @tynes
+/l2os/               @protolambda @norswap @trianglesphere @cfromknecht @mslipper @tynes
+/opnode/             @protolambda @norswap @trianglesphere @cfromknecht @mslipper @tynes
+/packages/contracts/ @protolambda @norswap @maurelian @karlfloersch @smartcontracts
```
maurelian, line 7: This will automatically assign reviews accordingly right? Seems like it'll be super noisy.

protolambda: 5+ reviewers per PR by default is indeed quite noisy

trianglesphere: We need to make sure that it is configured to select 1 or 2 reviewers (can be done somewhere in github), but yes, the intention is to automatically select reviewers.

trianglesphere: https://docs.github.com/en/organizations/organizing-members-into-teams/managing-code-review-settings-for-your-team

maurelian: IMO we should just choose one or two ppl who can actually _own_ each area of the repo. Either those people can review the pr, or will take responsibility for finding someone who can.

mslipper:
I'm very much in favor of having codeowners. However, I'll echo @protolambda's feedback that 5+ owners per area of the repo will produce too much noise. We also want the codeowners to actually perform the review, and having lots of code owners makes it really easy to ignore the review request because "someone else will do it." I think reducing it to 2 people per area is the best place for us to start.

PR #237, opened 2022-03-10
ref impl: Try to clean up source interfaces

trianglesphere:
**Description**

In particular, I was trying to reduce the number of interfaces and adapters
between the eth client API and high level usage of that API.

There are still lots of interfaces, but it is pared down.

Downloader
	- Provides: Receipts, Transactions, Blocks from L1
	- Depends: L1 Client
Chain Source
	- Provides: Block references
	- Depends: L1 and L2 Clients
Engine API
	- Provides: Block execution & fork selection
	- Depends: L2 Client

Driver State
	- Provides: Sequence Window
	- Depends: Chain Source, L1 head sub (via L1 client)
	- Note: Remove the polling & could wrap chain source in sync algo
Driver Step
	- Provides: Creation & Execution of block payloads
	- Depends: Downloader, Engine API, L2 Client

**To Do Changes**
- Clean up the Engine API in the L2 package - mixes definitions + several wrappers of `ExecutePayload` / `ForkChoiceUpdate`
- I don't love how everything is layered on top of everything.
- Things are also kinda scattered everywhere
- Different PR: I know how to turn the FindL1Range into something that can be cached.

**Additional context**

**Metadata**
- Fixes #[Link to Issue]


trianglesphere:
Diagram showing how the components are connected.
image_50334721

opnode/l1/downloader.go
```diff
@@ -11,24 +12,34 @@ import (
...
 	TransactionReceipt(context.Context, common.Hash) (*types.Receipt, error)
 }
 
+// Block
+// Receipts
+// Transactions (from range of blocks)
+
 type Downloader struct {
-	client EthClient
+	client DownloaderClient
```
norswap, line 31: Why do we need to wrap this in a struct, do we plan to add things in there in the future?

opnode/chain/source.go
```diff
@@ -0,0 +1,84 @@
...
+	L2BlockRefByHash(ctx context.Context, l2Hash common.Hash, genesis *rollup.Genesis) (eth.L2BlockRef, error)
+}
+
+var WrongChainErr = errors.New("wrong chain")
+var TooDeepReorgErr = errors.New("reorg is too deep")
+
+// TODO: Make configurable
+var MaxReorgDepth = 500
+
+// L1ChainClient is the subset of methods that Chain needs to determine the L1 block graph
```
norswap, line 30: Subset of methods of ... `ethclient.Client`  I assume? Same for L2ChainClient.

opnode/l1/downloader.go
trianglesphere: It does enable us to add things in the future (like a cache), but the primary thing that it enables is exposing different function signatures.

opnode/chain/source.go
trianglesphere: yep

norswap:
This still feels very complex to me.

While I appreciate, the diagram doesn't really help (mostly because the arrow mean a lot of different things).

Here's a very ugly, and unpolished tracking of my understanding of the various interfaces:

- interface `chain.ChainSource`
    - provides access to the L1 and L2 block graph
    - is implemented by type `chain.Source`
        - which wraps one instance of `L1ChainClient` and one instance of `L2ChainClients
            - interface `L1ChainClient`
                - is a subset of the methods `ethclient.Client` needed to determine L1 block graph
            - interface `L2ChainClient`
                - is a subset of the methods `ethclient.Client` needed to determine L2 block graph
- interface `DownloaderClient`
    - lets us download blocks and receipts
    - is a subset of the methods `ethclient.Client`
    - is wrapped within type `Downloader`
- type `Driver`
    - encapsulates a `state`
        - `state` contains all of the rollups state
        - ... and a `ChainSource` and `outputInterface` implementation
            - `outputInterface` is implemented in `outputImpl`
                - which wraps a `Downloader` and an `L2Client`
                    - `L2Client` is an interface for `BlockByHash`
                        - so a subset of `ethclient.Client`
                        
So at present we have no less than 4 subset interfaces of `ethclient.Client`.

Honestly, I'm just wondering why we don't use a single interface that has all the `ethclient.Client` interfaces directly.
Stick one implementation of that for the L1 node and one for the L2 node into the driver state directly, pass it down as needed.

Sure, it's more verbose to mock, but if that's a real concern then we can literally take one file to make an implem that is a struct with one function pointer per method, with default erroring implementations.

Is there any reason not to do this?

---

My second concern is that I've sort of lost the plot on downloader vs chainsource.

I'm guessing we're pulling things that can change directly from the chain source, the rest from the downloader? Is that usage consistent?

A better way of doing things would be to have an implementation of our "god interface" (the one in front of `ethclient.Client`) that performs caching as necessary. Probably only for cases where we want to get things by hashes, and then we'd try to use that as much as possible (a big reason being that we don't want to put cache invalidation there, as a big part of our algorithms which track L1 reorgs are basically all about "cache invalidation" on block numbers).

(I will say I have much less confidence / strong feelings in this point than in the previous one.)

trianglesphere:
> mostly because the arrow mean a lot of different things

The arrows specifically mean direct dependency. The exact set of functions that the higher level depends upon the lower level having does vary.


> Honestly, I'm just wondering why we don't use a single interface that has all the ethclient.Client interfaces directly.
Stick one implementation of that for the L1 node and one for the L2 node into the driver state directly, pass it down as needed.

That is tempting, and probably a good option. The reason to maybe not do it is that we still have all of the methods, just in a single implementation.  We need some amount of wrapping b/c we rely on `eth.BlockIDs` everywhere instead of the native headers/blocks (but this does make testing easier in places - the struct is much smaller to mock)

-------

Re downloader vs chain source: the chain source exposes the wrapped structures (`eth.BlockID`, `eth.L1BlockRef`, `eth.L2BlockRef`) while the downloader is doing bulk gets over the `ethclient` interface (i.e. `Receipts` from a block, `transactions` in a range of blocks)

norswap:
> That is tempting, and probably a good option. The reason to maybe not do it is that we still have all of the methods, just in a single implementation. We need some amount of wrapping b/c we rely on eth.BlockIDs everywhere instead of the native headers/blocks (but this does make testing easier in places - the struct is much smaller to mock)

Yeah, I think it's fine to make a single wrapper interface instead.

> Re downloader vs chain source: the chain source exposes the wrapped structures (eth.BlockID, eth.L1BlockRef, eth.L2BlockRef) while the downloader is doing bulk gets over the ethclient interface (i.e. Receipts from a block, transactions in a range of blocks)

So is the chainsource always "in front" of the downloader?

trianglesphere:
> So is the chainsource always "in front" of the downloader?

It's off to the side, but could be made to be in front. the block by number methods should not be cached, which limits how much the chainsource can effective cache (but I've got an idea to switch to block by hash as the bulk of the gets).

trianglesphere:
Closing. #256 makes some L2 changes to solve this and #260 fixes the rest.

trianglesphere:
Closing. https://github.com/ethereum-optimism/optimistic-specs/pull/256 makes some L2 changes to solve this and https://github.com/ethereum-optimism/optimistic-specs/pull/260 fixes the rest.

PR #238, opened 2022-03-10
specs: Add L2 output submitter service

maurelian:
**Description**

Adds a spec of the L2OS implemented in #179.

### DIFF
--- a/packages/contracts/contracts/L1/MockL2OutputOracle.sol
+++ b/packages/contracts/contracts/L1/MockL2OutputOracle.sol
@@ -5,7 +5,7 @@ pragma solidity >=0.8.10;
  * @title MockL2OutputOracle
  */
 contract MockL2OutputOracle {
-    uint256 public submissionFrequency;
+    uint256 public submissionInterval;
     uint256 public l2BlockTime;
     mapping(uint256 => bytes32) public l2Outputs;
     uint256 public historicalTotalBlocks;
@@ -14,7 +14,7 @@ contract MockL2OutputOracle {
 
     /**
      * Initialize the MockL2OutputOracle contract.
-     * @param _submissionFrequency The desired interval in seconds at which
+     * @param _submissionInterval The desired interval in seconds at which
      *        checkpoints must be submitted.
      * @param _l2BlockTime The desired L2 inter-block time in seconds.
      * @param _genesisL2Output The initial L2 output of the L2 chain.
@@ -22,12 +22,12 @@ contract MockL2OutputOracle {
      *        initialization of the L2 chain.
      */
     constructor(
-        uint256 _submissionFrequency,
+        uint256 _submissionInterval,
         uint256 _l2BlockTime,
         bytes32 _genesisL2Output,
         uint256 _historicalTotalBlocks
     ) {
-        submissionFrequency = _submissionFrequency;
+        submissionInterval = _submissionInterval;
         l2BlockTime = _l2BlockTime;
         l2Outputs[block.timestamp] = _genesisL2Output; // solhint-disable not-rely-on-time
         historicalTotalBlocks = _historicalTotalBlocks;
@@ -56,7 +56,7 @@ contract MockL2OutputOracle {
      * checkpointed.
      */
     function nextTimestamp() public view returns (uint256) {
-        return latestBlockTimestamp + submissionFrequency;
+        return latestBlockTimestamp + submissionInterval;
     }
 
     /**
--- a/packages/contracts/contracts/L1/MockL2OutputOracle.sol
+++ b/packages/contracts/contracts/L1/MockL2OutputOracle.sol
@@ -46,6 +46,7 @@ contract MockL2OutputOracle {
         require(block.timestamp > _timestamp, "Cannot append L2 output in future");
         require(_l2Output != bytes32(0), "Cannot submit empty L2 output");
         require(_timestamp == nextTimestamp(), "Timestamp not equal to next expected timestamp");
+        // todo: add require statement to ensure a specific prev-hash exists on the current chain
         l2Outputs[_timestamp] = _l2Output;
         latestBlockTimestamp = _timestamp;
     }
--- a/specs/README.md
+++ b/specs/README.md
@@ -8,7 +8,6 @@ that maintains 1:1 compatibility with Ethereum.
 
 - [Glossary](glossary.md)
 - [Deposits](deposits.md)
-- [Withdrawals](withdrawals.md)
 - [Execution Engine](exec-engine.md)
 - [L2 Output Root Proposals](proposals.md)
 - [Rollup Node](rollup-node.md)
--- a/specs/README.md
+++ b/specs/README.md
@@ -8,8 +8,9 @@ that maintains 1:1 compatibility with Ethereum.
 
 - [Glossary](glossary.md)
 - [Deposits](deposits.md)
+- [Withdrawals](withdrawals.md)
 - [Execution Engine](exec-engine.md)
-- [L2 output root Proposals](proposals.md)
+- [L2 Output Root Proposals](proposals.md)
 - [Rollup Node](rollup-node.md)
 
 ## Design Goals
--- a/specs/glossary.md
+++ b/specs/glossary.md
@@ -37,6 +37,7 @@
   - [L2 Derivation Inputs](#l2-derivation-inputs)
   - [Payload Attributes](#payload-attributes)
   - [L1 Attributes Predeployed Contract](#l1-attributes-predeployed-contract)
+  - [L2 output](#l2-output)
 - [Execution Engine Concepts](#execution-engine-concepts)
   - [Execution Engine](#execution-engine)
 
@@ -366,6 +367,14 @@ A [predeployed contract][predeploy] on L2 that can be used to retrieve the L1 bl
 block number or a given block hash.
 
 cf. [L1 Attributes Predeployed Contract Specification](deposits.md#l1-attributes-predeployed-contract)
+
+## L2 output
+
+[l2-output]: glossary.md#l2-output
+
+A 32 byte value representing the current state of the L2 chain.
+
+cf. [Proposing L2 output commitments](proposals.md#l2-output-root-proposals-specification)
 
 ------------------------------------------------------------------------------------------------------------------------
 
--- a/specs/glossary.md
+++ b/specs/glossary.md
@@ -37,7 +37,7 @@
   - [L2 Derivation Inputs](#l2-derivation-inputs)
   - [Payload Attributes](#payload-attributes)
   - [L1 Attributes Predeployed Contract](#l1-attributes-predeployed-contract)
-  - [L2 output](#l2-output)
+  - [L2 Output Root](#l2-output-root)
 - [Execution Engine Concepts](#execution-engine-concepts)
   - [Execution Engine](#execution-engine)
 
@@ -368,7 +368,7 @@ block number or a given block hash.
 
 cf. [L1 Attributes Predeployed Contract Specification](deposits.md#l1-attributes-predeployed-contract)
 
-## L2 output
+## L2 Output Root
 
 [l2-output]: glossary.md#l2-output
 
--- a/specs/glossary.md
+++ b/specs/glossary.md
@@ -372,7 +372,7 @@ cf. [L1 Attributes Predeployed Contract Specification](deposits.md#l1-attributes
 
 [l2-output]: glossary.md#l2-output
 
-A 32 byte value representing the current state of the L2 chain.
+A 32 byte value which commits to the current state of the L2 chain.
 
 cf. [Proposing L2 output commitments](proposals.md#l2-output-root-proposals-specification)
 
--- a/specs/proposals.md
+++ b/specs/proposals.md
@@ -105,8 +105,8 @@ upgrades. This keeps the static merkle structure forwards-compatible.
 ## L2 Output Oracle Smart Contract
 
 L2 blocks are produced at a constant rate of `L2_BLOCK_TIME` (2 seconds).
-A new L2 output MUST be appended to the chain once per `SUBMISSION_INTERVAL` (1800 seconds). Note that this interval is based
-on L2 time. It is OK to have L2 outputs submitted at larger or small intervals.
+A new L2 output MUST be appended to the chain once per `SUBMISSION_INTERVAL` (1800 seconds). Note that this interval is\
+based on L2 time. It is OK to have L2 outputs submitted at larger or small intervals.
 
 The L2 Output Oracle contract implements the following interface:
 
--- a/specs/proposals.md
+++ b/specs/proposals.md
@@ -1,4 +1,4 @@
-# L2 output root Proposals Specification
+# L2 Output Root Proposals Specification
 
 <!-- All glossary references in this file. -->
 [g-rollup-node]: glossary.md#rollup-node
@@ -7,8 +7,12 @@
 <!-- DON'T EDIT THIS SECTION, INSTEAD RE-RUN doctoc TO UPDATE -->
 **Table of Contents**
 
-- [Proposing L2 output commitments](#proposing-l2-output-commitments)
-- [L2 output commitment construction](#l2-output-commitment-construction)
+- [Constants](#constants)
+- [Proposing L2 Output Commitments](#proposing-l2-output-commitments)
+- [L2 Output Commitment Construction](#l2-output-commitment-construction)
+- [L2 Output Oracle Smart Contract](#l2-output-oracle-smart-contract)
+- [Security Considerations](#security-considerations)
+  - [L1 Reorgs](#l1-reorgs)
 
 <!-- END doctoc generated TOC please keep comment here to allow auto update -->
 
@@ -24,15 +28,39 @@ are part of later specification milestones.
 
 [cannon]: https://github.com/ethereum-optimism/cannon
 
-## Proposing L2 output commitments
+## Definitions
 
-The proposer's role is to construct and submit output commitments on a configurable interval to a contract on L1.
+### Constants
 
-TODO: describe integration with rollup node and L2 execution engine (see PR #179).
+| Name                   | Value | Unit    |
+| ---------------------- | ----- | ------- |
+| `SUBMISSION_FREQUENCY` | `100` | seconds |
+| `L2_BLOCK_TIME`        | `2`   | seconds |
 
-TODO: link to contract specification/source of L2 Output oracle on L1.
+### Types
 
-## L2 output commitment construction
+The `ForkSpec` type contains the height and blockhash for a block in the L1 chain.
+
+```js
+struct ForkSpec {
+    uint256 blockHeight;
+    bytes32 blockHash;
+}
+```
+
+## Proposing L2 Output Commitments
+
+The proposer's role is to construct and submit output commitments on a configurable interval to a contract on  , which
+it does by running the [L2 output submitter](../l2os/) service (AKA L2OSS). This service periodically queries the rollup
+ node's [`optimism_outputAtBlock` rpc method](./rollup-node.md#l2-output-rpc-method) for the latest output root derived
+ from the latest [finalized](rollup-node.md#finalization-guarantees) L1 block. The construction of this output root is
+ described [below](#l2-output-commitment-construction).
+
+If there is no newly finalized output, the service continues querying until it receives one. It then submits this
+output, and the appropriate timestamp, to the [L2 Output Commitment](#l2-output-commitment-smart-contract) contract's
+`appendL2Output()` function. The timestamp MUST be the next multiple of the `SUBMISSION_FREQUENCY` value.
+
+## L2 Output Commitment Construction
 
 This merkle-structure is defined with [SSZ], a type system for merkleization and serialization, used in
 L1 (beacon-chain). However, we replace `sha256` with `keccak256` to save gas costs in the EVM.
@@ -42,7 +70,7 @@ L1 (beacon-chain). However, we replace `sha256` with `keccak256` to save gas cos
 ```python
 class L2Output(Container):
   state_root: Bytes32
-  withdrawal_storage_root: Bytes32  # TODO: withdrawals specifcation work-in-progress
+  withdrawal_storage_root: Bytes32  # TODO: withdrawals specification work-in-progress
   latest_block: ExecutionPayload  # includes block hash
   history_accumulator_root: Bytes32  # Not functional yet
   extension: Bytes32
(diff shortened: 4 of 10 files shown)


specs/rollup-node.md
```diff
@@ -268,6 +270,25 @@ previous section to each successive L1 block until we have caught up with the L1
...
+The input and return types here are as defined by the [engine API specs][engine-structures]).
+
+[engine-structures]: https://github.com/ethereum/execution-apis/blob/main/src/engine/specification.md#structures
+
+- method: `optimism_outputAtBlock`
+- params:
+  1. `QUANTITY` - L1 integer block number, or the strings `"safe"`, `"latest"`, or `"pending"`
+- returns:
+  2.`DATA` - The 32 byte output root
+
```
maurelian, line 291: Design questions:
- is there an argument that this should go on the execution engine? I prefer it on the rollup node because it can handle evaluating the fork choice.
- This is the first rollup node specific RPC method to be spec'd afaik. Is `optimism_` the right method prefix here?
- Should we allow querying this by either an L2 or L1 block height? If so, should we do that by:
  - having two optional params (`l1block` and `l2block`) or
  - two methods (`optimism_outputAtL1Block`, `optimism_outputAtL2Block`?

specs/rollup-node.md
```diff
@@ -252,7 +254,7 @@ L1 Ethereum [reaches finality approximately every 12 minutes][l1-finality]. L2 b
 are "safer" than most recent L2 blocks because they will never disappear from the chain's history because of a re-org.
 However, they can still be challenged by a fault proof until the end of the fault proof window.
 
-[l1-finality]: https://www.paradigm.xyz/2021/07/ethereum-reorgs-after-the-merge/
+[l1-finality]: https://www.paradigm.xyz/2021/07/ethereum-reorgs-after-the-merge
```
maurelian, line 257: I made a minor edit just so I could comment here. 
1. I don't think this is a great resource to reference. It's probably not being kept up to date.
2. The [specs](https://github.com/ethereum/consensus-specs/blob/v1.0.0/specs/phase0/beacon-chain.md#time-parameters) give 6.4 minutes per epoch, so 12.8 minutes to finality. We should be precise.
3. This does not apply until after the merge. I think what we should do is try to estimate an L1 block depth that would have similar economic penalties to undoing a post-merge finalized consensus block.

Would appreciate thoughts on this tho.

specs/proposals.md
```diff
@@ -26,11 +26,14 @@ are part of later specification milestones.
...
 
-The proposer's role is to construct and submit output commitments on a configurable interval to a contract on L1.
+The proposer's role is to construct and submit output commitments on a configurable interval to a contract on L1, which
+it does by running the [L2 output submitter](../l2os/) service (AKA L2OSS). This service periodically queries the rollup node's
+[`optimism_outputAtBlock` rpc method](./rollup-node.md#l2-output-rpc-method) for the latest output root derived from the latest [finalized](rollup-node.md#finalization-guarantees)
+L1 block. The construction of this output root is described [below](#l2-output-commitment-construction).
 
-TODO: describe integration with rollup node and L2 execution engine (see PR #179).
-
-TODO: link to contract specification/source of L2 Output oracle on L1.
```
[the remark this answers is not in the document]
maurelian: just a test comment for slack notifs

specs/rollup-node.md
maurelian: I talked to @karlfloersch. The conclusion was: 
1. No.
2. Sure.
3. It should be queried by the l2Block number. No need to have multiple methods.

specs/glossary.md
```diff
@@ -338,6 +339,14 @@ block number or a given block hash.
 
 cf. [L1 Attributes Predeployed Contract Specification](deposits.md#l1-attributes-predeployed-contract)
 
+## L2 output
```
norswap, line 342: Pedantic: should this be "L2 Output Root"?

specs/proposals.md
```diff
@@ -24,13 +26,23 @@ are part of later specification milestones.
 
 [cannon]: https://github.com/ethereum-optimism/cannon
 
-## Proposing L2 output commitments
+## Constants
+
+| Name                   | Value  |
+| ---------------------- | ------ |
+| `SUBMISSION_FREQUENCY` | `100?` |
```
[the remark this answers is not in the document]
norswap: It actually annoys me haha, because I have no context to understand them.

But it's great to have them listed together somewhere, for reference!

Suggestion:
- move this to the end
- recall the default value in the text, so that someone that reads the spec does not have to scroll up/down to see what the expected value is

specs/proposals.md
```diff
@@ -24,13 +27,23 @@ are part of later specification milestones.
```
norswap, line 27: Reminds me I have a local branch where I attempted to reword this document!
Let's get this merged, and I'll rebase on top of that and make another PR.

specs/proposals.md
```diff
@@ -24,13 +27,23 @@ are part of later specification milestones.
...
 
-The proposer's role is to construct and submit output commitments on a configurable interval to a contract on L1.
+## Proposing L2 output commitments
 
-TODO: describe integration with rollup node and L2 execution engine (see PR #179).
+The proposer's role is to construct and submit output commitments on a configurable interval to a contract on L1, which
+it does by running the [L2 output submitter](../l2os/) service (AKA L2OSS). This service periodically queries the rollup
+ node's [`optimism_outputAtBlock` rpc method](./rollup-node.md#l2-output-rpc-method) for the latest output root derived
+ from the latest [finalized](rollup-node.md#finalization-guarantees) L1 block. The construction of this output root is
+ described [below](#l2-output-commitment-construction).
```
norswap, line 42: Pedantic: is it really necessary to call it a "service"? "L2 output submitted" seems quite descriptive enough to me, not sure what "service" adds.

specs/proposals.md
```diff
@@ -71,3 +83,15 @@ This is a work-in-progress, see [issue 181](https://github.com/ethereum-optimism
...
+
+> **TODO** spec for the [L2OutputOracle contract]
+[L2OutputOracle contract]: <https://github.com/ethereum-optimism/optimistic-specs/blob/8c1d509e3882ca49f2415c2b724c8a58a134cec7/packages/contracts/contracts/L1/MockL2OutputOracle.sol#L1>
+
+## Security Considerations
+
+> **TODO**
+>
+> - list invariants
+> - list things that could go wrong and ways to avoid that
```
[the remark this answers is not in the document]
norswap: @cfromknecht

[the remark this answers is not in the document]
norswap: One thing that could go wrong is submitting an output root straight after the chain re-orged. This is guarded against by including the target L1 (or L2?) blockhash along with the output root. Maybe something that should figure in this section?

specs/rollup-node.md
norswap: I think realistically Bedrock is going to go live after the merge, unless something bad and unforeseen happens. So I'm fine with speccing with the merge in mind, rather than having to do extra work which we then need to rip out.

I think this is better than no resources, but I think something more accurate would be much better indeed. Maybe referencing the L2 specs directly if there's a section we can point to?

Also the fact that finalization is 12.8 minutes but ONLY if 2/3 of nodes are online/operating is an important caveat.

specs/rollup-node.md
```diff
@@ -268,6 +270,25 @@ previous section to each successive L1 block until we have caught up with the L1
 
 > **TODO** specify genesis block
 
+## L2 Output RPC method
+
+The Rollup node has its own RPC method, `optimism_output` which returns the
```
norswap, line 275: `optimism_outputAtBlock` right?

specs/rollup-node.md
```diff
@@ -268,6 +270,25 @@ previous section to each successive L1 block until we have caught up with the L1
...
+a 32 byte hash corresponding to the [SSZ] encoded [L2Output](./proposals.md#l2-output-commitment-construction).
+
+[SSZ]: https://github.com/ethereum/consensus-specs/blob/dev/ssz/simple-serialize.md
+
+### Output Method API
```
norswap, line 280: This is pretty short, really worth splitting into a sub-section?

specs/proposals.md
```diff
@@ -7,8 +7,11 @@
 <!-- DON'T EDIT THIS SECTION, INSTEAD RE-RUN doctoc TO UPDATE -->
 **Table of Contents**
 
+- [Constants](#constants)
 - [Proposing L2 output commitments](#proposing-l2-output-commitments)
 - [L2 output commitment construction](#l2-output-commitment-construction)
+- [L2 Output Commitment Smart Contract](#l2-output-commitment-smart-contract)
+- [Security Considerations](#security-considerations)
```
norswap, line 14: OCD: Make casing consistent accross titles.

specs/proposals.md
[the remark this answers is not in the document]
maurelian: good point. This came up in discussion with Karl. I will add that protection.

specs/proposals.md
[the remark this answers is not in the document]
maurelian: I'd be onboard with this. I'd like to review the full doc for this kind of thing, which I'm tracking here and you might wish to comment on: #252

specs/proposals.md
maurelian: Mostly I don't like the term 'OS' it confuses me.

specs/rollup-node.md
maurelian: Updated to reference both this post, and the [eth2 specs](https://github.com/ethereum/consensus-specs/blob/v1.0.0/specs/phase0/beacon-chain.md#time-parameters).

specs/proposals.md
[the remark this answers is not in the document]
maurelian: > * move this to the end

Actually, I'd prefer to keep it at the top, might as well follow EIP convention.

tynes:
Has there been discussion around why the next blocks timestamp can be computed exactly? This weakens any existing RNG that seeds using `block.timestamp` - could we instead have a range for the next timestamp?

I suppose a second order effect of having a range would give more power to the sequencer, since it could modify the timestamp out of band and have a flashbots like auction for what the timestamp should be. Having it be known what the next timestamp is would make that knowledge of the next timestamp common knowledge

packages/contracts/contracts/L1/MockL2OutputOracle.sol
```diff
@@ -46,6 +46,7 @@ contract MockL2OutputOracle {
         require(block.timestamp > _timestamp, "Cannot append L2 output in future");
         require(_l2Output != bytes32(0), "Cannot submit empty L2 output");
         require(_timestamp == nextTimestamp(), "Timestamp not equal to next expected timestamp");
+        // todo: add require statement to ensure a specific prev-hash exists on the current chain
         l2Outputs[_timestamp] = _l2Output;
         latestBlockTimestamp = _timestamp;
```
tynes, line 51: No event emitted?

packages/contracts/contracts/L1/MockL2OutputOracle.sol
```diff
@@ -55,7 +56,7 @@ contract MockL2OutputOracle {
      * checkpointed.
      */
     function nextTimestamp() public view returns (uint256) {
```
tynes, line 58: What do you think about having an internal `_nextTimestamp` that this wraps and then call the internal function in `appendL2Output`?

tynes:
> > Has there been discussion around why the next blocks timestamp can be computed exactly?
> 
> It's a UX design goal to have a constant block time of 2 seconds, as this ensures that the sequencer always provide a tx confirmation to the user.
> 
> Interesting point about "randomness", although that was never a good practice, this would really break it as there will be only 1 or 2 possible timestamps for a given tx.

Have we discussed the concept of "lazy block production" where blocks are only produced if transactions exist? This would likely add a lot of complexity so therefore likely best to skip, but would be ideal for not producing empty blocks and wasting disk space

packages/contracts/contracts/L1/MockL2OutputOracle.sol
```diff
@@ -5,7 +5,7 @@ pragma solidity >=0.8.10;
  * @title MockL2OutputOracle
  */
 contract MockL2OutputOracle {
```
tynes, line 7: I see this is a `MockL2OutputOracle` - it doesn't have the concept of auth for who can submit batches, is this functionality being tracked/specified?

specs/proposals.md
```diff
@@ -24,15 +28,39 @@ are part of later specification milestones.
...
-## Proposing L2 output commitments
+## Definitions
 
-The proposer's role is to construct and submit output commitments on a configurable interval to a contract on L1.
+### Constants
 
-TODO: describe integration with rollup node and L2 execution engine (see PR #179).
+| Name                   | Value | Unit    |
+| ---------------------- | ----- | ------- |
+| `SUBMISSION_FREQUENCY` | `100` | seconds |
```
karlfloersch, line 37: I think submission frequency should be more like `1800` seconds. The difference between withdrawals happening in 7 days vs in 7 days and 30 mins makes the $ cost of submission more of a limiting factor here.

If we had a shorter withdrawal period then reducing this number might matter, but until then I think 30 mins or more is super reasonable.

maurelian:
Would empty blocks still have a number? Like in L1, we want to maintain the property that there is a constant number of blocks per period of time.

specs/proposals.md
norswap: Fair point but maybe it's better to rename `l2os` to `l2output`?

norswap:
@tynes 

> Has there been discussion around why the next blocks timestamp can be computed exactly? This weakens any existing RNG that seeds using block.timestamp - could we instead have a range for the next timestamp?

afaik, this will be the case on L1 post-merge too.

packages/contracts/contracts/L1/MockL2OutputOracle.sol
norswap: This won't stay a mock. I hope it's tracked but haven't checked!

As for speccing the auth, I think there won't be much more to say at first than "there is an address owning this". Anything more you'd like to suggest there?

norswap:
> Have we discussed the concept of "lazy block production" where blocks are only produced if transactions exist? This would likely add a lot of complexity so therefore likely best to skip, but would be ideal for not producing empty blocks and wasting disk space

Block production is already lazy as far as batch go (not batches submitted for empty blocks) but not in the execution engine.
Imho this is an optimization — i.e. it can be implemented in the execution engine without changing any of the interfaces that we expose, and so does not need to be a spec concern.

packages/contracts/contracts/L1/MockL2OutputOracle.sol
tynes: Whoever submits batches should have a deposit that is slashable if the fraud proof executes and the proposer is shown to have submitted fraudulent data. Right now it looks like anybody can append an L2 output with no mechanism to prevent liars

norswap: That will be added when Cannon goes live, right now it's permissioned (which is what I meant by "there is an address owning this", should have expressed that better).

specs/proposals.md
maurelian: just going with l2os so we can move forward

packages/contracts/contracts/L1/MockL2OutputOracle.sol
maurelian: I tested this out, and it appears to make no zero difference, which is nice actually.

packages/contracts/contracts/L1/MockL2OutputOracle.sol
maurelian: I will also make that change in a separate PR.

maurelian: Will be done here: #273

specs/proposals.md
```diff
@@ -24,15 +30,39 @@ are part of later specification milestones.
...
+struct ForkSpec {
+    uint256 blockHeight;
+    bytes32 blockHash;
+}
+```
+
+## Proposing L2 Output Commitments
+
+The proposer's role is to construct and submit output commitments on a configurable interval to a contract on , which
+it does by running the [L2 output submitter](../l2os/) (AKA L2OS). This service periodically queries the rollup
```
tynes, line 56: When I hear L2OS i think of "layer 2 operating system"

maurelian: Same. WDYT about L2OSS, for output submitter service?
Otherwise open to suggestions.

norswap: Personally I like to avoid acronyms unless it's really cumbersome or already in common use, so we could just use "L2 output submitter" everywhere. It is annoying for the package name, but there my suggestion is `l2output`.

Not a hill I'm going to die on, but I suggest discussing this on a call someday, reaching a decision and moving on.

specs/proposals.md
```diff
@@ -26,11 +26,14 @@ are part of later specification milestones.
 
 ## Proposing L2 output commitments
 
-The proposer's role is to construct and submit output commitments on a configurable interval to a contract on L1.
+The proposer's role is to construct and submit output commitments on a configurable interval to a contract on L1, which
+it does by running the [L2 output submitter](../l2os/) service (AKA L2OSS). This service periodically queries the rollup node's
+[`optimism_outputAtBlock` rpc method](./rollup-node.md#l2-output-rpc-method) for the latest output root derived from the latest [finalized](rollup-node.md#finalization-guarantees)
+L1 block. The construction of this output root is described [below](#l2-output-commitment-construction).
 
-TODO: describe integration with rollup node and L2 execution engine (see PR #179).
```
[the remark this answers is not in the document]
tynes: I feel like we can leave the inner workings as implementation details for now, it should be straight forward to add in the details in the future if necessary. We should prioritize adding them now if they will make a significant difference for the audit

specs/glossary.md
```diff
@@ -367,6 +368,14 @@ block number or a given block hash.
 
 cf. [L1 Attributes Predeployed Contract Specification](deposits.md#l1-attributes-predeployed-contract)
 
+## L2 Output Root
+
+[l2-output]: glossary.md#l2-output
+
+A 32 byte value representing the current state of the L2 chain.
```
tynes, line 375: I think that the phrase "commits to" is better than "representing", ie "A 32 byte value that commits to the current state of the L2 chain"

And I believe that we did not opt to have it commit to more values than just the state trie? We should be clear here with what exactly it commits to.

Also is there a versioning scheme? ie so that it can be updated to verkle trees easily or commit to a merkle tree that includes multiple interesting things like @protolambda has suggested

specs/proposals.md
tynes: L2OSS -> L2 Open Source Software lol

L2 output submitter submitter I think is nice, also not opposed to just using what we use now (batch submitter)

specs/rollup-node.md
```diff
@@ -347,6 +351,25 @@ block forward each step, until there is an insufficient number of L1 blocks left
...
+The input and return types here are as defined by the [engine API specs][engine-structures]).
+
+[engine-structures]: https://github.com/ethereum/execution-apis/blob/main/src/engine/specification.md#structures
+
+- method: `optimism_outputAtBlock`
+- params:
+  1. `QUANTITY` - L2 integer block number, or the strings `"safe"`, `"latest"`, or `"pending"`
+- returns:
+  1. `DATA` - The 32 byte output root
+
```
tynes, line 372: I think that this should either accept a "type" param or return JSON that includes data and type so that we can upgrade the commitments more easily, cc @protolambda

tynes: We could embed the type enum in the first byte and just use typed hashes

tynes:
This might be really dumb, but where is the equivalent to the CTC? ie where are the transaction batches sent?

maurelian:
Not dumb, they are sent to an address with no code, which is weird at first. But makes sense since the rollup node can just read those transactions right from the block.

specs/glossary.md
maurelian: It will initially commit to:
- the block
- the state trie
- the storage root of the withdrawals predeploy
- the version

There is a more complex output [currently specified](https://github.com/ethereum-optimism/optimistic-specs/blob/429079094899becdf4d37921938650d2f8d8b139/specs/proposals.md#l2-output-commitment-construction), but I plan to simplify it to the above in #253.

specs/rollup-node.md
maurelian: I'm fine with it.

specs/rollup-node.md
maurelian: How is this?

```
- method: `optimism_outputAtBlock`
- params:
  1. `TYPE` - A versioning enum which describes the data structure used to generate the output root. Defaults to the latest version, which is currently '0'
  1. `QUANTITY` - L2 integer block number, or the strings `"safe"`, `"latest"`, or `"pending"`
- returns:
  1. `DATA` - The 32 byte output root
```

Also, should it return just the output root, or all the leaves? 

Should you also be able to get a proof from an `optimism_` method on the rollup node?

specs/README.md
```diff
@@ -8,8 +8,9 @@ that maintains 1:1 compatibility with Ethereum.
 
 - [Glossary](glossary.md)
 - [Deposits](deposits.md)
+- [Withdrawals](withdrawals.md)
```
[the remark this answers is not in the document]
maurelian: Good catch. It's added in #253. Will remove from this PR.

specs/proposals.md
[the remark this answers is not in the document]
maurelian: > We should prioritize adding them now if they will make a significant difference for the audit

I don't think so.

specs/proposals.md
maurelian: Going to handle this for now by just removing the "AKA" note.

specs/rollup-node.md
norswap: Shouldn't the type be a return value? The sequencer choses the type of the commitment he submits after all, the user doesn't get a say.

The submitter should also specify this output type. We may want to validate the output type in the contract (i.e. probably just check that it is the latest known type) — although that will also naturally be done in the rollup-node / minigeth (i.e. minigeth will always derive the correct ouptut root type, so submitting an old output root type will be contestable via a fault proof). Still a bit of redundant security doesn't hurt.

maurelian: Yes, you're totally right.

protolambda (changes requested):
Looks good, but the JSON RPC definition is not defined correctly w.r.t. expected typing and naming.

specs/rollup-node.md
```diff
@@ -347,6 +351,25 @@ block forward each step, until there is an insufficient number of L1 blocks left
...
+The input and return types here are as defined by the [engine API specs][engine-structures]).
+
+[engine-structures]: https://github.com/ethereum/execution-apis/blob/main/src/engine/specification.md#structures
+
+- method: `optimism_outputAtBlock`
+- params:
+  1. `QUANTITY` - L2 integer block number, or the strings `"safe"`, `"latest"`, or `"pending"`
+- returns:
+  1. `DATA` - The 32 byte output root
```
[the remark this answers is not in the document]
maurelian: OK, I updated it using the [execution API](https://github.com/ethereum/execution-apis/blob/main/src/engine/specification.md#structures) for reference, please take a look again

https://github.com/ethereum-optimism/optimistic-specs/blob/69aa7cecae9531ba281a68e3f03b0b4b77812943/specs/rollup-node.md#L361-L373

specs/rollup-node.md
tynes: A few thoughts with this design:
- The rollup node will be responsible for knowing what version to return. Since it can speak to the execution client, it has the context of the chain and can alter the version at a specific height
- The output submitter can be dumb and not need to understand what the types mean
- The contract needing to know the type means that it'll consume extra gas when submitting L2 outputs, do we need logic that says: `if blocknumber == x; then update version` - how exactly do we govern that? The contracts aren't specified to be behind delegatecall proxies right now

protolambda (approved):
LGTM

norswap (approved):
Good for me to merge. I would like the definitions moved to the end, but this doesn't need to happen in this PR.

maurelian:
@norswap I went ahead and moved that stuff to the bottom anyways.

PR #245, opened 2022-03-15
Update golang dependencies

trianglesphere:
**Description**
Primarily to update to a newer version of OP's geth client.
Also updates all other dependencies to their latest version.

See https://github.com/ethereum-optimism/reference-optimistic-geth and the `optimism-rebased` branch for what this updates the reference implementation to.


protolambda (approved):
LGTM. Updated the geth dependency to link to the single-commit diff on default branch of the fork, and updated the API bindings with the new bool flag.

PR #246, opened 2022-03-15
ref impl: Add simple batch serialization

trianglesphere:
**Description**
Adds simple batch serialization. This is the simplest version from the specs: `(epoch, timestamp, transactions)`.
It does include the version byte.

A bundle of multiple batches are encoded as an RLP list of these versioned batches, also with version byte. Version 2 may support some type of compression of the bundle of batches.

### DIFF
--- a/opnode/rollup/derive/batch.go
+++ b/opnode/rollup/derive/batch.go
@@ -3,68 +3,151 @@ package derive
 import (
 	"bytes"
 	"errors"
+	"fmt"
+	"io"
+	"sync"
 
-	"github.com/ethereum-optimism/optimistic-specs/opnode/l2"
 	"github.com/ethereum-optimism/optimistic-specs/opnode/rollup"
 	"github.com/ethereum/go-ethereum/common/hexutil"
 	"github.com/ethereum/go-ethereum/rlp"
 )
 
 // Batch format
-// first byte is type followed by bytstring.
+// first byte is type followed by bytestring.
 //
 // BatchV1Type := 0
-// BatchV1Type ++ RLP([epoch, timestamp, transaction_list]
+// batchV1 := BatchV1Type ++ RLP([epoch, timestamp, transaction_list]
+//
+// An empty input is not a valid batch.
+//
+// Batch-bundle format
+// first byte is type followed by bytestring
+//
+// payload := RLP([batch_0, batch_1, ..., batch_N])
+// bundleV1 := BatchBundleV1Type ++ payload
+// bundleV2 := BatchBundleV2Type ++ compress(payload)  # TODO: compressed bundle of batches
+//
+// An empty input is not a valid bundle.
+//
+// Note: the type system is based on L1 typed transactions.
+
+// encodeBufferPool holds temporary encoder buffers batch encoding
+var encodeBufferPool = sync.Pool{
+	New: func() interface{} { return new(bytes.Buffer) },
+}
 
 const (
 	BatchV1Type = iota
 )
 
+const (
+	BatchBundleV1Type = iota
+	BatchBundleV2Type
+)
+
 type BatchV1 struct {
-	Epoch        uint64
-	Timestamp    uint64
+	Epoch     rollup.Epoch // aka l1 num
+	Timestamp uint64
+	// no feeRecipient address input, all fees go to a L2 contract
 	Transactions []hexutil.Bytes
 }
 
 type BatchData struct {
-	Epoch     rollup.Epoch // aka l1 num
-	Timestamp uint64
-	// no feeRecipient address input, all fees go to a L2 contract
-	Transactions []l2.Data
+	BatchV1
+	// batches may contain additional data with new upgrades
 }
 
-func ParseBatch(data []byte) (BatchData, error) {
-	var v1 BatchV1
-	if err := v1.UnmarshalBinary(data); err != nil {
-		return BatchData{}, err
+func DecodeBatches(config *rollup.Config, r io.Reader) ([]*BatchData, error) {
+	var typeData [1]byte
+	if _, err := r.Read(typeData[:]); err != nil {
+		return nil, fmt.Errorf("failed to read batch bundle type byte: %v", err)
 	}
-	return BatchData{Epoch: rollup.Epoch(v1.Epoch), Timestamp: v1.Timestamp, Transactions: v1.Transactions}, nil
+	switch typeData[0] {
+	case BatchBundleV1Type:
+		var out []*BatchData
+		if err := rlp.Decode(r, &out); err != nil {
+			return nil, fmt.Errorf("failed to decode v1 batches list: %v", err)
+		}
+		return out, nil
+	case BatchBundleV2Type:
+		// TODO: implement compression of a bundle of batches
+		return nil, errors.New("bundle v2 not supported yet")
+	default:
+		return nil, fmt.Errorf("unrecognized batch bundle type: %d", typeData[0])
+	}
 }
 
-// // TODO: Is this needed?
-// // EncodeRLP implements rlp.Encoder
-// func (b *BatchV1) EncodeRLP(w io.Writer) error {
-// 	return rlp.Encode(w, b)
-// }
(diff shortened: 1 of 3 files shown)


opnode/rollup/derive/payload_attributes.go
```diff
@@ -328,14 +334,3 @@ func DeriveDeposits(l1Info L1Info, receipts []*types.Receipt) ([]l2.Data, error)
...
 }
-
-type BatchData struct {
-	Epoch     rollup.Epoch // aka l1 num
-	Timestamp uint64
-	// no feeRecipient address input, all fees go to a L2 contract
-	Transactions []l2.Data
-}
-
-func ParseBatches(data l2.Data) []BatchData {
```
protolambda, line 339: Parsing multiple batches from the same transaction call-data would be great to have. There are many L2 blocks (one batch each), we should be bundling them.

opnode/rollup/derive/batch.go
```diff
@@ -0,0 +1,70 @@
...
+func ParseBatch(data []byte) (BatchData, error) {
+	var v1 BatchV1
+	if err := v1.UnmarshalBinary(data); err != nil {
+		return BatchData{}, err
+	}
+	return BatchData{Epoch: rollup.Epoch(v1.Epoch), Timestamp: v1.Timestamp, Transactions: v1.Transactions}, nil
+}
+
+// // TODO: Is this needed?
+// // EncodeRLP implements rlp.Encoder
```
protolambda, line 45: We are currently not encoding/decoding the batch in an RLP context, so it's not right now. The geth RLP package is not aware of `MarshalBinary` / `UnmarshalBinary`, so we'll need the `EncodeRLP`/`DecodeRLP` methods when e.g. RLP-encoding a list of batches. And with support for multiple batches that's exactly what I plan to do :)

protolambda:
Updated this PR to support bundling of multiple batches, typing of that bundle, and to improve the RLP/Binary marshal methods by matching the type-transaction approach that geth implements. And tried to make it easier to implement compression, which would be typed with v2 of bundle encoding/decoding. And the rollup-config is available during batch encoding/decoding, to help decide on encoding type and maybe to e.g. parametrize the compression.

opnode/rollup/derive/batch.go
```diff
@@ -0,0 +1,153 @@
...
+// Batch-bundle format
+// first byte is type followed by bytestring
+//
+// payload := RLP([batch_0, batch_1, ..., batch_N])
+// bundleV1 := BatchBundleV1Type ++ payload
+// bundleV2 := BatchBundleV2Type ++ compress(payload)  # TODO: compressed bundle of batches
+//
+// An empty input is not a valid bundle.
+//
+// Note: the type system is based on L1 typed transactions.
```
norswap, line 32: Would it be clearer to say "the encoding of individual transactions"?

opnode/rollup/derive/batch.go
```diff
@@ -0,0 +1,153 @@
...
+// An empty input is not a valid batch.
+//
+// Batch-bundle format
+// first byte is type followed by bytestring
+//
+// payload := RLP([batch_0, batch_1, ..., batch_N])
+// bundleV1 := BatchBundleV1Type ++ payload
+// bundleV2 := BatchBundleV2Type ++ compress(payload)  # TODO: compressed bundle of batches
+//
+// An empty input is not a valid bundle.
```
norswap, line 30: What's "input" here? An empty (zero-byte) bundle? A zero-byte payload?

opnode/rollup/derive/payload_attributes.go
```diff
@@ -180,25 +181,31 @@ func UserDeposits(height uint64, receipts []*types.Receipt) ([]*types.DepositTx,
...
 				continue // bad signature, ignore
 			}
 			// some random L1 user might have sent a transaction to our batch inbox, ignore them
 			if seqDataSubmitter != config.BatchSenderAddress {
 				continue // not an authorized batch submitter, ignore
 			}
-			out = append(out, ParseBatches(tx.Data())...)
+			batches, err := DecodeBatches(config, bytes.NewReader(tx.Data()))
+			if err != nil {
+				// TODO: log error
```
norswap, line 199: Let's do this before merging?

norswap (commented):
LGTM

opnode/rollup/derive/payload_attributes.go
mslipper: I say we return the error here and handle logging later. This is library code anyway, which IMO shouldn't log.

trianglesphere:
FYI: There is a problem in the full system with batch decoding. Not sure where it is, but looking into it rn.

opnode/rollup/derive/batch.go
```diff
@@ -0,0 +1,153 @@
...
+		return fmt.Errorf("unrecognized batch bundle type: %d", bundleType)
+	}
+}
+
+// EncodeRLP implements rlp.Encoder
+func (b *BatchData) EncodeRLP(w io.Writer) error {
+	buf := encodeBufferPool.Get().(*bytes.Buffer)
+	defer encodeBufferPool.Put(buf)
+	buf.Reset()
+	if err := b.encodeTyped(buf); err != nil {
```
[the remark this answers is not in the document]
protolambda: Disagree, we should, see:
- https://github.com/ethereum/go-ethereum/blob/830231c1c40f45f32cbf4ab69954b87a79313f44/core/types/transaction.go#L143
- https://github.com/ethereum/go-ethereum/blob/830231c1c40f45f32cbf4ab69954b87a79313f44/core/types/transaction.go#L99

The typed functions then decode/encode with the prefix byte.

The method is called `EncodeRLP` not because of the contents of the batch, but because of the place it is embedded into (a bundle is a RLP list of batches)

opnode/rollup/derive/batch.go
```diff
@@ -0,0 +1,153 @@
...
+		return out, nil
+	case BatchBundleV2Type:
+		// TODO: implement compression of a bundle of batches
+		return nil, errors.New("bundle v2 not supported yet")
+	default:
+		return nil, fmt.Errorf("unrecognized batch bundle type: %d", typeData[0])
+	}
+}
+
+func EncodeBatches(config *rollup.Config, batches []*BatchData, w io.Writer) error {
```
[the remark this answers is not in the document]
protolambda: Based on the configuration and batch input we may change the compression or encoding method, e.g. we compare some configured fork-number, or we simply check if the config enables compression at all. I'm adding the argument so we don't have to refactor the calls later

opnode/rollup/derive/batch.go
```diff
@@ -0,0 +1,153 @@
...
+//
+// payload := RLP([batch_0, batch_1, ..., batch_N])
+// bundleV1 := BatchBundleV1Type ++ payload
+// bundleV2 := BatchBundleV2Type ++ compress(payload)  # TODO: compressed bundle of batches
+//
+// An empty input is not a valid bundle.
+//
+// Note: the type system is based on L1 typed transactions.
+
+// encodeBufferPool holds temporary encoder buffers batch encoding
```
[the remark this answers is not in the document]
protolambda: > encodeBufferPool holds temporary encoder buffers *for* batch encoding

My bad, missed a word

opnode/rollup/derive/batch.go
protolambda: A 0-length byte string, e.g. a transaction without any input data

opnode/rollup/derive/batch.go
```diff
@@ -0,0 +1,153 @@
...
+}
+
+type BatchData struct {
+	BatchV1
+	// batches may contain additional data with new upgrades
+}
+
+func DecodeBatches(config *rollup.Config, r io.Reader) ([]*BatchData, error) {
+	var typeData [1]byte
+	if _, err := r.Read(typeData[:]); err != nil {
```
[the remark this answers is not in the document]
protolambda: Changed it to use `io.ReadFull` to ensure the error check is sufficient

protolambda (approved):
LGTM. Spec PR is updated to match also.

PR #248, opened 2022-03-15
ref impl: Simple batch submitter

trianglesphere:
**Description**
Note: very simple.

### DIFF
--- a/opnode/bss/submitter.go
+++ b/opnode/bss/submitter.go
@@ -1,0 +1,103 @@
+package bss
+
+import (
+	"context"
+	"crypto/ecdsa"
+	"errors"
+	"math/big"
+	"time"
+
+	"github.com/ethereum-optimism/optimistic-specs/opnode/rollup/derive"
+	"github.com/ethereum/go-ethereum"
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/core/types"
+	"github.com/ethereum/go-ethereum/crypto"
+	"github.com/ethereum/go-ethereum/ethclient"
+)
+
+type BatchSubmitter struct {
+	Client    *ethclient.Client
+	ToAddress common.Address
+	ChainID   *big.Int
+	PrivKey   *ecdsa.PrivateKey
+}
+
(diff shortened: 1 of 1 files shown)


protolambda:
Rebased on new `jg/batch_struct` to support bundles of batches, and small change to take a rollup-config and bundle as argument. Small step away from compressing the bundle of L2 block inputs :)

norswap:
Trying to think about what needs to change here to accomodate what we said in today's office hours...

- The big problem with this as-is is if we the sequencer sends us a batch, and creates the corresponding block on the execution engine, if we crash before submitting the batch, then the batch will never be submitted. This will eventually cause a L2 re-org, but it's not great.
- The ideal implementation is to have a loop that is wired to the current view of the "unsafe L2 head" (on the execution engine) and the "safe L2 head" (everything that can be derived from L1) and turns unsafe blocks into batches (it will also need to track which unsafe blocks have been submitted but haven't made it to the safe head yet).
     - We *might* want to track in-flight transactions somehow? Might not be necessary.
- In the future, we can consider a "remote signer" model where we send batches to be sent and submitted to a remote server that can only talk to the sequencer.

Hardening concerns:
- This never retries batch sends. Probably should have some shared logic with the output submitter to handle this.
    - Should we, in the future, enable the sequencer to cancel a batch? This could make sense if we detect a reorg before a sequencer is submitted, as remaking the batch would avoid bad situation like transactions with not enough gas. Not a priority **at all**.
- This currently does allow multiple transactions in flight at the same time. This is something we want long term, but probably not as naively as this. (Related to how we handle submission re-tries.)

mslipper:
I think this is fine for now. We have library code for the current batch submitter that will retry sends, and manage gas price increments. Same for concurrent batch submissions; we're working on that in the current one and can port it here.

I like Norswap's submission loop idea since it separates the logic of "what goes into the batch" from the logic of "here's how to generate and submit the batch."

mslipper (approved):

trianglesphere:
reopened in https://github.com/ethereum-optimism/optimistic-specs/pull/254

protolambda:
Oh I see, you merged into the old branch instead of main. Will merge new PR :+1:

PR #249, opened 2022-03-15
ref impl: Sequencing

trianglesphere:
**Description**
Adds a sequencer mode that runs ahead of the verifier.
Mostly working, but some things disabled.

Still a WIP, some of the stuff around timestamps / batches is super shaky.

**Additional context**

**Metadata**
- Fixes #[Link to Issue]


protolambda:
update:
- rebased to handle the batch encoding/decoding changes, and ability to submit multiple batches
- updated payload attributes usage to be explicit about yes/no tx-pool usage of geth during block building (verifier always provides pure list of all transactions, sequencer submits deposits (if any) but also takes from pool)

opnode/rollup/derive/payload_attributes.go
```diff
@@ -192,17 +193,18 @@ func BatchesFromEVMTransactions(config *rollup.Config, txs []*types.Transaction)
...
 			if seqDataSubmitter != config.BatchSenderAddress {
+				// TODO: log/record metric
 				continue // not an authorized batch submitter, ignore
 			}
 			batches, err := DecodeBatches(config, bytes.NewReader(tx.Data()))
 			if err != nil {
-				// TODO: log error
-				continue
+				// TODO: Maybe log, otherwise this is unrecoverable
+				return nil, err
```
protolambda, line 202: This should be an ignored error, that's very important: a single bad tx shouldn't cause us to completely fail to parse the L1 data. If something is invalid, we should make the verifier skip it, and thus the L2 can continue making progress.

opnode/rollup/driver/state.go
```diff
@@ -130,6 +141,38 @@ func (s *state) loop() {
 		// case <-l2Poll.C:
 		case <-s.done:
 			return
+		case <-l2BlockCreation.C:
+			// 0. Make sure to only do this as a sequencer
+			if !s.sequencer {
+				continue
+			}
```
protolambda, line 148: To optionally ignore channel events there's a better trick: set the channel to `nil` if it should not be listened from. The select statement will not read from the `nil` channel.

opnode/rollup/driver/step.go
```diff
@@ -36,6 +38,54 @@ type outputImpl struct {
 	Config rollup.Config
 }
 
+func (d *outputImpl) newBlock(ctx context.Context, l2Finalized eth.BlockID, l2Parent eth.BlockID, l1Origin eth.BlockID, includeDeposits bool) (eth.BlockID, *derive.BatchData, error) {
+	d.log.Info("creating new block", "l2Parent", l2Parent, "l1Origin", l1Origin, "includeDeposits", includeDeposits)
+	fetchCtx, cancel := context.WithTimeout(ctx, time.Second*20)
+	defer cancel()
+	l2Info, err := d.rpc.BlockByHash(fetchCtx, l2Parent.Hash)
```
protolambda, line 45: We can check if the previous block has the same L1 information to determine if we've included deposits yet, and then we don't need the `includeDeposits` argument

opnode/rollup/driver/step.go
```diff
@@ -120,3 +176,40 @@ func AddBlock(ctx context.Context, logger log.Logger, rpc DriverAPI,
 	logger.Info("updated fork-choice with block")
 	return payload.ID(), nil
 }
+
+// AddBlockForBSS extends the L2 chain by deriving the full execution payload from inputs,
+// and then executing and persisting it.
+//
+// After the step completes it returns the block ID of the last processed L2 block, even if an error occurs.
+func AddBlockForBSS(ctx context.Context, logger log.Logger, rpc DriverAPI, epoch uint64,
```
protolambda, line 184: The cleanup approach (other PR) was to have one `AddBlock` that returns the full payload, and then either compute the block ID (as verifier) or the batch (as sequencer) from that right? Or are there other differences?

protolambda (changes requested):
We should ignore invalid batches, not completely stop the derivation process. We should also add a test where we encounter such invalid data

opnode/rollup/driver/step.go
trianglesphere: There are other differences, but I can bring back the single add block function. If all you get is the block ID it's impossible to create the batch, but if you have the payload, it's not that hard.

opnode/rollup/derive/payload_attributes.go
trianglesphere: This is true, but an error in the validation code here will cause a chain split if it's not the same across different rollup node versions.

tynes: The attack is the sequencer purposely includes a faulty transaction in the batch to halt all of the verifiers. I think adding a comment here and having a `MUST` in the spec is a good idea to ensure that this is taken into account

opnode/rollup/driver/state.go
```diff
@@ -130,6 +141,38 @@ func (s *state) loop() {
...
+			// 2. Ask output to create new block
+			newUnsafeL2Head, batch, err := s.output.newBlock(context.Background(), s.l2Finalized, s.l2UnsafeHead, s.l1Origin, firstOfEpoch)
+			if err != nil {
+				s.log.Error("Could not extend chain as sequencer", "err", err, "l2UnsafeHead", s.l2UnsafeHead, "l1Origin", s.l1Origin)
+				continue
+			}
+			// 3. Update unsafe l2 head + epoch
+			s.l2UnsafeHead = newUnsafeL2Head
```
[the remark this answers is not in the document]
trianglesphere: No, the state object is only accesses in a single threaded manner (primarily by the loop, but also initialized in the start function).

opnode/rollup/driver/step.go
trianglesphere: Doesn't work against the origin block

opnode/rollup/derive/payload_attributes.go
trianglesphere: done

opnode/rollup/driver/step.go
protolambda: We can make the l1-info parsing function just return an artificial l1 info, because the genesis block cannot have transactions. Not a big deal for this PR though, I'll open an issue to try reduce the function arguments here.

PR #253, opened 2022-03-16
Add Withdrawals specification

maurelian:
I'm opening this PR as a draft most to give others some visibility into the status. 

**Todo:** 

- [x] summarize tradeoffs with having the withdrawal verification code in the deposit feed
- [ ] Add pseudo code for verification path (won't fix)
- [ ] expand on security considerations 

This necessarily builds on #238


specs/withdrawals.md
```diff
@@ -0,0 +1,164 @@
...
+| `WithdrawalsRootInclusionProof`   | `storageRootProof` |
+| `WithdrawalMessageInclusionProof` | `messageProof`     |
+
+These inputs must satisfy the following conditions:
+
+1. The `timestamp` is at least `FINALIZATION_WINDOW` seconds old.
+1. `OutputOracle.l2Outputs(timestamp)` returns a non-zero value `l2Output`.
+1. The `storageRootProof` is valid for the `l2Output` returned in step 2, according to the SSZ encoding described in the
+   [L2 output commitment construction](./proposals.md#l2-output-commitment-construction).
+1. The `messageProof` is valid for the provided `message` and `messageProof`
```
maurelian, line 95: Appreciate review here to ensure no checks are missing.
ASK that pseudocode would be helpful, I will add it time allowing.

specs/withdrawals.md
```diff
@@ -0,0 +1,176 @@
...
+## Withdrawal Flow
+
+**On L2:**
+
+1. An L2 account sends a withdrawal message (and possibly also ETH) to the `Withdrawoooor` predeploy contract.
+   This is a very simple contract that stores a mapping from the hash of the `WithdrawalMessage` as defined above to a
+   boolean value. (`mapping (bytes32 => bool) withdrawalMessages`)
+2. If ETH is being withdrawn, it can eventually be burned by deploying a contract which immediately `SELFDESTRUCT`s.
+   This has the benefit of using an existing EVM mechanism for removing the ETH from the world state, without having to
+   add a diff to L2 Geth.
```
norswap, line 107: ACD has had `SELFDESTRUCT` in its crosshair for a while now, there's been talk of at least refusing new contracts that use it, etc. So this might introduce more complexity down the line.

That being said, maybe we could implement like this now, as long as we have a possible upgrade path (which would simply be the execution engine fishing out these transactions and burning the ETH).

specs/withdrawals.md
```diff
@@ -0,0 +1,176 @@
...
+
+**On L1:**
+
+1. A [relayer][g-relayer] submits the required inputs to the `DepositFeed` contract. The relayer may or may not be the same entity which
+   initiated the withdrawal on L2.
```
norswap, line 112: Wait, why are we tangling withdrawals with deposits??

specs/withdrawals.md
```diff
@@ -0,0 +1,176 @@
...
+2. The `DepositFeed` contract retrieves the output root from the `OutputOracle`'s `l2Outputs()` function, and performs
+   the remainder of the verification process internally.
```
norswap, line 114: Would put a forward reference here, otherwise you're wondering what the rest of the process is and if it's already been explained. (But it's not yet since this is still a draft.)

specs/withdrawals.md
norswap: Do we in general expect users to send both transactions, being both withdrawer and relayers? If not, then a mechanism for including a relayer payment might be in scope here (probably for later though, after the sequencing milestone).

norswap: This section should state explicitly that relaying must occur at least 7 days (`FINALIZATION_WINDOW` seconds) after the output commitment for the withdrawal was submitted to L1.

specs/withdrawals.md
```diff
@@ -0,0 +1,176 @@
...
+
+## Security Considerations
+
+### Key Properties of Withdrawal Verification
+
+1. It should not be possible 'double spend' a withdrawal, ie. to relay a `WithdrawalMessage` on L1 which does not correspond to a message
+  initiated on L2. For reference, see [this writeup][polygon-dbl-spend] of a vulnerability of this type found on Polygon.
+
+    [polygon-dbl-spend]: https://gerhard-wagner.medium.com/double-spending-bug-in-polygons-plasma-bridge-2e0954ccadf1
+
```
norswap, line 158: Addittionally (can be a separate bullet), it should not be possible to relay the same message multiple times.

norswap (commented):
Still a draft, so no definitive conclusions, but right now this is very much gives rules & constraints but only with minimal explanations / rationale for the process.

I'm still really hating the front-loaded definitions, trying to understand things I really felt like I had to piece things together and jump around instead of being gently guided through the process with things introduced when they become useful.

I do think everything that's written is pretty damn good though, mostly would add things / reorder.

(Again, this is still a draft, so I understand this is not the expected end point.)

specs/glossary.md
```diff
@@ -257,6 +260,26 @@ cf. [Deposits Specification](deposits.md)
...
+
+[withdrawals]: glossary.md#withdrawals
+
+In general, a withdrawal is a transaction sent from L2 to L1 that may transfer of data and/or value.
+
+This term *withdrawal* is somewhat ambiguous as these "transactions" exist at multiple levels. In order to differentiate
+ between the L1 and L2 components of a withdrawal we introduce the following terms:
+
+- A *withdrawing* call or transaction occurs on L2 and initiates a withdrawal.
+- A *withdrawn* call or transaction occurs on l1 as the result of an withdrawal being relayed on L1.
```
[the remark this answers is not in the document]
karlfloersch: Agree -- these terms are a bit too similar. Another option would be: "withdrawal initiation" and "withdrawal finalization" - initiation on L2 and finalization on L1. Got the term "finalization" from [our current docs](https://community.optimism.io/docs/developers/bridge/messaging/#for-l1-%E2%87%92-l2-transactions). An alternative to "finalization" could be "execution".

specs/withdrawals.md
```diff
@@ -0,0 +1,176 @@
...
+| --------------------------------- | ------------------ |
+| `WithdrawalMessage`               | `message`          |
+| `L2OutputTimestamp`               | `timestamp`        |
+| `WithdrawalsRootInclusionProof`   | `storageRootProof` |
+| `WithdrawalMessageInclusionProof` | `messageProof`     |
+
+These inputs must satisfy the following conditions:
+
+1. The `timestamp` is at least `FINALIZATION_WINDOW` seconds old.
+1. `OutputOracle.l2Outputs(timestamp)` returns a non-zero value `l2Output`.
```
karlfloersch, line 93: `OutputOracle.l2Outputs(timestamp)` probably needs to be updated to reflect the introduction of a timestamp which is recorded at time of submission

specs/withdrawals.md
karlfloersch: Other than the timestamp stuff I think this is right.

specs/withdrawals.md
karlfloersch: Ugh using SELFDESTRUCT upsets me. It is probably preferable to adding an EVM diff though... We should keep our eyes out for a better solution but this is a fine default (until we remove SELFDESTRUCT from the EVM)

specs/withdrawals.md
karlfloersch: I think `DepositFeed` here should actually be the [`L1CrossDomainMessenger`](https://github.com/ethereum-optimism/optimism/blob/develop/packages/contracts/contracts/L1/messaging/L1CrossDomainMessenger.sol). And btw it's important we maintain backwards compatibility with the interface defined there as much as possible

specs/withdrawals.md
```diff
@@ -0,0 +1,176 @@
...
+
+## Data Structures and Type Aliases
+
+1. A `WithdrawalMessage` is encoded in a struct as follows:
+
+    ```js
+    struct WithdrawalMessage {
+        uint256 nonce;
+        address sender;
+        uint256 value;
```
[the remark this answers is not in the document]
maurelian: Yes, I've though a lot about this and am pretty convinced we need to do this.

specs/withdrawals.md
```diff
@@ -0,0 +1,176 @@
...
+1. `OutputOracle.l2Outputs(timestamp)` returns a non-zero value `l2Output`.
+1. The `storageRootProof` is valid for the `l2Output` returned in step 2, according to the SSZ encoding described in the
+   [L2 output commitment construction](./proposals.md#l2-output-commitment-construction).
+1. The `messageProof` is valid for the provided `message` and `messageProof`
+
+## Withdrawal Flow
+
+**On L2:**
+
+1. An L2 account sends a withdrawal message (and possibly also ETH) to the `Withdrawoooor` predeploy contract.
```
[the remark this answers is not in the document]
maurelian: Sorry, I just don't know what to call it. Suggestions?

specs/glossary.md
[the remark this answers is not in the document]
norswap: I'd love this to be more general -- these are simply messages from L2 to L1 whose main use case happen to be withdrawals. I like the relayed terminology for that reason.

specs/withdrawals.md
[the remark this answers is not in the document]
norswap: Haha no need to be sorry! Could have been just `Withdrawals` but I'd like it to be more general, so `Relay`?

specs/proposals.md
```diff
@@ -71,3 +84,15 @@ This is a work-in-progress, see [issue 181](https://github.com/ethereum-optimism
...
+
+> **TODO** spec for the [L2OutputOracle contract]
+[L2OutputOracle contract]: <https://github.com/ethereum-optimism/optimistic-specs/blob/8c1d509e3882ca49f2415c2b724c8a58a134cec7/packages/contracts/contracts/L1/MockL2OutputOracle.sol#L1>
+
+## Security Considerations
+
+> **TODO**
+>
+> - list invariants
+> - list things that could go wrong and ways to avoid that
```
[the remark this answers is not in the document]
maurelian: yes, it builds on it unfortunately.

specs/withdrawals.md
maurelian: The reason I've suggested handling withdrawals via the `DepositFeed` is because that's where the ETH is, so we need to go through there to handle withdrawals with value.

The alternative would be to deposit ETH into the `L1CrossDomainMessenger`, but then you need to add an auth check in the `DepositFeed` that gives the `L1CrossDomainMessenger` the sole permission to mint ETH on L2. That also forces all ETH deposit transactions to have the L1 messenger as the caller. 

My [preference](https://github.com/ethereum-optimism/optimistic-specs/discussions/242#discussioncomment-2409083) would be to build in separate backwards compatibility affordances for allowing the messenger and bridge contracts to continue functioning.

packages/contracts/contracts/L1/MockL2OutputOracle.sol
```diff
@@ -10,7 +10,7 @@ contract MockL2OutputOracle {
     mapping(uint256 => bytes32) public l2Outputs;
     uint256 public historicalTotalBlocks;
     uint256 public latestBlockTimestamp;
```
tynes, line 12: Can't these all be immutables?

specs/glossary.md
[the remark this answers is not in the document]
tynes: I see @norswap's point of making it general, I think its elegant from the point of view of the protocol but also would be confusing for end users who don't understand the implementation. I think we should use the more general terms, call out here that withdrawals are expected to be a main usecase and then ensure that they are called withdrawals in user facing UIs

specs/withdrawals.md
```diff
@@ -0,0 +1,176 @@
...
+- _withdrawn transaction_ refers specifically to an L1 transaction
+- _withdrawal_ can refer to the transaction at various stages
+
+## Constants
+
+| Name                  | Value     | Unit    |
+| --------------------- | --------- | ------- |
+| `FINALIZATION_WINDOW` | `604_800` | seconds |
+
+This `FINALIZATION_WINDOW` value is equivalent to 7 days.
```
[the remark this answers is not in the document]
tynes: Perhaps this is out of scope, but with the merge on the horizon, has there been any consideration of lowering the `FINALIZATION_WINDOW`? Under PoS, it should be easier to coordinate around censorship attacks, even tho I don't particularly love imposing irregular state transitions as part of the security model

specs/withdrawals.md
```diff
@@ -0,0 +1,176 @@
...
+  [L2 output commitment construction](./proposals.md#l2-output-commitment-construction).
+
+1. A `WithdrawalMessageInclusionProof` is an MPT proof encoded in a struct as follows:
+
+   ```js
+   struct WithdrawalMessageInclusionProof {
+       WithdrawalMessage message;
+       bytes32 l2withdrawalsRoot;
+       bytes memory _key;      // storage key of the withdrawal message commitment
+       bytes memory _value;    // Always bytes32(1) (boolean validity status of the message)
```
tynes, line 72: Why does this have to be included in the struct if it is a constant?

maurelian: Good point, it does not.

specs/glossary.md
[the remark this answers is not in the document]
norswap: Imho the end user won't be confronted to the spec and not even to contract names. Any appearance of those would be in an explanation piece where explaining that withdrawals are implemented by sending a message from L2 to L1 actually helps the explanation.

Of course, I'd expect UI for token withdrawal to call them "withdrawal".

specs/withdrawals.md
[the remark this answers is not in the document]
norswap: @tynes Was not aware of that, what changes that helps with censorship?

specs/withdrawals.md
```diff
@@ -0,0 +1,176 @@
...
+  [L2 output commitment construction](./proposals.md#l2-output-commitment-construction).
+
+## Withdrawal verification
+
+In order to verify a `WithdrawalMessage`, the following inputs must be provided:
+
+| Type                              | Name               |
+| --------------------------------- | ------------------ |
+| `WithdrawalMessage`               | `message`          |
+| `L2OutputTimestamp`               | `timestamp`        |
```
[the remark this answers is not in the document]
maurelian: That makes sense, but just so I'm clear, what is the intent with the timestamp used as the key in this mapping? 

https://gist.github.com/karlfloersch/02ce7938000680ce7820f39ba06bd9b1#file-staterootoracle-sol-L45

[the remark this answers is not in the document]
karlfloersch: Ahhhh! My goodness I'm sorry @maurelian -- I forgot that I implemented this to be super simple instead of building in features that will eventually be needed in the final version. For now we should just use the `timestamp` inside the output because it's easier, but in the long term we will definitely need to move to the model that I outlined in my comment. The reason why it's fine to use the output timestamp for now is because it doesn't effect the (already quite strong) trust assumptions around the sequencer who's submitting outputs.

specs/glossary.md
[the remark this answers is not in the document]
maurelian: Doing this for now: 
```

**Vocabulary note**: *withdrawal* can refer to the transaction at various stages of the process, but we introduce
more specific terms to differentiate between the transaction=
- *withdrawal initiating transaction* refers specifically to a transaction on L2 sent to the Withdrawals predeploy.
- *withdrawal finalizing transaction* refers specifically to an L1 transaction which finalizes and relays the withdrawal.
```

specs/withdrawals.md
```diff
@@ -0,0 +1,176 @@
...
+        uint256 nonce;
+        address sender;
+        uint256 value;
+        bytes message;
+    }
+    ```
+
+1. The `L2OutputTimestamp` is an alias for `uint256`, and MUST be a multiple of the `SUBMISSION_INTERVAL` described
+  in the [L2 Output](./proposals.md#constants) document.
+
```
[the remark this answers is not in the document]
maurelian: Resolved by moving this content to the end of the doc per your request

maurelian:
replaced by #310

PR #254, opened 2022-03-16
ref impl: Simple batch submitter

trianglesphere:
**Description**
Barebones batch submitter.


protolambda (approved):
This was already reviewed / merged in PR #248 but directed to the wrong branch. Merging again now.
This is a good base to have basic batch-submission, but I'll continue with it to handle multiple batches and make it more robust (i.e. can't rely on one-time push with restarts etc.)

PR #256, opened 2022-03-17
ref impl: Clean up sequencer

trianglesphere:
**Description**
This better handles the fork choice update in the rollup node, exposes the fork choice method throughout,
cleans up the L2 Engine API interface, and random aspects of block creation. 

The `l2.Source` is the Ethereum client wrapper for the L2 execution engine. It directly wraps some block accessing methods (like `BlockByNumber` and `BlockByHash`). It also wraps raw RPC calls to create a golang interface for the Engine API (`newBlock`, `forkchoiceUpdated`, `getPayload`).

This PR also pulls out the ethclient initialization loop into a standalone function.


norswap:
Note: https://github.com/ethereum-optimism/optimistic-specs/pull/260 is squashed as the first commit of this PR.

norswap:
As for the new organization of interfaces, I think the litmus test would be a good written explanation of what all of these interfaces are, relate to one another, are created/held/passed.

trianglesphere:
> What's the relationship between #260 and this?

#260 is a follow on targeting more of the L1 sources + sync start algorithm

trianglesphere:
> As for the new organization of interfaces, I think the litmus test would be a good written explanation of what all of these interfaces are, relate to one another, are created/held/passed.

The `l2.Source` is the Ethereum client wrapper for the L2 execution engine. It directly wraps some block accessing methods (like `BlockByNumber` and `BlockByHash`). It also wraps raw RPC calls to create a golang interface for the Engine API (`newBlock`, `forkchoiceUpdated`, `getPayload`).


For the follow on:
This flattens the dependency hierarchy. Now there are two `sources` - one for L1 and one for L2 - that implement all access to L1/L2 data/instructions. It implements methods like the `BlockReference` directly in the L1/L2 source. The chainsource concept has been removed.

At sites that depend on L1/L2 data, they define an interface with the subset of method that they need, but the `l1.Source` and the `l2.Source` fulfill every required method (except for `FindSyncStart` but that requires both L1 and L2 data).

opnode/rollup/derive/payload_attributes.go
```diff
@@ -236,91 +224,68 @@ func FilterBatches(config *rollup.Config, epoch rollup.Epoch, minL2Time uint64,
...
-			Timestamp:             l2.Uint64Quantity(batch.Timestamp),
-			Random:                randomnessSeed,
-			SuggestedFeeRecipient: config.FeeRecipientAddress,
-			Transactions:          txns,
-		}
+// SortedAndPreparedBatches turns a collection of batches to the input batches for a series of blocks
+func SortedAndPreparedBatches(batches []*BatchData, epoch, blockTime, minL2Time, maxL2Time uint64) []*BatchData {
+	m := make(map[uint64]*BatchData)
+	for _, b := range batches {
+		b := b // not sure if copy is required, but go does not make this easy
```
protolambda, line 254: Should not be required. We're only dealing with a pointer here, we already copy the pointer when we set `m[b.BatchV1.Timestamp]` below.

opnode/rollup/driver/step.go
```diff
@@ -137,52 +159,73 @@ func (d *outputImpl) step(ctx context.Context, l2Head eth.BlockID, l2Finalized e
...
+	if updateSafe {
+		fc.SafeBlockHash = payload.BlockHash
+	}
+	if updateUnsafe {
+		fc.HeadBlockHash = payload.BlockHash
+	}
+	_, err = d.l2.ForkchoiceUpdate(ctx, &fc, nil)
 	if err != nil {
-		return nil, fmt.Errorf("failed to persist execution payload: %v", err)
+		return nil, fmt.Errorf("Could not make the new L2 block canonical via forkchoice: %w", err)
```
protolambda, line 228: nit: don't start error messages with a capital, Go style rules: https://github.com/golang/go/wiki/CodeReviewComments#error-strings  (I know, probably the most subjective thing in the whole guide, but my IDE will complain about it every time I open the file)

protolambda:
Two minor things, generally looks good, and #260 addresses the API design, so I'm going to ignore that part here. Also one small merge conflict. Will review/approve when PR is updated

opnode/rollup/driver/step.go
trianglesphere: I didn't realize this. Will fix. Maybe we should add a lint for this

opnode/rollup/derive/payload_attributes.go
trianglesphere: Right yes. Double checked and the problem is only when taking the address of the loop variable.

protolambda (approved):
LGTM

PR #260, opened 2022-03-17
ref impl: Flatten sources & Cleanup Sync Start

trianglesphere:
**Description**
This makes the following changes:
1. Flatten the L1 & L2 source into one object that fulfills several interfaces (primarily the `BlockReference`)
2. Place the `L1Range` function in the L1 Source (removed from `sync` package)
3. Provide `FindSyncStart` with the L2 block to use


opnode/rollup/derive/payload_attributes.go
```diff
@@ -271,7 +271,7 @@ func SortedAndPreparedBatches(batches []*BatchData, epoch, blockTime, minL2Time,
 	return out
 }
 
-func L1InfoDepositBytes(l1Info L1Info) (l2.Data, error) {
+func L1InfoDepositBytes(l1Info L1Info) (hexutil.Bytes, error) {
```
trianglesphere, line 274: Changed to avoid import cycle (`l2` depends on `derive` to implement `BlockReference` API, but `derive` depended on `l2` for types definition)

protolambda: That's fine, wasn't supposed to be `l2.Data` in the first place. `l2.Data` is only there to describe the `DATA` type in JSON RPC, used in the transactions list for each element.

opnode/l1/source.go
```diff
@@ -13,6 +14,11 @@ import (
 	"github.com/ethereum/go-ethereum/ethclient"
 )
 
+var WrongChainErr = errors.New("wrong chain")
+var TooDeepReorgErr = errors.New("reorg is too deep")
+var MaxReorgDepth = 500
+var MaxBlocksInL1Range = uint64(100)
```
protolambda, line 20: These can be constants

opnode/rollup/driver/driver.go
```diff
@@ -21,23 +22,52 @@ type BatchSubmitter interface {
...
+type Downloader interface {
+	// FetchL1Info fetches the L1 header information corresponding to a L1 block ID
+	FetchL1Info(ctx context.Context, id eth.BlockID) (derive.L1Info, error)
+	// FetchReceipts of a L1 block
+	FetchReceipts(ctx context.Context, id eth.BlockID) ([]*types.Receipt, error)
+	// FetchTransactions from the given window of L1 blocks
+	FetchTransactions(ctx context.Context, window []eth.BlockID) ([]*types.Transaction, error)
+}
+
+type BlockPreparer interface {
```
protolambda, line 34: Can't we just call it the `ExecutionEngine`? Or maybe just `Engine`?

opnode/rollup/driver/step.go
```diff
@@ -194,7 +176,7 @@ func (d *outputImpl) step(ctx context.Context, l2Head eth.BlockID, l2Finalized e
 			return last, fmt.Errorf("failed to extend L2 chain at block %d/%d of epoch %d: %w", i, len(batches), epoch, err)
 		}
 		last = payload.ID()
-		fc.HeadBlockHash = last.Hash
+		fc.HeadBlockHash = last.Hash // should be safe block, but geth is broken
```
protolambda, line 179: If this is the new tip of the chain, and it is derived from L1, then it should be setting **both** `fc.HeadBlockHash` and `fc.SafeBlockHash`. Maybe that works as expected with geth?

protolambda (commented):
Looks reasonable, maybe try to set both `fc.SafeBlockHash` and `fc.HeadBlockHash`. If that doesn't work we'll resolve it later with the remaining unsafe-chain extension work.
And some other nits.

opnode/l1/source.go
trianglesphere: Yep, plus I removed some that were duplicated

opnode/rollup/driver/driver.go
trianglesphere: Renamed to engine

opnode/rollup/driver/step.go
trianglesphere: Yes, it should be setting both (when working as a verifier). The thing that geth needed was setting head block.

protolambda (approved):
LGTM

PR #263, opened 2022-03-17
rollup node config

### LINKED ISSUE #222: Opnode chain config loading and improved defaults
There are two types of configuration:
- Free configuration that the user can change at any time
- Chain configuration, affecting the L2 derivation: all nodes should configure the same, but e.g. a testnet may want it different than mainnet

I think we should keep the first to just CLI options, as there are not many things to configure here (only RPC endpoints basically).

Currently the chain configuration is represented as a `rollup.Config` struct (introduced in #219), which we can improve in several ways:
- load from a file
- provide some sane defaults (eventually default to a `mainnet` set of values)
- support labeled pre-configurations (e.g. `--network=bedrocktestnet` configures the node for a testnet)
- include the genesis block-hashes/numbers that we previously configured via CLI in this config
- override config values via CLI

This is very similar to the config system of go-ethereum and other blockchain nodes, we may be able to use the same patterns/libraries.

protolambda:
- Split config flags into flags package, like the output submitter does
- Define `--sequencing.enabled` flag to enable sequencer
- Define `--batchsubmitter.key` key flag to load key for batch submitter, from a file with hex-encoded priv key.
- De-duplicate genesis config values
- Properly set L2 genesis time in e2e test
- Load rollup-config (chain params) from json file, a required flag `rollup.config`
- Implement `Check()` for rollup-config, prevent common configuration errors
- Update config / block id / block reference types for JSON encoding
- Move flag parsing and config creation to `opnode` package root (mirrors how output submitter does it), to make it reusable
- Define environment vars for rollup node flags

No rollup-config defaults yet, there's no public testnet to point it to.

Fix #222

### DIFF
--- a/opnode/flags/flags.go
+++ b/opnode/flags/flags.go
@@ -1,0 +1,84 @@
+package flags
+
+import "github.com/urfave/cli"
+
+// Flags
+
+const envVarPrefix = "ROLLUP_NODE_"
+
+func prefixEnvVar(name string) string {
+	return envVarPrefix + name
+}
+
+var (
+	/* Required Flags */
+	L1NodeAddr = cli.StringFlag{
+		Name:     "l1",
+		Usage:    "Address of L1 User JSON-RPC endpoint to use (eth namespace required)",
+		Required: true,
+		Value:    "http://127.0.0.1:8545",
+		EnvVar:   prefixEnvVar("L1_ETH_RPC"),
+	}
+	L2EngineAddrs = cli.StringSliceFlag{
+		Name:     "l2",
+		Usage:    "Addresses of L2 Engine JSON-RPC endpoints to use (engine and eth namespace required)",
+		Required: true,
+		EnvVar:   prefixEnvVar("L2_ENGINE_RPC"),
+	}
+	RollupConfig = cli.StringFlag{
+		Name:     "rollup.config",
+		Usage:    "Rollup chain parameters",
+		Required: true,
+		EnvVar:   prefixEnvVar("ROLLUP_CONFIG"),
+	}
+
+	/* Optional Flags */
+
+	SequencingEnabledFlag = cli.BoolFlag{
+		Name:   "sequencing.enabled",
+		Usage:  "enable sequencing",
+		EnvVar: prefixEnvVar("SEQUENCING_ENABLED"),
+	}
+
+	// TODO: move batch submitter to stand-alone process
+	BatchSubmitterKeyFlag = cli.BoolFlag{
+		Name:   "batchsubmitter.key",
+		Usage:  "key for batch submitting",
+		EnvVar: prefixEnvVar("BATCHSUBMITTER_KEY"),
+	}
+
+	LogLevelFlag = cli.StringFlag{
+		Name:   "log.level",
+		Usage:  "The lowest log level that will be output",
+		Value:  "info",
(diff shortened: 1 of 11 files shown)


mslipper (approved):

opnode/rollup/types.go
```diff
@@ -10,35 +11,58 @@ import (
...
+	BatchSenderAddress common.Address `json:"batch_sender_address"`
+}
+
+// Check verifies that the given configuration makes sense
+func (cfg *Config) Check() error {
+	if cfg.BlockTime == 0 {
+		return errors.New("block time cannot be 0")
+	}
+	if cfg.SeqWindowSize == 0 {
+		return errors.New("sequencing window size cannot be 0")
```
trianglesphere, line 54: Should be at least 2 (otherwise cannot submit batches in time).

opnode/flags/flags.go
```diff
@@ -0,0 +1,84 @@
...
+}
+
+var (
+	/* Required Flags */
+	L1NodeAddr = cli.StringFlag{
+		Name:     "l1",
+		Usage:    "Address of L1 User JSON-RPC endpoint to use (eth namespace required)",
+		Required: true,
+		Value:    "http://127.0.0.1:8545",
+		EnvVar:   prefixEnvVar("L1_ETH_RPC"),
```
trianglesphere, line 20: Nice, we definitely needed these.

mslipper:
This is great, thanks Proto!

One quick note re: the batch submitter's private key. Using env vars for the mnemonic/private key is really useful since it lets us define it in a Kubernetes secret. While we can put the keystore itself into a secret and mount it as a volume, we'd still have to specify the keystore password in to unlock it which defeats the purpose of having the keystore in the first place.

If we want to be super secure about this we should define a signer interface with a couple implementations:

- Environment
- Keystore

Then, users can choose which implementation they want to use. We could also extend it further, and separate our internal signers into standalone secure services.

protolambda:
@mslipper agreed, for testing I don't think it matters much though, so I'm going with a minimal option of just loading the key from a file. If you want we can add a env-var option for the key itself instead of the key path, but at that point I'd just open a PR to do the proper thing with keystores as well.

mslipper:
Yep, sounds fine re: the file - we can mount the secret as a file pretty easily.

opnode/rollup/types.go
```diff
@@ -10,35 +12,58 @@ import (
...
-	BatchSenderAddress common.Address
+	BatchSenderAddress common.Address `json:"batch_sender_address"`
+}
+
+// Check verifies that the given configuration makes sense
+func (cfg *Config) Check() error {
+	if cfg.BlockTime == 0 {
+		return fmt.Errorf("block time cannot be 0, got %d", cfg.BlockTime)
+	}
+	if cfg.SeqWindowSize >= 2 {
```
trianglesphere, line 54:
```suggestion
	if cfg.SeqWindowSize < 2 {
```

protolambda: ayy, my bad, flipped the sign

protolambda:
@mslipper not sure if this affects your demo setup (changing flags and options), you can merge it when you like (or checkout the previous state of main branch for demo)

mslipper:
This makes the demo better by removing a bunch of hardcoded things in my demo setup 😄 . Merging. Thank you!

PR #264, opened 2022-03-17
opnode,specs: define max_l2_seq_time_diff, default to l2 block time

protolambda:
- Define the time a sequencer can get ahead of the last L1 block time as a rollup config param
- Default to block time
- Update specs
- Catch error when producing block that's too far ahead. Simply retry the next L2 block time interval to produce a block that's not too far ahead. This error was previously making noise in e2e.

I think we should consider experimenting with increasing the `max_l2_seq_time_diff`, since I expect the L2 blocks to be build at their timestamp. If we wait for the next L1 block, the L2 blocks we are building between the previous L1 head and the new L1 head are already old. I would expect this to be set to the *expected* L1 block time.

Also, when the L1 is not producing blocks for whatever reason, then this is the only parameter we have to keep the L2 going without making code changes, a good thing to have.

### DIFF
--- a/opnode/rollup/driver/step.go
+++ b/opnode/rollup/driver/step.go
@@ -16,6 +16,8 @@ import (
 	"github.com/ethereum/go-ethereum/log"
 )
 
+var newBlockTooNew = errors.New("cannot build new l2 block, constrained by l1 time")
+
 type outputImpl struct {
 	dl     Downloader
 	l2     Engine
@@ -37,8 +39,8 @@ func (d *outputImpl) newBlock(ctx context.Context, l2Finalized eth.BlockID, l2Pa
 	}
 
 	timestamp := l2Info.Time() + d.Config.BlockTime
-	if timestamp >= l1Info.Time() {
-		return l2Parent, nil, errors.New("L2 Timestamp is too large")
+	if timestamp >= l1Info.Time()+d.Config.MaxSequencerTimeDiff {
+		return l2Parent, nil, newBlockTooNew
 	}
 
 	var receipts types.Receipts
@@ -143,7 +145,7 @@ func (d *outputImpl) step(ctx context.Context, l2Head eth.BlockID, l2Finalized e
 	}
 	// Make batches contiguous
 	minL2Time := l2Info.Time() + d.Config.BlockTime
-	maxL2Time := l1Info.Time()
+	maxL2Time := l1Info.Time() + d.Config.MaxSequencerTimeDiff
 	batches = derive.FilterBatches(&d.Config, epoch, minL2Time, maxL2Time, batches)
 	batches = derive.SortedAndPreparedBatches(batches, uint64(epoch), d.Config.BlockTime, minL2Time, maxL2Time)
 
--- a/opnode/rollup/driver/state.go
+++ b/opnode/rollup/driver/state.go
@@ -2,6 +2,7 @@ package driver
 
 import (
 	"context"
+	"errors"
 	"time"
 
 	"github.com/ethereum-optimism/optimistic-specs/opnode/eth"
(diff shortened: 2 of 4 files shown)


opnode/rollup/driver/step.go
```diff
@@ -51,9 +51,9 @@ func (d *outputImpl) newBlock(ctx context.Context, l2Finalized eth.BlockID, l2Pa
 		return l2Parent, nil, fmt.Errorf("failed to fetch L1 block info of %s: %v", l1Origin, err)
 	}
 
-	timestamp := l2Info.Time() + d.Config.BlockTime
+	timestamp := l2Info.Time() + d.Config.MaxSequencerTimeDiff
```
trianglesphere, line 54: Shouldn't this still be the block time? This is for the block that is being created rather than verifying that a timestamp is valid.

protolambda: ah right, good catch

norswap:
Relevant discussion here: https://github.com/ethereum-optimism/optimistic-specs/discussions/251

We should probably try to hash out a solution to these issues beforehand, feel like we're not entirely on sync on these issues. Maybe a good topic for sync / office hours?

protolambda:
Rebased this onto main. Regardless of the discussion I think we should get this in. It cleans up the e2e test from creating an error log that's not actually an error. If you simply set the new config var to 0 it's not changing anything. With the parameter everyone can try it out so we don't have to make guesses in the discussion.

protolambda:
Rebased this onto main. Regardless of the discussion I think we should get this in. It cleans up the e2e test from creating an error log that's not actually an error. If you simply set the new config var to 0 it's not changing anything. With the parameter everyone can try it out so we don't have to make guesses in the discussion.

opnode/rollup/driver/step.go
```diff
@@ -37,8 +39,8 @@ func (d *outputImpl) newBlock(ctx context.Context, l2Finalized eth.BlockID, l2Pa
 	}
 
 	timestamp := l2Info.Time() + d.Config.BlockTime
-	if timestamp >= l1Info.Time() {
-		return l2Parent, nil, errors.New("L2 Timestamp is too large")
+	if timestamp >= l1Info.Time()+d.Config.MaxSequencerTimeDiff {
+		return l2Parent, nil, newBlockTooNew
```
tynes, line 43: Do we have any test coverage for this error case?

protolambda: The e2e test used to always hit this and then not recognize this error is expected behavior. Probably should have a more explicit test though, will try that

norswap:
@protolambda Agreed, it doesn't hurt to merge this

opnode/rollup/driver/step.go
```diff
@@ -143,7 +145,7 @@ func (d *outputImpl) step(ctx context.Context, l2Head eth.BlockID, l2Finalized e
 	}
 	// Make batches contiguous
 	minL2Time := l2Info.Time() + d.Config.BlockTime
-	maxL2Time := l1Info.Time()
+	maxL2Time := l1Info.Time() + d.Config.MaxSequencerTimeDiff
```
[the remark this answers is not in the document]
protolambda: That's required for a constant block time though (one of the bedrock design goals)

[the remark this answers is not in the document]
trianglesphere: So part of this is `SortedAndPreparedBatches` isn't a great name for what the function does (I'm thinking of switching it to `FillMissingBatches`), but I'm still pretty sure that this will break things if the `MaxSequencerTimeDiff` isn't set right in comparison to the L1 and L2 blocktimes

protolambda:
Closing in favor of #315

PR #265, opened 2022-03-18
Add forge

maurelian:
**Description**

1. Adds testing with forge, while keeping hardhat and the default hardhat dir structure. 
2. Replaces the hh tests for the `L1Block` contract with solidity tests.


packages/contracts/package.json
```diff
@@ -8,9 +8,12 @@
     "hardhat": "^2.7.1"
   },
   "scripts": {
-    "test": "hardhat test",
-    "test:slither": "slither .",
-    "build": "hardhat compile",
+    "build": "forge build --root . -c contracts --out forge-artifacts",
+    "test:hh": "yarn hardhat test",
+    "test:forge": "forge test --root . -c contracts --out forge-artifacts",
```
maurelian, line 13: The one janky thing I had to do to make this work was setup separate artifacts for each testing framework, because forge writes the artifacts into a different dir layout than hh does. 

I think it's fine for now.

tynes:
Going to need to add this as a step in CI: https://github.com/marketplace/actions/foundry-toolchain

tynes:
I thought that `hardhat compile` would choke on the `.t.sol` file because of this issue here: https://github.com/NomicFoundation/hardhat/issues/2306

maurelian:
I definitely had some challenges getting this set up with both tools, but I don't think that was one of them. The `.t.sol` files are valid solidity aren't they?

packages/contracts/contracts/test/L1Block.t.sol
```diff
@@ -0,0 +1,38 @@
...
+
+    CheatCodes cheats = CheatCodes(HEVM_ADDRESS);
+    L1Block lb;
+    address depositor;
+    bytes32 immutable NON_ZERO_HASH = keccak256(abi.encode(1));
+
+    function setUp() external {
+        lb = new L1Block();
+        depositor = lb.DEPOSITOR_ACCOUNT();
+        cheats.prank(depositor);
```
norswap, line 20: `prank` sets the passed address as the sender of any subsequent call?

packages/contracts/package.json
norswap: So currently `yarn build && yarn test:hh` does not work, right?

maurelian: I think test:hh actually compiles if it doesn't find the artifacts.

packages/contracts/contracts/test/L1Block.t.sol
maurelian: Correct

tynes:
See this https://onbjerg.github.io/foundry-book/

maurelian:
I've updated this PR now to fix the issues previously found in CI.

gakonst:
Pumped to see this merged. Let us know how we can be most helpful in making the integration easier!

maurelian:
@gakonst the biggest hurdle I faced was the discrepancy between how HH and Forge write the artifacts (HH maintaining the dir structure of the source contracts, forge placing them all in the root) ([screenshot](https://user-images.githubusercontent.com/23033765/159482841-f4a3d827-4ef6-4c8e-b523-186f326f0742.png)). 

The advice in the [tg chat](https://t.me/foundry_support/3112) didn't work for me because we have subfolders in the `./contracts` dir. So for now we're just using both build processes and have two distinct artifacts dirs.

PR #269, opened 2022-03-21
Add devnet docs + deposit task

mslipper:


README.md
```diff
@@ -16,6 +16,39 @@ the `alpha` branch.
...
+
+Then, run the following:
+
+```bash
+make devnet-up # starts the devnet
+make devnet-down # stops the devnet
+make devnet-clean # removes the devnet by deleting images and persistent volumes
+```
+
+L1 is accessible at `http://localhost:9545`, and L2 is accessible at `http://localhost:8545`. Any Ethereum tool - Metamask, `seth`, etc. - can use these endpoints. Note that you will need to specify the L2 chain ID manually if you use Metamask. The devnet's L2 chain ID is 901.
```
tynes, line 31: Perhaps we rip off the bandaid now and swap the ports because L1 is normall `8545` but we have had L2 at that for the longest time and its caused a lot of headaches for people and since this is a switch using a new codebase it could warrant us moving to something that is sane, altho it will be a breaking change for applications, but unless we migrate this to the monorepo then it'll be a breaking change anyways

tynes, line 31: Perhaps we rip off the bandaid now and swap the ports because L1 is normall `8545` but we have had L2 at that for the longest time and its caused a lot of headaches for people and since this is a switch using a new codebase it could warrant us moving to something that is sane, altho it will be a breaking change for applications, but unless we migrate this to the monorepo then it'll be a breaking change anyways

tynes: Thoughts @smartcontracts?

tynes: Thoughts @smartcontracts?

packages/contracts/slither.db.json
```diff
@@ -1 +1,306 @@
...
+[
```
tynes, line 1: Is this supposed to be in the gitignore?

tynes, line 1: Is this supposed to be in the gitignore?

packages/contracts/tasks/deposits.ts
```diff
@@ -0,0 +1,54 @@
+import { task, types } from 'hardhat/config'
+import { Contract, providers, utils, Wallet } from 'ethers'
+import dotenv from 'dotenv'
+
+import * as depositFeedArtifact from '../artifacts/contracts/L1/DepositFeed.sol/DepositFeed.json'
+
+task('deposit', 'Deposits funds unto Optimism bedrock.')
```
tynes, line 7: `unto` -> `into`

packages/contracts/tasks/deposits.ts
```diff
@@ -0,0 +1,54 @@
...
+    'depositContractAddr',
+    'Address of deposit contract.',
+    'deaddeaddeaddeaddeaddeaddeaddeaddead0001',
+    types.string
+  )
+  .setAction(async ({ to, amountEth, depositContractAddr }) => {
+    dotenv.config()
+
+    if (
+      !process.env.L1_PROVIDER_URL ||
```
tynes, line 20: nit: having 2 ways to config, these should be default values in hardhat params so that there is a unified CLI experience

packages/contracts/tasks/deposits.ts
tynes: also do we want to keep the name bedrock forever? i thought it was the release, then new releases would have new names

README.md
protolambda: We're supposed to switch over to port `8551` for the isolated engine-API actually. And we should use implement authentication on our client calls: https://github.com/ethereum/execution-apis/blob/main/src/engine/authentication.md

README.md
```diff
@@ -16,6 +16,39 @@ the `alpha` branch.
...
+
+Then, run the following:
+
+```bash
+make devnet-up # starts the devnet
+make devnet-down # stops the devnet
+make devnet-clean # removes the devnet by deleting images and persistent volumes
+```
+
+L1 is accessible at `http://localhost:8545`, and L2 is accessible at `http://localhost:8546`. Any Ethereum tool - Metamask, `seth`, etc. - can use these endpoints. Note that you will need to specify the L2 chain ID manually if you use Metamask. The devnet's L2 chain ID is 901.
```
protolambda, line 31: 8546 is commonly used for websocket RPC. (some clients have it on the same port, some don't, iirc)
Why not just use something like `9545` for L2?

mslipper: Sure, this is easy to change.

.gitignore
```diff
@@ -1,6 +1,8 @@
 node_modules
 yarn-error.log
 
+slither.db.json
```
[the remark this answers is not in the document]
maurelian: basically, I have a weak preference to keep it. But would accept if there's a strong reason to delete.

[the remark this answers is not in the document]
mslipper: It was causing lint errors and increasing the size of the diff. I can put it back if you'd prefer.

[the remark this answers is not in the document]
maurelian: I think the move is just to make sure the linters are ignoring it rather than git.

Makefile
```diff
@@ -2,6 +2,10 @@ opnode:
 	go build -o ./bin/op ./opnode/cmd
 .PHONY: opnode
 
+contracts:
+	cd packages/contracts && yarn build
```
maurelian, line 6: this doesn't seem to work on macos. 
This works:
```suggestion
	cd ./packages/contracts && yarn build
```

maurelian: Should it also do `yarn install` before build in case of a fresh clone?

maurelian (approved):

Makefile
mslipper: I think the error message when that happens is pretty clear.

PR #270, opened 2022-03-21
fix: test_basefee now tests basefee

mds1:
**Description**

Fixes basefee test, which previously tested the timestamp field

**Metadata**

Introduced in https://github.com/ethereum-optimism/optimistic-specs/pull/265


maurelian (approved):
Thanks!

PR #271, opened 2022-03-21
opnode/rollup/derive,specs: fix / harden deposit processing

protolambda:
- bugfix: verify batch does not contain deposit tx type (otherwise sequencers can insert fake deposits into batches)
- sanity-check receipts (most common deposit processing problem in eth2)
- use L2 block-height to ensure uniqueness of deposits when there are multiple L2 blocks with L1 info deposits per L1 block
- update specs (clarify deposit contents, move design text about uniqueness)
- update mock type in tests (l1 info interface has receipt hash now)
- test `ValidBatch` function

### DIFF
--- a/opnode/l1/source.go
+++ b/opnode/l1/source.go
@@ -6,6 +6,8 @@ import (
 	"fmt"
 	"math/big"
 
+	"github.com/ethereum/go-ethereum/trie"
+
 	"github.com/ethereum-optimism/optimistic-specs/opnode/eth"
 	"github.com/ethereum-optimism/optimistic-specs/opnode/rollup/derive"
 	"github.com/ethereum/go-ethereum"
@@ -64,9 +66,19 @@ func (s Source) FetchL1Info(ctx context.Context, id eth.BlockID) (derive.L1Info,
 	return s.client.BlockByHash(ctx, id.Hash)
 }
 
-func (s Source) FetchReceipts(ctx context.Context, id eth.BlockID) ([]*types.Receipt, error) {
+func (s Source) FetchReceipts(ctx context.Context, id eth.BlockID, receiptHash common.Hash) ([]*types.Receipt, error) {
 	_, receipts, err := s.Fetch(ctx, id)
-	return receipts, err
+	if err != nil {
+		return nil, err
+	}
+	// Sanity-check: external L1-RPC sources are notorious for not returning all receipts,
+	// or returning them out-of-order. Verify the receipts against the expected receipt-hash.
+	hasher := trie.NewStackTrie(nil)
+	computed := types.DeriveSha(types.Receipts(receipts), hasher)
+	if receiptHash != computed {
+		return nil, fmt.Errorf("failed to validate receipts of %s, computed receipt-hash %s does not match expected hash %d", id, computed, receiptHash)
+	}
+	return receipts, nil
 }
 
 func (s Source) FetchTransactions(ctx context.Context, window []eth.BlockID) ([]*types.Transaction, error) {
@@ -79,7 +91,6 @@ func (s Source) FetchTransactions(ctx context.Context, window []eth.BlockID) ([]
 		txns = append(txns, block.Transactions()...)
 	}
 	return txns, nil
-
 }
 func (s Source) L1HeadBlockRef(ctx context.Context) (eth.L1BlockRef, error) {
 	return s.l1BlockRefByNumber(ctx, nil)
--- a/opnode/rollup/derive/invert_test.go
+++ b/opnode/rollup/derive/invert_test.go
@@ -5,16 +5,21 @@ import (
 	"math/rand"
 	"testing"
 
+	"github.com/ethereum/go-ethereum/params"
+
+	"github.com/ethereum/go-ethereum/core/types"
+
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/stretchr/testify/assert"
 )
 
 type l1MockInfo struct {
-	num       uint64
-	time      uint64
-	hash      common.Hash
-	baseFee   *big.Int
-	mixDigest [32]byte
+	num         uint64
+	time        uint64
+	hash        common.Hash
+	baseFee     *big.Int
+	mixDigest   [32]byte
+	receiptRoot common.Hash
 }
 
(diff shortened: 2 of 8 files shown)


opnode/l1/source.go
```diff
@@ -74,8 +76,19 @@ func (s Source) Close() {
...
+func (s Source) FetchReceipts(ctx context.Context, id eth.BlockID, receiptHash common.Hash) ([]*types.Receipt, error) {
 	_, receipts, err := s.Fetch(ctx, id)
+	if err != nil {
+		return nil, err
+	}
+	// Sanity-check: external L1-RPC sources are notorious for not returning all receipts,
+	// or returning them out-of-order. Verify the receipts against the expected receipt-hash.
+	hasher := trie.NewStackTrie(nil)
+	computed := types.DeriveSha(types.Receipts(receipts), hasher)
+	if receiptHash != computed {
```
trianglesphere, line 89: I like this. Do we need to implement a retry loop?
Also, does this mean that we may need to do a cache purge on this if we don't compute the correct result?

specs/deposits.md
```diff
@@ -54,6 +55,8 @@ fields (rlp encoded in the order they appear here):
 
 [EIP-2718]: https://eips.ethereum.org/EIPS/eip-2718
 
+- `uint64 blockHeight`: the block-height of the L2 block
+- `uint64 transactionIndex`: the transaction-index within the L2 transactions list
 - `address from`: The address of the sender account.
```
trianglesphere, line 60: Maybe make explicit what is from the deposit itself and what is generated by the rollup node when translating the transaction.

trianglesphere (commented):
Nice find with the deposit tx. I also really like having a test for `validBatch`

opnode/l1/source.go
protolambda: The cache could wrap this, so it never caches when the error hits. Will look at that in the upcoming downloader PR. I don't think we should implement a retry-loop, the error will just appear like any other receipt retrieval issue, and it will retry later. Good point about cache purging though :+1:

specs/deposits.md
protolambda: That's already explicit in the `rollup-node.md` derivation process

opnode/rollup/derive/invert_test.go
```diff
@@ -37,17 +40,22 @@ func (l *l1MockInfo) MixDigest() common.Hash {
...
 func randomL1Info(rng *rand.Rand) *l1MockInfo {
 	return &l1MockInfo{
-		num:     rng.Uint64(),
-		time:    rng.Uint64(),
-		hash:    randomHash(rng),
-		baseFee: big.NewInt(rng.Int63n(1000_0000 * 1e9)), // a million GWEI
+		num:         rng.Uint64(),
+		time:        rng.Uint64(),
+		hash:        randomHash(rng),
+		baseFee:     big.NewInt(rng.Int63n(1000_0000 * 1e9)), // a million GWEI
```
tynes, line 57: slight nit, shouldn't `1000_0000` be `1_000_000` for ` a million gwei`, also there is `params.Gwei` instead of `1e9`

https://github.com/ethereum/go-ethereum/blob/d8ff53dfb8a516f47db37dbc7fd7ad18a1e8a125/params/denomination.go#L26

protolambda: Good catch! Although not intended part of the PR. Will include it though, small fix

protolambda:
will rebase, brb

opnode/rollup/driver/step.go
```diff
@@ -142,11 +141,11 @@ func (d *outputImpl) step(ctx context.Context, l2Head eth.BlockID, l2Finalized e
 	if err != nil {
 		return l2Head, fmt.Errorf("failed to create l1InfoTx: %w", err)
 	}
-	receipts, err := d.dl.FetchReceipts(fetchCtx, l1Input[0])
+	receipts, err := d.dl.FetchReceipts(fetchCtx, l1Input[0], l1Info.ReceiptHash())
```
tynes, line 144: I don't think this requires an interface change unless you wanted to be explicit because `FetchReceipts` internally has the block in context but its not currently assigned to a variable, see
https://github.com/ethereum-optimism/optimistic-specs/blob/b6f8fa6ad0a422e8772e2bcf2979d142fdd25a4b/opnode/l1/source.go#L81

opnode/rollup/derive/payload_attributes.go
```diff
@@ -240,6 +229,14 @@ func ValidBatch(batch *BatchData, config *rollup.Config, epoch rollup.Epoch, min
...
 		return false // too far in future
 	}
+	for _, txBytes := range batch.Transactions {
+		if len(txBytes) == 0 {
+			return false // transaction data must not be empty
+		}
+		if txBytes[0] == types.DepositTxType {
+			return false // sequencers may not embed any deposits into batch data
+		}
+	}
```
tynes, line 239: Could the sequencer embed an unknown transaction type to halt verifiers? I think that would be fraud provable with cannon but not 100% sure. This extends to sending malformed batches

protolambda: Invalid batches simply get skipped, see https://github.com/ethereum-optimism/optimistic-specs/blob/b7c16710b4463ce79ebbf14f2ad658a04ab004e3/opnode/rollup/derive/payload_attributes.go#L202-L203

So there's no halting of the chain, the sequencer is just wasting their gas on ignored data.

protolambda: The reason why we have to check the deposit type specifically here is because in no case it's valid, and the execution-layer cannot verify the reason for deposit inclusion, only the consensus-layer (rollup node) can during construction of the L2 block inputs.

opnode/rollup/driver/step.go
protolambda: We had separated the Downloader calls, as we may not pull all data from one place in the future (e.g. blobs). I'll look if this can be optimized when working on #187

opnode/l1/source.go
trianglesphere: Sounds good to rely on higher level retry.

trianglesphere (approved):
lgtm

PR #273, opened 2022-03-22
Update and test output oracle

maurelian:
WIP, but getting pretty close to completion.


packages/contracts/contracts/test/L2OutputOracle.t.sol
```diff
@@ -0,0 +1,119 @@
+//SPDX-License-Identifier: MIT
+pragma solidity 0.8.10;
+
+import { DSTest } from "../../lib/ds-test/src/test.sol";
+import { L2OutputOracle } from "../L1/L2OutputOracle.sol";
+
+interface CheatCodes {
```
tynes, line 7: This interface can be imported from here: https://github.com/brockelmore/forge-std/blob/master/src/Vm.sol

packages/contracts/contracts/L1/L2OutputOracle.sol
```diff
@@ -42,13 +53,16 @@ contract MockL2OutputOracle {
      * @param _l2Output The L2 output of the checkpoint block.
      * @param _timestamp The L2 block timestamp that resulted in _l2Output.
      */
-    function appendL2Output(bytes32 _l2Output, uint256 _timestamp) external {
+    function appendL2Output(bytes32 _l2Output, uint256 _timestamp) external payable onlyOwner {
+        // todo: separate owner and sequencer roles
```
tynes, line 57: If I am reading this correctly, there must be an L2 output for each L2 block? Based on the timestamp logic

packages/contracts/contracts/L1/L2OutputOracle.sol
```diff
@@ -1,10 +1,18 @@
...
-contract MockL2OutputOracle {
+contract L2OutputOracle is Ownable {
+
+    event l2OutputAppended(
+        bytes32 indexed _l2Output,
+        uint256 indexed _timestamp
+    );
+
     uint256 public submissionInterval;
     uint256 public l2BlockTime;
```
tynes, line 17: Some of these could be `immutable`

packages/contracts/contracts/L1/L2OutputOracle.sol
tynes: Also what can the `owner` do that the `sequencer` can't? I don't see an upgrade latch

maurelian: > If I am reading this correctly, there must be an L2 output for each L2 block? Based on the timestamp logic

I'm not seeing that, can you clarify what you mean? I think [this check](https://github.com/ethereum-optimism/optimistic-specs/blob/c2aff7c46e943c4df5f6c7a297c7980fcf0f45a6/packages/contracts/contracts/test/L2OutputOracle.t.sol#L116-L117) should cover it.

> Also what can the `owner` do that the `sequencer` can't? I don't see an upgrade latch

There's no upgrade path yet, but there will be. That said, I'll remove the todo for now, we'll track that elsewhere.

norswap: Also confused as to what a sequencer role would mean. Figure there's only one role, which is "whitelisted output submitter". Currently that's the sequencer (as an EOA whose key will be stored in a remote signer). In the future, multiple bonded submitter.

maurelian: If your point that this role should be called "proposer"then I agree.

packages/contracts/contracts/L1/L2OutputOracle.sol
```diff
@@ -0,0 +1,113 @@
+//SPDX-License-Identifier: MIT
```
maurelian, line 1: I tried very hard to make this appear as a diff against the MockL2OutputOracle.sol file which this was renamed from, but was not able to.

packages/contracts/contracts/L1/L2OutputOracle.sol
```diff
@@ -0,0 +1,113 @@
...
+     * @param _l2timestamp The L2 block timestamp of the target block.
+     */
+    function computeL2BlockNumber(uint256 _l2timestamp) external view returns (uint256) {
+        require(
+            _l2timestamp >= startingBlockTimestamp,
+            "Timestamp prior to startingBlockTimestamp"
+        );
+        // If _l2timestamp == startingBlockTimestamp, then the L2BlockNumber should be
+        // historicalTotalBlocks + 1
+        return historicalTotalBlocks + 1 + (_l2timestamp - startingBlockTimestamp) / l2BlockTime;
```
maurelian, line 111: The previous mock had what I think was an off-by-one error in it, but I'd appreciate if someone can sanity check that this comment is the correct behavior.

packages/contracts/contracts/L1/L2OutputOracle.sol
```diff
@@ -0,0 +1,120 @@
...
+     * Accepts an L2 output checkpoint and the timestamp of the corresponding L2
+     * block. The timestamp must be equal to the current value returned by
+     * `nextTimestamp()` in order to be accepted.
+     * This function may only be called by the Sequencer.
+     * @param _l2Output The L2 output of the checkpoint block.
+     * @param _l2timestamp The L2 block timestamp that resulted in _l2Output.
+     * @param _blockhash A block hash which must be included in the current chain.
+     * @param _blocknumber The block number with the specified block hash.
+     */
+    function appendL2Output(
```
tynes, line 68: This change in interface is going to break the go code in this codebase, but I suppose that is ok as long as we note that someplace to fix in the future

packages/contracts/contracts/L1/L2OutputOracle.sol
```diff
@@ -0,0 +1,120 @@
...
+    ) external payable onlyOwner {
+        require(_l2timestamp < block.timestamp, "Cannot append L2 output in future");
+        require(_l2timestamp == nextTimestamp(), "Timestamp not equal to next expected timestamp");
+        require(_l2Output != bytes32(0), "Cannot submit empty L2 output");
+
+        if (_blockhash != bytes32(0)) {
+            // This check allows the sequencer to append an output based on a given L1 block,
+            // without fear that it will be reorged out.
+            // It will also revert if the blockheight provided is more than 256 blocks behind the
+            // chain tip (as the hash will return as zero). This does open the door to a grieffing
```
tynes, line 82: `grieffing` -> `griefing`

packages/contracts/contracts/L1/L2OutputOracle.sol
```diff
@@ -0,0 +1,120 @@
...
+    uint256 public immutable historicalTotalBlocks;
+
+    // The timestamp of the first L2 block recorded in this contract.
+    uint256 public immutable startingBlockTimestamp;
+
+    // The timestamp of the most recent L2 block recorded in this contract.
+    uint256 public latestBlockTimestamp;
+
+    // A mapping from L2 timestamps to the output root for the block with that timestamp
+    mapping(uint256 => bytes32) public l2Outputs;
```
tynes, line 30: Any opinions on making this internal and adding a getter function? I'm not super opinionated

packages/contracts/contracts/L1/L2OutputOracle.sol
```diff
@@ -0,0 +1,120 @@
...
+        bytes32 _l2Output,
+        uint256 _l2timestamp,
+        bytes32 _blockhash,
+        uint256 _blocknumber
+    ) external payable onlyOwner {
+        require(_l2timestamp < block.timestamp, "Cannot append L2 output in future");
+        require(_l2timestamp == nextTimestamp(), "Timestamp not equal to next expected timestamp");
+        require(_l2Output != bytes32(0), "Cannot submit empty L2 output");
+
+        if (_blockhash != bytes32(0)) {
```
tynes, line 78: If I understand this correctly, its to protect against time bandit attacks?

packages/contracts/README.md
```diff
@@ -22,7 +22,6 @@ yarn build
 
 ## Running Tests
 
-
```
tynes, line 25: You should add here that you need to install the submodules to run the tests or to compile

`git submodule init` and `git submodule update`

packages/contracts/contracts/L1/L2OutputOracle.sol
```diff
@@ -0,0 +1,120 @@
+//SPDX-License-Identifier: MIT
+pragma solidity >=0.8.10;
+
+import { Ownable } from "../../lib/openzeppelin-contracts/contracts/access/Ownable.sol";
```
tynes, line 4: To make this be able to compile with both hardhat and foundry when the deps are installed with foundry vs npm, not sure how to modify the remappings for hardhat but with foundry its in the `foundry.toml`

```
[default]
src = 'src'
out = 'out'
libs = ['lib']
remappings = [
    'ds-test/=lib/ds-test/src/',
]
```

I think this is fine and we don't need to mess with it unless we decide to completely ditch hardhat for this repo

packages/contracts/contracts/L1/L2OutputOracle.sol
tynes: I would like to see some fuzz tests for this or something to see if there are any problems with precision

specs/proposals.md
```diff
@@ -62,6 +51,10 @@ If there is no newly finalized output, the service continues querying until it r
 output, and the appropriate timestamp, to the [L2 Output Commitment](#l2-output-commitment-smart-contract) contract's
 `appendL2Output()` function. The timestamp MUST be the next multiple of the `SUBMISSION_INTERVAL` value.
 
+> **Note regarding future work:** In the initial version of the system, the proposer will be the same entity as the
+> sequencer, which is a trusted role. In the future proposers will need to submit a bond in order to post L2 output
+> roots, and some or all of this bond may be taken in the event of a faulty proposal.
+
```
tynes, line 57: Do we need to think through an upgrade plan to save ourselves a headache in the future?

norswap: Yes, definitely (though not for this milestone). Should probably be part of a broader discussion on what sequencer decentralization is going to look like.

packages/contracts/contracts/L1/L2OutputOracle.sol
norswap: Yes - the `_blockhash` prevents the tx from being replayed in case of a L1 reorg (re-orgin past `_blocknumber`), which would allow a successful fault proof agains the now-incorrect output root.

The blockhash can only be check for ~256 blocks though, so L1 validators might censor for the duration. In that case we can omit the block hash and wait long enough we are sure the block at `_blocknumber` won't be re-orged, in which case replays won't do any damage.

specs/proposals.md
maurelian: Moved to a TLC issue

packages/contracts/contracts/L1/L2OutputOracle.sol
maurelian: Hardhat is OK with the paths used above. 
However I did use this as a prompt to setup a basic `foundry.toml` file.

packages/contracts/contracts/L1/L2OutputOracle.sol
maurelian: Good callout. 

I took a stab at it, managed update the makefile generate new bindings, but wasn't able to easily figure out how to provide the sequencer address to the constructor. 

[Here's where I got stuck](https://github.com/ethereum-optimism/optimistic-specs/commit/bdecf17fda04f92818aa48f43f20cfe716a9a76a#diff-310ca044c8a6e5b9faaab0c60de3c0b5f9d6952fbf11fa073a4c9f7f6c3759b8L237-L238).

specs/proposals.md
```diff
@@ -91,32 +95,28 @@ based on L2 time. It is OK to have L2 outputs submitted at larger or small inter
...
 /**
  * Accepts an L2 output checkpoint and the timestamp of the corresponding L2
  * block. The timestamp must be equal to the current value returned by
  * `nextTimestamp()` in order to be accepted.
+ * This function may only be called by the Sequencer.
  * @param _l2Output The L2 output of the checkpoint block.
- * @param _timestamp The L2 block timestamp that resulted in _l2Output.
- * @param _forkSpecifier A commitment to a specific fork.
+ * @param _l2timestamp The L2 block timestamp that resulted in _l2Output.
+ * @param _blockhash A block hash which must be included in the current chain.
```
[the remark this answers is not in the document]
maurelian: yes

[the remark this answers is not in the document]
maurelian: I think it makes sense to update the variable names to clarify that tho.

PR #274, opened 2022-03-22
feat: DepositFeed receive function

tynes:
**Description**

It is convenient for users to be able to deposit funds
directly to L2 by simply doing an ETH send to an address.
The prevents needing an ABI.

### DIFF
--- a/packages/contracts/contracts/L1/DepositFeed.sol
+++ b/packages/contracts/contracts/L1/DepositFeed.sol
@@ -27,6 +27,14 @@ contract DepositFeed {
     );
 
     /**
+     * Accepts value deposits where users can send ETH directly to
+     * the smart contract address and have the funds be deposited
+     */
+    receive() external payable {
+        depositTransaction(msg.sender, msg.value, 30000, false, bytes(""));
+    }
+
+    /**
      * Accepts deposits of ETH and data, and emits a TransactionDeposited event for use in deriving
      * deposit transactions.
      * @param _to The L2 destination address.
@@ -41,7 +49,7 @@ contract DepositFeed {
         uint256 _gasLimit,
         bool _isCreation,
         bytes memory _data
-    ) external payable {
+    ) public payable {
         if (_isCreation && _to != address(0)) {
             revert NonZeroCreationTarget();
         }
--- a/packages/contracts/test/L1/DepositFeed.spec.ts
+++ b/packages/contracts/test/L1/DepositFeed.spec.ts
@@ -310,6 +310,32 @@ describe('DepositFeed', () => {
           data: NON_ZERO_DATA,
         })
       })
+
+      it('when doing a simple send', async () => {
+        const balBefore = await ethers.provider.getBalance(depositFeed.address)
+
+        const tx = await signer.sendTransaction({
+          to: depositFeed.address,
+          value: NON_ZERO_VALUE,
+        })
+
+        const receipt = await tx.wait()
+        expect(receipt.status).to.equal(1)
+
+        const balAfter = await ethers.provider.getBalance(depositFeed.address)
+        const eventArgs = await decodeDepositEvent(depositFeed)
+
+        expect(balAfter.sub(balBefore)).to.equal(NON_ZERO_VALUE)
+        expect(eventArgs).to.deep.equal({
+          from: signerAddress,
+          to: await signer.getAddress(),
+          value: NON_ZERO_VALUE,
+          mint: NON_ZERO_VALUE,
+          gasLimit: BigNumber.from(30000),
+          isCreation: false,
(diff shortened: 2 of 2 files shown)


norswap:
Stating the obvious, this will require extra processing in the rollup-driver to work.

I suppose the idea is that we'll deduct the fee from the sent amount. If the sent amount is not even enough to pay for the L2 fee, then the amount is lost (sent to sequencer?).

Is the added complexity of multiplying the ways to deposit worth the gains? There is probably going to be an SDK that makes deposits relatively easy. Even if languages without the SDK, it's "just* a contract call on L1.

Seems like the real benefit to me is the gas savings. This is basically a trustless teleportr that can be used for large amounts too.

I feel like bridging small amounts from L1 will become less and less of a concern as time goes, and we can always keep running teleportr. Not convinced it's warranted to bake this into the design, but I'm not feeling extremely strongly about it either, worth discussing!

tynes:
> Stating the obvious, this will require extra processing in the rollup-driver to work.
> 
> I suppose the idea is that we'll deduct the fee from the sent amount. If the sent amount is not even enough to pay for the L2 fee, then the amount is lost (sent to sequencer?).
> 
> Is the added complexity of multiplying the ways to deposit worth the gains? There is probably going to be an SDK that makes deposits relatively easy. Even if languages without the SDK, it's "just* a contract call on L1.
> 
> Seems like the real benefit to me is the gas savings. This is basically a trustless teleportr that can be used for large amounts too.
> 
> I feel like bridging small amounts from L1 will become less and less of a concern as time goes, and we can always keep running teleportr. Not convinced it's warranted to bake this into the design, but I'm not feeling extremely strongly about it either, worth discussing!

I'm ok with closing this if it adds too much complexity!

protolambda:
`depositTransaction` is called right? So it still emits the same event etc. Why would it add extra complexity in the derivation function? I think it's fine. The hardcoded gas may not be as nice though, and fee payments of deposits may also need a default.

norswap:
Oooh, right. It emits the event. Thanks @protolambda for pointing out. Sorry @tynes, yeah I think this makes a lot of sense.
Feeling dumb  😅

tynes:
> Oooh, right. It emits the event. Thanks @protolambda for pointing out. Sorry @tynes, yeah I think this makes a lot of sense. Feeling dumb sweat_smile

All good! Was thinking we could name the deposit contract, something like `deposits.optimistic-ethereum.eth` or something, then people could just send eth to it

packages/contracts/contracts/L1/DepositFeed.sol
```diff
@@ -26,6 +26,14 @@ contract DepositFeed {
         bytes data
     );
 
+    /**
+     * Accepts value deposits where users can send ETH directly to
+     * the smart contract address and have the funds be deposited
+     */
+    receive() external payable {
+        depositTransaction(msg.sender, msg.value, 30000, false, bytes(""));
```
trianglesphere, line 34: Is 30k the intrinsic gas or should it be lower?
Someone else will have to verify that gas works / there is not trickery that can be done here & this can be integrated with deposit fees.

trianglesphere (commented):
I really like this UX

packages/contracts/contracts/L1/DepositFeed.sol
norswap: Isn't an ETH transfer only 21k?

maurelian:
you can fix the slither check with this patch: 
[0001-Update-slither-db.patch.txt](https://github.com/ethereum-optimism/optimistic-specs/files/8345166/0001-Update-slither-db.patch.txt)

I think it's too finnicky, we should reduce it to a warning.

packages/contracts/contracts/L1/DepositFeed.sol
tynes: To be honest, I just picked a number here than I was sure would be high enough. I feel like we want a buffer over the minimal 21k and an e2e test to make sure that the number works with L2 geth

maurelian: I like this change, I think the `receive()` function is a definite UX improvement, but...

> and an e2e test to make sure that the number works with L2 geth

If we think this would be a useful test to have, then IMO we should add it in this PR.

maurelian: That could be done fairly easily building on #269

trianglesphere: We can also make this part of the e2e go tests.

mslipper:
I'm going to close this for now since @maurelian has taken it over. @maurelian - please reopen when you're ready.

PR #275, opened 2022-03-23
L1 Downloader: caching, batching, improved typing, trust-RPC flag

### LINKED ISSUE #187: Harden Downloader w/ batching, concurrency limits and memoization

protolambda:
Changes:
- New downloader
  - define `HeaderInfo` type that implements `derive.L1Info`, extend interface with state-root.
    - caches block-hash and uint64-typed number
    - used to avoid uncle header lookups in block calls
    - optional to verify cached RPC information when loading it
  - cache headers from range calls, supply max-blocks as argument instead of hardcoding
  - fetch all headers of range call in batch
  - cache headers from any individual call (cache by hash)
  - cache receipt lists by block hash
  - cache transaction lists by block hash
  - fetch all receipts of a block with a single batch, with retries of failed items
  - fetch all blocks with all transaction list (with per-block caching) of l1 window call. Any cache-hits are not part of batch-request.
  - small interface changes to separate `derive.L1Info` and `types.Transactions`, to avoid unnecessary `types.Block` inefficiencies, and fetching the L1 info with receipts at the same time, as in previous PR comment from Mark
  - Wrap the batchCallContextFn function to call the actual `batchCallContextFn` in multiple parallel requests to make the downloader faster. (Assuming geth does not already do parallel lookups for batch calls, based on benchmarking from Josh). This also splits a large batch into smaller batches, and incorporates retries into batches when individual items fail.
- Testing:
  - The parallel/splitting etc. logic of the batching is more extensively encapsulated & tested. (`batching_test.go`)
  - The `Source` is tested with a mock RPC, fetching blocks with transactions and headers, including hitting the cache.  (`source_test.go`)

Fix #187


opnode/l1/receipts.go
```diff
@@ -0,0 +1,85 @@
...
+			return nil, fmt.Errorf("no transactions, but got non-empty receipt trie root: %s", receiptHash)
+		}
+		return nil, nil
+	}
+
+	receipts := make([]*types.Receipt, len(txs))
+	receiptRequests := make([]rpc.BatchElem, len(txs))
+	for i := 0; i < len(txs); i++ {
+		receipts[i] = new(types.Receipt)
+		receiptRequests[i] = rpc.BatchElem{
```
tynes, line 30: I would imagine there are some limits in geth or at infura/alchemy that impose a limit on the number of batch elements in an rpc request. For example, our `proxyd` allows for only 100 per RPC request, see https://github.com/ethereum-optimism/optimism/blob/3b82da3e886dad9a1bf9ed441b032f192ce9bed9/go/proxyd/server.go#L27

protolambda: Good catch, I'll refactor it to split it into multiple batch requests

tynes: Just looked at geth a bit, I think that geth will only block on size of request/response and duration of request processing, I didn't see any code that restricted based on size of batch but I didn't test anything. Unsure what other clients do and unsure if infra providers do anything special at their proxy layer

protolambda: Implemented `fetchBatched`, with edge-case tests for splitting/retries/parallel/errors behavior. It's a drop-in replacement for the existing batch-call function type. During testing of the source we just mock the outer single batch calls, but we use the tested splitting&parallelizing implementation of it in E2E.

trianglesphere (commented):
I've looked at everything but batching/batching_test.

At a high level it doesn't seem like we're limiting in-flight parallelism and I'm gonna need to look through the batching code a bit more to be comfortable with it. Otherwise it integrates pretty well.

opnode/l1/batching.go
```diff
@@ -0,0 +1,132 @@
...
+	awaited := len(requests)
+
+	// start initial round of parallel requests
+	count := parallelRequests()
+
+	// We slow down additional batch requests to not spam the server.
+	retryTicker := time.NewTicker(time.Millisecond * 20)
+	defer retryTicker.Stop()
+
+	for {
```
trianglesphere, line 109: Could you leave comment here explaining this a bit more? It took me a while to convince myself that this is correct and that `requests` did not have a data race.

trianglesphere (commented):
I went through `batching.go` and `batching_test.go`. I think they're correct, but it took a bit to convince myself of that.

opnode/l1/batching.go
```diff
@@ -0,0 +1,132 @@
...
+				elem.Error = nil // reset, we'll try this element again
+				failed = append(failed, elem)
+				continue
+			}
+		}
+		return failed, nil
+	}
+
+	// limit capacity, don't write to underlying array on retries
+	requests = requests[:len(requests):len(requests)]
```
trianglesphere, line 49: I prefer an explicit copy or stating that `fetchBatched` modifies `requests`

opnode/l1/batching.go
```diff
@@ -0,0 +1,132 @@
...
+
+type batchResult struct {
+	failed  []rpc.BatchElem // if anything has to be retried
+	err     error           // if the batch as a whole failed
+	success int             // amount of items that completed successfully
+}
+
+// fetchBatched fetches the given requests in batches of at most maxPerBatch elements, and with at most maxRetry retries per batch.
+// Batch requests may be split into maxParallel go-routines.
+func fetchBatched(ctx context.Context, log log.Logger, requests []rpc.BatchElem, getBatch batchCallContextFn, maxRetry int, maxPerBatch int, maxParallel int) error {
```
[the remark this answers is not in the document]
trianglesphere: I figured this out, and I don't love how the low level API works in this regard.

opnode/rollup/driver/state.go
```diff
@@ -137,6 +141,7 @@ func (s *state) loop() {
 		// case <-l1Poll.C:
 		// case <-l2Poll.C:
 		case <-s.done:
+			atomic.AddUint32(&s.closed, 1)
```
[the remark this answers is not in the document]
protolambda: agreed, temporary work around to make tests stable: when pausing go routines during debugging I hit a crash because the logger would run after closing

opnode/rollup/driver/step.go
```diff
@@ -6,13 +6,14 @@ import (
 	"fmt"
 	"time"
 
+	"github.com/ethereum/go-ethereum/core/types"
```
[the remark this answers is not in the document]
protolambda: No, will revert

opnode/l1/source.go
```diff
@@ -4,161 +4,335 @@ import (
...
-	for i := begin.Number + 1; i < begin.Number+maxBlocks+1; i++ {
-		n, err := s.L1BlockRefByNumber(ctx, i)
-		if err != nil {
-			return nil, fmt.Errorf("failed to fetch L1 block %v: %w", i, err)
+// L1Range returns a range of L1 block beginning just after begin, up to max blocks.
+// This batch-requests all blocks by number in the range at once, and then verifies the consistency
+func (s *Source) L1Range(ctx context.Context, begin eth.BlockID, max uint64) ([]eth.BlockID, error) {
+
+	headerRequests := make([]rpc.BatchElem, max)
+	for i := uint64(0); i < max; i++ {
```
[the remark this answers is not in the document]
protolambda: Currently it fetches everything by number, and then returns the block ids after checking the headers form one canonical chain. If it reorgs or misses anything at any point in the batch response, it will just return the prefix of the range (function may return any number of blocks to add to the existing buffer). I think this is fastest, because the batch can be optimized and fully parallel if we configure it that way.

opnode/l1/source.go
```diff
@@ -4,161 +4,335 @@ import (
...
 type Source struct {
-	client     *ethclient.Client
-	downloader *Downloader
+	client RPCClient
+
+	batchCall batchCallContextFn
+
+	trustRPC bool
+
+	// cache receipts in bundles per block hash
```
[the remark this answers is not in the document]
protolambda: In general there are lots of places we can add metrics, think it'll be one of the post-audit-prep tasks

opnode/l1/batching.go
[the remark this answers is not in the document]
protolambda: I preserved the existing batch RPC method that geth uses. It's not pretty, but does work well in some ways. I think it's primarily an artifacts of not having generics.

opnode/l1/batching.go
protolambda: This way we avoid the unnecessary copy of a potentially large request slice in the happy case when no item is retried. This is what the capacity-slicing in golang is meant for: to limit writes to the backing array.

protolambda:
> At a high level it doesn't seem like we're limiting in-flight parallelism

Added in [0c075da](https://github.com/ethereum-optimism/optimistic-specs/pull/275/commits/0c075da9825c045a9d6462d67ab013b468935855) by wrapping the RPC client with a simple semaphore to limit the number of concurrent function calls.

norswap:
Just a meta comment that in general, smaller PRs would be nicer. Especially isolating refactorings is high value since it touches so many files. I understand the current context is a little bit special, but something to consider for later.

PR #279, opened 2022-03-24
opnode: Switch to BlockRefs in state loop

trianglesphere:
**Description**
This removes having to separately keep track of L1Origin and L1Base.
This is nice because those name are confusing and separately used
and it means less information to keep track of.

### DIFF
--- a/opnode/rollup/driver/state.go
+++ b/opnode/rollup/driver/state.go
@@ -13,13 +13,11 @@ import (
 
 type state struct {
 	// Chain State
-	l1Head      eth.BlockID   // Latest recorded head of the L1 Chain
-	l2Head      eth.BlockID   // L2 Unsafe Head
-	l1Origin    eth.BlockID   // L1 Origin of the L2 Unsafe head. For sequencing only.
-	l2SafeHead  eth.BlockID   // L2 Safe Head - this is the head of the L2 chain as derived from L1 (thus it is Sequencer window blocks behind)
-	l1Base      eth.BlockID   // L1 Parent of L2 Safe Head block
-	l2Finalized eth.BlockID   // L2 Block that will never be reversed
-	l1Window    []eth.BlockID // l1Window buffers the next L1 block IDs to derive new L2 blocks from, with increasing block height.
+	l1Head      eth.L1BlockRef // Latest recorded head of the L1 Chain
+	l2Head      eth.L2BlockRef // L2 Unsafe Head
+	l2SafeHead  eth.L2BlockRef // L2 Safe Head - this is the head of the L2 chain as derived from L1 (thus it is Sequencer window blocks behind)
+	l2Finalized eth.BlockID    // L2 Block that will never be reversed
+	l1Window    []eth.BlockID  // l1Window buffers the next L1 block IDs to derive new L2 blocks from, with increasing block height.
 
 	// Rollup config
 	Config    rollup.Config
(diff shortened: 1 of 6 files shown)


opnode/rollup/driver/step.go
```diff
@@ -84,18 +89,18 @@ func (d *outputImpl) newBlock(ctx context.Context, l2Finalized eth.BlockID, l2Pa
...
+	ref, err := derive.BlockReferences(payload, &d.Config.Genesis)
+	return ref, batch, err
 }
 
 // DriverStep derives and processes one or more L2 blocks from the given sequencing window of L1 blocks.
 // An incomplete sequencing window will result in an incomplete L2 chain if so.
 //
 // After the step completes it returns the block ID of the last processed L2 block, even if an error occurs.
-func (d *outputImpl) step(ctx context.Context, l2Head eth.BlockID, l2Finalized eth.BlockID, unsafeL2Head eth.BlockID, l1Input []eth.BlockID) (out eth.BlockID, err error) {
+func (d *outputImpl) step(ctx context.Context, l2Head eth.L2BlockRef, l2Finalized eth.BlockID, unsafeL2Head eth.BlockID, l1Input []eth.BlockID) (out eth.L2BlockRef, err error) {
```
protolambda, line 103: Shouldn't we be removing `l1Input` now that the `l2Head` has L1 block ID info? What's the point of changing to `BlockRef` without using the extra info?

opnode/rollup/driver/step.go
```diff
@@ -23,14 +23,15 @@ type outputImpl struct {
 	Config rollup.Config
 }
 
-func (d *outputImpl) newBlock(ctx context.Context, l2Finalized eth.BlockID, l2Parent eth.BlockID, l2Safe eth.BlockID, l1Origin eth.BlockID, includeDeposits bool) (eth.BlockID, *derive.BatchData, error) {
-	d.log.Info("creating new block", "l2Parent", l2Parent, "l1Origin", l1Origin, "includeDeposits", includeDeposits)
+func (d *outputImpl) newBlock(ctx context.Context, l2Finalized eth.BlockID, l2Parent eth.L2BlockRef, l2Safe eth.BlockID, l1Origin eth.BlockID) (eth.L2BlockRef, *derive.BatchData, error) {
```
protolambda, line 26: Same question here as in the other comment

opnode/rollup/driver/state.go
```diff
@@ -36,6 +34,11 @@ type state struct {
 	done chan struct{}
 }
 
+// // shouldRunEpoch returns true if there is a full sequencing window between the L2 Safe Head's L1 Origin and the L1 Head.
+// func (s *state) shouldRunEpoch() bool {
+// 	return s.l1Head.Self.Number-s.l2SafeHead.L1Origin.Number >= s.Config.SeqWindowSize
+// }
+
```
protolambda, line 41: ?

protolambda (changes requested):
confused by some of the id -> ref changes

opnode/rollup/driver/state.go
trianglesphere: Dead code from other changes. Will cut.

opnode/rollup/driver/step.go
trianglesphere: This is for timestamps later.

We could also be checking that `l1Input[0].Parent.Hash == l2Head.L1Origin.Hash` as this function advances full epochs at a time.

trianglesphere:
See https://github.com/ethereum-optimism/optimistic-specs/pull/280#issuecomment-1079369761

PR #280, opened 2022-03-24
opnode: Add Time field to BlockId

trianglesphere:
**Description**
This field is very helpful for sequencing. Otherwise it needs to be
maintained on the side.

### DIFF
--- a/opnode/l1/source.go
+++ b/opnode/l1/source.go
@@ -114,7 +114,7 @@ func (s Source) l1BlockRefByNumber(ctx context.Context, number *big.Int) (eth.L1
 		parentNum -= 1
 	}
 	return eth.L1BlockRef{
-		Self:   eth.BlockID{Hash: header.Hash(), Number: l1Num},
+		Self:   eth.BlockID{Hash: header.Hash(), Number: l1Num, Time: header.Time},
 		Parent: eth.BlockID{Hash: header.ParentHash, Number: parentNum},
 	}, nil
 }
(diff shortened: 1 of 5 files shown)


opnode/l1/source.go
```diff
@@ -114,7 +114,7 @@ func (s Source) l1BlockRefByNumber(ctx context.Context, number *big.Int) (eth.L1
 		parentNum -= 1
 	}
 	return eth.L1BlockRef{
-		Self:   eth.BlockID{Hash: header.Hash(), Number: l1Num},
+		Self:   eth.BlockID{Hash: header.Hash(), Number: l1Num, Time: header.Time},
 		Parent: eth.BlockID{Hash: header.ParentHash, Number: parentNum},
```
protolambda, line 118: This highlights how not everywhere we can have the timestamp readily available. The parent hash, and number are known from the next block header, but time is not.

Can we please introduce `eth.TimestampedID` (or `eth.TimeBlockID` or something) so we are explicit about the timestamp requirement? I like the existing minimal ID and don't want to leave time zero for optimization, that seems prone to bugs.

protolambda (changes requested):
Good idea, but I think we should keep it as a separate thing from `eth.BlockID`, and change the uses where applicable, so we don't have zeroed timestamps by accident / optimization.

trianglesphere:
Closing this and #279 in favor to of the following:

Migrate block ID to block ref and update the block refs to the following:
```
L2 Block Ref
- Hash
- Number
- Parent Hash
- Time
- L1 Origin Hash
- L1 Origin Number

L1 Block Ref
- Hash
- Number
- Parent hash
- Time
```

PR #281, opened 2022-03-25
Parallelize solidity CI checks

maurelian:
Splits the workflow into 3 different steps which can run in parallel.

It also allows the checks to pass in the case of a slither error, because slither is way to finnicky, and not all that helpful during active development.


protolambda (approved):
LGTM

PR #282, opened 2022-03-25
Implement opnode RPC server

### LINKED ISSUE #261: Add inital Opnode server endpoints
We need to expose the following RPC endpoints on the opnode itself (taken from the spec [here](https://github.com/ethereum-optimism/optimistic-specs/blob/32fe4173cd0c86b28a12002b93a95c012123406a/specs/rollup-node.md?plain=1#L280)):

- method: `optimism_outputAtBlock`
- params:
  1. `QUANTITY` - L2 integer block number, or the strings `"safe"`, `"latest"`, or `"pending"`
- returns:
  2.`DATA` - The 32 byte output root

Additionally, the opnode should expose a `/healthz` endpoint that exposes the following payload:

- Git commit
- Git version

Inphi:
- Implement optimism_outputAtBlock to be used by L2OSS
- /healthz endpoint

Metrics collection will be added later once I have a better idea on how the endpoint will be used.
Fixes #261


opnode/node/server.go
```diff
@@ -0,0 +1,257 @@
...
+		writeRPCError(s.log, w, nil, err)
+		return
+	}
+
+	if err := ValidateRPCReq(req); err != nil {
+		writeRPCRes(s.log, w, NewRPCErrorRes(nil, err))
+		return
+	}
+
+	if req.Method != "optimism_outputAtBlock" {
```
tynes, line 116: Any opinions on using the rpc package from `go-ethereum`?

https://github.com/ethereum/go-ethereum/blob/a8040bc2c51605a4cca9e48cac83ff14928d50c2/node/rpcstack.go#L284

tynes: A few reasons why I think this is a good idea. It makes us become more familiar with the geth codebase, it makes writing RPC handlers really easy, its very hardened and it makes it easy for contributors familiar with geth be familiar with our codebase. I'm not 100% sure it can be easily used standalone but I'm pretty sure that it can be

Inphi: I have no objections to using the package. I'll see if that can be used.

opnode/node/server.go
```diff
@@ -0,0 +1,257 @@
...
+		HistoryAccumulatorRoot: l2.Bytes32{}, // unused. zeroed
+		Extension:              l2.Bytes32{}, // unused. zeroed
+	}
+
+	var buf bytes.Buffer
+	bufWriter := bufio.NewWriter(&buf)
+	l2OutputSSZ := zssz.GetSSZ((*l2.L2Output)(nil))
+	if _, err := zssz.Encode(bufWriter, &l2Output, l2OutputSSZ); err != nil {
+		return NewRPCErrorRes(req.ID, err)
+	}
```
protolambda, line 239: please don't use ZSSZ. Although it works and it's fast, and was fuzzed at some point, I have not updated it for 2 years. Instead, use https://github.com/protolambda/ztyp if you need recent SSZ code (spoiler: we don't need it at all, for now). It's a bit slower, but doesn't use unsafe Go, and in https://github.com/protolambda/zrnt/ it's widely tested against the ethereum standard tests.

Also, encoding and then hashing the SSZ data effectively **does nothing useful for the verifier**. The original intention was to use **hash-tree-root**, a **different** feature of SSZ than encoding, that merkleizes the structure, enabling an efficient proof to the contents of the block.

Not sure if I want the dependency at all though. We are trying to simplify the withdrawals format, and don't need SSZ for that.

If we are going with the simplified output-root format let's just do something like:
```
var buf bytes.Buffer
buf.Write([]byte{versionPrefixByte})  // or we can do a 32 bytes version, if that is nicer to proof in EVM
buf.Write(l2Output.StateRoot[:])
buf.Write(l2Output.WithdrawalStorageRoot[:])
buf.Write(l2Output.BlockHash[:])  // not the ssz hash-tree-root, just regular eth1 block-hash
hash := crypto.Keccak256(buf.Bytes())
```

cc @maurelian we've to update the spec with the simplified version, so we don't end up implementing the wrong thing (sorry @Inphi !)

opnode/node/server.go
protolambda: If we can, definitely just import things from the go-ethereum `rpc` package. We already use the client-side of it in other places, and we don't have to roll our own (cool stuff though)

opnode/node/server.go
```diff
@@ -0,0 +1,142 @@
...
+	err = n.l2RPCClient.CallContext(ctx, &getProofResponse, "eth_getProof", n.withdrawalContractAddr, nil, toBlockNumArg(numberB))
+	if err != nil {
+		return nil, err
+	} else if getProofResponse == nil {
+		return nil, ethereum.NotFound
+	}
+
+	var l2OutputRootVersion l2.Bytes32 // it's zero for now
+
+	var buf bytes.Buffer
```
tynes, line 120: Do you think you could move this implementation into a common package?

opnode/node/server.go
```diff
@@ -0,0 +1,142 @@
...
+	withdrawalContractAddr common.Address
+	log                    log.Logger
+}
+
+type getProof struct {
+	Address     common.Address `json:"address"`
+	StorageHash common.Hash    `json:"storage_hash"`
+}
+
+func (n *nodeAPI) OutputAtBlock(ctx context.Context, number rpc.BlockNumber) ([]l2.Bytes32, error) {
```
tynes, line 98: Do you mind moving this function to `opnode/node/handlers.go`?

Inphi: yup

opnode/node/server.go
Inphi: you mean the state root computation?

opnode/node/api.go
```diff
@@ -45,17 +46,22 @@ func (n *nodeAPI) OutputAtBlock(ctx context.Context, number rpc.BlockNumber) ([]
...
 	}
 
-	storageHash, err := n.client.GetProof(ctx, n.withdrawalContractAddr, toBlockNumArg(number))
+	accountRoot, proof, err := n.client.GetProof(ctx, n.withdrawalContractAddr, toBlockNumArg(number))
 	if err != nil {
 		n.log.Error("failed to get contract proof", "err", err)
 		return nil, err
 	}
-	if storageHash == nil {
+	if accountRoot == nil {
```
protolambda, line 52: You're still returning the storage hash in the `(c *l2EthClientImpl) GetProof` implementation. The account-proof will fail because of that.

protolambda: fixed it

opnode/flags/flags.go
```diff
@@ -31,6 +31,30 @@ var (
...
+		Usage:    "RPC listening port",
+		Required: true,
+		EnvVar:   prefixEnvVar("RPC_PORT"),
+	}
+	WithdrawalContractAddr = cli.StringFlag{
+		Name:     "rpc.withdrawalcontractaddress",
+		Usage:    "Address of the Withdrawal contract. Used to serve RPC",
+		Required: true,
+		EnvVar:   prefixEnvVar("WITHDRAWAL_CONTRACT_ADDR"),
+	}
```
maurelian, line 57: This is the L2 contract address we're proving the state of right? 
If so, then it is currently spec'd to be a predeploy at `0x4200000000000000000000000000000000000015`. 

I'd propose either: 
1. hardcode it in which would likely simplify some code.
2. make the flag optional and default to the the above address.

Inphi: Yup, that's the L2 contract address. I'll make it optional with the default so the contract itself can be mocked for integration testing if needed.

opnode/flags/flags.go
```diff
@@ -31,6 +31,30 @@ var (
 		Required: true,
 		EnvVar:   prefixEnvVar("ROLLUP_CONFIG"),
 	}
+	L2NodeAddr = cli.StringFlag{
+		Name:     "eth.l2",
```
tynes, line 35: Slight nit - there should be a general 1:1 mapping between env vars and flag names

Inphi: 👍 . changed the flag name here.

protolambda (approved):
LGTM

PR #284, opened 2022-03-25
opnode: New Sync

trianglesphere:
**Description**
Semi WIP new sync + some other changes to make the system still work.

**Additional context**

**Metadata**
- Fixes #[Link to Issue]

### DIFF
--- a/opnode/l1/source.go
+++ b/opnode/l1/source.go
@@ -92,12 +92,29 @@ func (s Source) FetchTransactions(ctx context.Context, window []eth.BlockID) ([]
 	}
 	return txns, nil
 }
+
 func (s Source) L1HeadBlockRef(ctx context.Context) (eth.L1BlockRef, error) {
 	return s.l1BlockRefByNumber(ctx, nil)
 }
 
 func (s Source) L1BlockRefByNumber(ctx context.Context, l1Num uint64) (eth.L1BlockRef, error) {
 	return s.l1BlockRefByNumber(ctx, new(big.Int).SetUint64(l1Num))
+}
+
+func (s Source) L1BlockRefByHash(ctx context.Context, hash common.Hash) (eth.L1BlockRef, error) {
+	header, err := s.client.HeaderByHash(ctx, hash)
+	if err != nil {
+		return eth.L1BlockRef{}, err
+	}
+	l1Num := header.Number.Uint64()
+	parentNum := l1Num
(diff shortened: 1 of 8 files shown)


opnode/l1/source.go
```diff
@@ -100,6 +101,22 @@ func (s Source) L1BlockRefByNumber(ctx context.Context, l1Num uint64) (eth.L1Blo
 	return s.l1BlockRefByNumber(ctx, new(big.Int).SetUint64(l1Num))
 }
 
+func (s Source) L1BlockRefByHash(ctx context.Context, hash common.Hash) (eth.L1BlockRef, error) {
+	header, err := s.client.HeaderByHash(ctx, hash)
+	if err != nil {
+		return eth.L1BlockRef{}, err
+	}
+	l1Num := header.Number.Uint64()
+	parentNum := l1Num
```
tynes, line 110: Does this need the intermediate variable here? Also should it be `if parentNum == 0`?

opnode/rollup/driver/fake_chain.go
```diff
@@ -117,6 +117,16 @@ func (m *fakeChainSource) L1BlockRefByNumber(ctx context.Context, l1Num uint64)
 	return m.l1s[m.l1reorg][l1Num], nil
 }
 
+func (m *fakeChainSource) L1BlockRefByHash(ctx context.Context, l1Hash common.Hash) (eth.L1BlockRef, error) {
+	m.log.Trace("L1BlockRefByHash", "l1Hash", l1Hash, "l1Head", m.l1head, "reorg", m.l1reorg)
```
tynes, line 121: Does this need a mutex?

opnode/l1/source.go
trianglesphere: I'm pretty sure that this is correct. The intent is to saturate subtraction at 0 (i.e. parent block of genesis is not well formed and has 0 as the parent block number).

See https://github.com/ethereum-optimism/optimistic-specs/pull/285/files#r836779844

opnode/rollup/driver/fake_chain.go
trianglesphere: I don't believe so. There is a data race in the driver test (and only test), but I think it's related to reading state variables.

trianglesphere:
Closing in favor of #299 (this + some changes rebased on top the new block ref format).

PR #285, opened 2022-03-25
opnode: New blockref format

trianglesphere:
**Description**
This flattens the Block ID fields and adds a time field to
the Block References. The time field is important for sequencing.
It needs to be in the BlockRef instead of the BlockID because
it is hard to reliably fill out the timestamp field for
Parent and L1Origin block ids.

This approach also removes the ParentNumber because it can
be easily calculated if needed, but was not being used.

It then removes having to separately keep track of L1Origin and L1Base.
This is nice because those name are confusing and separately used
and it means less information to keep track of.

### DIFF
--- a/opnode/eth/id.go
+++ b/opnode/eth/id.go
@@ -22,32 +22,74 @@ func (id BlockID) TerminalString() string {
 }
 
 type L2BlockRef struct {
-	Self     BlockID `json:"self"`
-	Parent   BlockID `json:"parent"`
-	L1Origin BlockID `json:"l1_origin"`
+	Hash       common.Hash `json:"hash"`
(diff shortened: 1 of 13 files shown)


opnode/eth/id.go
```diff
@@ -22,32 +22,74 @@ func (id BlockID) TerminalString() string {
...
+	return BlockID{
+		Hash:   id.Hash,
+		Number: id.Number,
+	}
+}
+
+func (id L1BlockRef) ParentId() BlockID {
+	n := id.Id().Number
+	// Saturate at 0 with subtraction
+	if n > 0 {
```
trianglesphere, line 69: @tynes new version

PR #286, opened 2022-03-26
specs: Add overview page

mslipper:
Hey all! Opening this PR to get feedback on the Bedrock overview doc I've been talking about. Its goals are as follows:

- Provide a high-level technical overview of how Bedrock works without being a formal specification.
- Walk readers through core flows step-by-step, and use diagrams to illustrate each step.
- Link out to other parts of the spec in case people want to learn more.

The ideal audience is someone on the systems team who wants to understand the spec from a high-level before jumping into details.

I focused on the system components, deposits, and L2 chain derivation. I diagrammed the safe head/unsafe head concept under a sequencing heading, but will need additional time and familiarity with the spec to properly articulate that section. Would love to chat with @trianglesphere about this to improve my understanding here.

Let me know what you think, and please check for accuracy - I'm learning the protocol myself and my understanding may be wrong 😄.


mslipper:
Updated to bring back the `introduction.md` file from the `alpha` branch, with updates to match how the current spec works.

Also updated some copy around sequencing.

protolambda:
Some feedback on the diagrams in overview.md:

The components diagram is wrong: "Read Batches"/"Read Outputs" are not reading from the submitter processes, but from any L1 RPC endpoint. And currently we're not reading outputs, at all. And we are not reading from a "sequencer" (red color), any verifier could be reading from any L1 node.

The other diagram in the tx/block propagation section shows the same confusing relation.

In the first Engine API diagram: The "FSS" should be "FCS" to represent ForkChoiceState in the diagram.

And the colors in the sequencing section are not explained (but intuitively seem wrong, i.e. the arrow from 4 to the red 6 block, and then 7 being yellow-ish again). Can we leave this section out, and move that to a separate PR, instead of putting in a confusing placeholder?

specs/introduction.md
```diff
@@ -0,0 +1,127 @@
...
+
+There are three actors in Optimistic Ethereum: users, sequencers, and verifiers.
+
+![Network Overview](./assets/network-participants-overview.svg)
+
+### Users
+
+At the heart of the network are users (us!). Users can:
+
+1. Deposit or withdraw tokens by sending transactions to Ethereum mainnet.
```
trianglesphere, line 67: Can also do arbitrary contract interactions with deposits

specs/overview.md
```diff
@@ -0,0 +1,158 @@
...
+
+## Components
+
+![Components](./assets/components.svg)
+
+### L1 Components
+
+- **DepositFeed**: A feed of L2 transactions which originated as smart contract calls in the L1 state.
+	- The `DepositFeed` contract emits `TransactionDeposited` events, which the rollup driver reads in order to process deposits.
+	- Deposits are guaranteed to be reflected in the L2 state within the _sequencing window_.
```
trianglesphere, line 51: Actually the first L2 block that is based off the respective L1 block. Deposits also cannot be reordered.

specs/overview.md
```diff
@@ -0,0 +1,158 @@
...
+
+- Requires no state other than what is easily accessible using L1 and L2 execution engine APIs.
+- Supports sequencers and sequencer consensus.
+- Is resilient to sequencer censorship.
+
+#### Epochs and the Sequencing Window
+
+The rollup chain is subdivided into epochs. There is a 1:1 correspondence between L1 block numbers and epoch numbers. For L1 block number `n`, there is a corresponding rollup epoch `n` which can only be derived after L1 block number `n` is added to the L1 chain. 
+
+Each epoch contains at least one block. Every block in the epoch contains an L1 info transaction which contains contextual about L1 such as the block hash and timestamp. The first block in the epoch also contains all deposits initiated via the `DepositFeed` contract on L1. All subsequent blocks contain batches of transactions submitted directly to the sequencer.
```
trianglesphere, line 123: Not strictly true as specced (but should be) - the result will depend what we do with the timestamp discussion.

specs/overview.md
```diff
@@ -0,0 +1,158 @@
...
+
+The sequencer can create blocks for future epochs, and reference them before confirmation on L1. This is what enables _fast transaction confirmations_ on the sequencer. To prevent censorship, however, the sequencer must batch-submit the blocks belonging to each epoch within a fixed number of L1 blocks. This submission window is called the _sequencing window_. The `epoch`s described above correspond to the first block in each sequencing window. Since transaction batches for a given epoch can be submitted anywhere within the sequencing window, verifiers must search all blocks within the window for transaction batches.
+
+The following diagram describes this relationship, and how L2 blocks are derived from L1 blocks (L1 info transactions have been elided):
```
trianglesphere, line 127: Diagram looks good. Only comment is that with a larger sequencing window, batches can be in any order and more than one block later.

specs/overview.md
```diff
@@ -0,0 +1,158 @@
...
+
+1. Call `engine_forkChoiceUpdatedV1` with the payload attributes object. We'll skip over the details of the fork choice state parameter for now - just know that one of its fields is the L2 chain's `headBlockHash`, and that it is set to the block hash of the tip of the L2 chain. The Engine API returns a payload ID.
+2. Call `engine_getPayloadV1` with the payload ID returned in step 1. The engine API returns a payload object that includes a block hash as one of its fields.
+3. Call `engine_executePayloadV1` with the payload returned in step 2.
+4. Call `engine_forkChoiceUpdatedV1` with the fork choice parameter's `headBlockHash` set to the block hash returned in step 2. The tip of the L2 chain is now the block created in step 1.
+
+The swimlane diagram below visualizes the process:
```
trianglesphere, line 150: Is FSS in the diagram a typo?

trianglesphere (commented):
@mslipper I'm happy to chat about unsafe/safe head. Here's how I would describe it:

Safe Head: A block that can be fully derived from L1. It's L1 Origin is `seq window` blocks behind the L1 head.
Unsafe Head:  A block has not be guaranteed to have it's batch submitted. If the sequencer is live, the unsafe head's L1 Origin will be the L1 head.

Reorgs in L1 will actually roll back both of these heads (even a shallow reorg with a large sequencing window must do a L2 reorg even though it's not likely to cause changes).

specs/introduction.md
```diff
@@ -0,0 +1,127 @@
+# Optimistic Ethereum Introduction
+
+<!-- START doctoc generated TOC please keep comment here to allow auto update -->
+<!-- DON'T EDIT THIS SECTION, INSTEAD RE-RUN doctoc TO UPDATE -->
+**Table of Contents**
+
+- [Foundations](#foundations)
+  - [What is Ethereum scalability?](#what-is-ethereum-scalability)
+  - [What is Optimistic Rollup?](#what-is-optimistic-rollup)
```
tynes, line 9: I think "optimistic rollup" should be qualified with "an", ie "what is an optimistic rollup"?

specs/introduction.md
```diff
@@ -0,0 +1,127 @@
...
+  - [Sequencers](#sequencers)
+  - [Verifiers](#verifiers)
+- [Key Interaction Diagrams](#key-interaction-diagrams)
+  - [Depositing and Sending Transactions](#depositing-and-sending-transactions)
+  - [Withdrawing](#withdrawing)
+- [Next Steps](#next-steps)
+
+<!-- END doctoc generated TOC please keep comment here to allow auto update -->
+
+Optimistic Ethereum is an _EVM equivalent_, _optimistic rollup_ protocol designed to _scale Ethereum_ while remaining maximally compatible with existing Ethereum infrastructure. This document provides an overview of the protocol to provide context for the rest of the specification.
```
tynes, line 24: Optimistic Ethereum or Optimism?

specs/introduction.md
```diff
@@ -0,0 +1,127 @@
...
+
+## Foundations
+
+### What is Ethereum scalability?
+
+Scaling Ethereum means increasing the number of useful transactions the Ethereum network can process. Ethereum's limited resources, specifically bandwidth, computation, and storage, constrain the number of transactions which can be processed on the network. Of the three resources, computation and storage are currently the most significant bottlenecks. These bottlenecks limit the supply of transactions, leading to extremely high fees. Scaling Ethereum and reducing fees can be achieved by increasing the supply of these limited resources.
```
tynes, line 30: Scaling is increasing the supply of gas while holding the hardware constraints constant. Another way to put that is that scalability increases the throughput of the system while not increasing the minimum hardware requirement.

tynes: > Scaling Ethereum and reducing fees can be achieved by increasing the supply of these limited resources.

Scaling ethereum and reducing fees can be achieved by better utilizing bandwidth, computation and storage. Increasing the supply of those 3 isn't really increasing the scalability of the protocol itself

specs/overview.md
norswap: I think the above sentence is correct and fine to leave as-is, since at this stage the idea that L2 blocks belong to epoch derived from L1 blocks hasn't been introduced yet.

specs/overview.md
```diff
@@ -0,0 +1,158 @@
...
+
+L1 attributes deposits are used to register L1 block attributes (number, timestamp, etc.) on L2 via a call to the L1 Attributes Predeploy. They cannot be initiated by users, and are instead added to L2 blocks automatically by the rollup node.
+
+Both deposit types are represented by a custom EIP-2718 transaction type on L2.
+
+### Block Derivation
+
+#### Overview
+
+The rollup chain can be deterministically derived given an L1 Ethereum chain. The fact that the entire rollup chain can be derived based on L1 blocks is _what makes OE a rollup_. This process can be represented as:
```
norswap, line 107: s/OE/Optimism

specs/overview.md
```diff
@@ -0,0 +1,158 @@
...
+
+Optimism's block derivation function is designed such that it:
+
+- Requires no state other than what is easily accessible using L1 and L2 execution engine APIs.
+- Supports sequencers and sequencer consensus.
+- Is resilient to sequencer censorship.
+
+#### Epochs and the Sequencing Window
+
+The rollup chain is subdivided into epochs. There is a 1:1 correspondence between L1 block numbers and epoch numbers. For L1 block number `n`, there is a corresponding rollup epoch `n` which can only be derived after L1 block number `n` is added to the L1 chain.
```
norswap, line 121: Not false, but incomplete: epoch `n` can only be fully derived after block `n` + SEQUENCER_WINDOW_SIZE blocks have been added to the L1 chain.

specs/overview.md
norswap: s/contextual/contextual information

Notes:
- The first block can also contain sequenced transactions!
- Every single block L2 gets the L1 attribute deposit (which is identical for all blocks in the epoch). Not sure I'm a fan of that, but that's how we do it right now!

specs/overview.md
```diff
@@ -0,0 +1,158 @@
...
+
+Each epoch contains at least one block. Every block in the epoch contains an L1 info transaction which contains contextual about L1 such as the block hash and timestamp. The first block in the epoch also contains all deposits initiated via the `DepositFeed` contract on L1. All subsequent blocks contain batches of transactions submitted directly to the sequencer.
+
+The sequencer can create blocks for future epochs, and reference them before confirmation on L1. This is what enables _fast transaction confirmations_ on the sequencer. To prevent censorship, however, the sequencer must batch-submit the blocks belonging to each epoch within a fixed number of L1 blocks. This submission window is called the _sequencing window_. The `epoch`s described above correspond to the first block in each sequencing window. Since transaction batches for a given epoch can be submitted anywhere within the sequencing window, verifiers must search all blocks within the window for transaction batches.
```
norswap, line 125: A bit vague what "future epochs" means here.

The above is true if "future epoch"  == an epoch `n` such that the sequencing window `n-1` isn't finished yet.

It isn't true if "future epoch" == an epoch `n` such that the L1 block `n` has not been create yet.

In general a bit unsure about this paragraph. I think the idea you want to convey here is that to include transactions in a L2 block contained within a given epoch, the sequencer must submit a batch containing the transaction and the epoch number within the sequencing window of the epoch.

specs/overview.md
norswap: I think this will never be my favorite diagram, but it seems correct!

Two notes though:
- Currently epochs will almost always have a fixed size (i.e. time between two L1 blocks / 2, so usually 6 blocks per epoch).
- If we adopt the variable-size epoch proposal (cf. "timestamp"  discussion) then it becomes possible to make bigger epochs and hence delay the first L2 block of an epoch much more. So if you're diagram is supposed to be "time-aligned" then the L1 block and first L2 block of the corresponding epoch won't necessarily line up so well anymore.

specs/overview.md
```diff
@@ -0,0 +1,158 @@
...
+
+A sub-component of the rollup node called the _rollup driver_ is actually responsible for performing block derivation. The rollup driver is essentially an infinite loop that runs the block derivation function. For each epoch, the block derivation function performs the following steps:
+
+1. Downloads deposit and transaction batch data for each block in the sequencing window.
+2. Converts the deposit and transaction batch data into payload attributes for the Engine API.
+3. Submits the payload attributes to the Engine API, where they are converted into blocks and added to the canonical chain.
+
+This process is then repeated with incrementing epohcs until the tip of L1 is reached.
+
+### Engine API
```
norswap, line 141: In the overview for infra provider, you call the execution engine "bedrock geth". I kinda like it, do we want to introduce this terminology here / in the specs? (In addition to "execution engine", which we've used a lot).

specs/overview.md
```diff
@@ -0,0 +1,158 @@
...
+
+The rollup driver doesn't actually create blocks. Instead, it directs the engine API to do so via the Engine API. For each iteration of the block derivation loop described above, the rollup driver will craft a _payload attributes_ object and send it to the Engine API. The Engine API will then convert the payload attributes object into a block, and add it to the chain. The basic sequence the rollup driver is as follows:
```
norswap, line 143: All uses of "engine API" should be "execution engine"  *excepted* "via the Engine API".

norswap (commented):
Lots of comments, but I really like the thrust of this document! It strikes a good balance between being high-level and giving enough details to answer people's questions.

specs/overview.md
mslipper: Yep, fixed.

specs/overview.md
```diff
@@ -0,0 +1,170 @@
...
+  - [L2 Chain Derivation](#l2-chain-derivation)
+  - [Sequencing](#sequencing)
+
+<!-- END doctoc generated TOC please keep comment here to allow auto update -->
+
+This document is a high-level technical overview of the Optimism protocol. It aims to explain how the protocol works in an informal manner, and direct readers to other parts of the specification so that they may learn more. A basic understanding of how optimistic rollups work is a prerequisite to reading this document.
+
+## Components
+
+![Components](./assets/components.svg)
```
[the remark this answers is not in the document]
mslipper: Done.

specs/overview.md
```diff
@@ -0,0 +1,170 @@
...
+
+- A smart contract that stores L2 output roots.
+
+### L2 Components
+
+**Rollup Node**
+
+- A standalone binary that sits between the execution engine and L1.
+- Derives data from L1, and handles L1 reorgs.
+- Validates incoming block data.
```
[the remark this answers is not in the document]
mslipper: I removed this.

specs/overview.md
```diff
@@ -0,0 +1,170 @@
...
+**Output Submitter**
+
+- A background process that submits L2 state roots to the `L2OutputOracle`.
+
+### Types of Rollup Nodes
+
+There are two types of nodes on Optimism: sequencers, and verifiers.
+
+- **Sequencers** are responsible for accepting user-sent and deposit transactions, ordering them into L2 blocks, and submitting batch/output roots to L1.
+- **Verifiers** reconstruct the L2 chain from data on l1.
```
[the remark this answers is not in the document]
mslipper: Removed.

specs/overview.md
```diff
@@ -0,0 +1,170 @@
...
+
+Both node types run the EE and Rollup Node to maintain their view of the rollup state.
```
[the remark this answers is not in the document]
mslipper: Removed.

specs/overview.md
```diff
@@ -0,0 +1,170 @@
...
+
+Sequencers, however, also perform block production and batch/output submission. There is currently only one sequencer. We may support distributed sequencing in the future.
+
+### Transaction/Block Propagation
+
+**Spec links:**
+
+- [Execution Engine](specs/exec-engine.md)
+
+Since the EE uses Geth under the hood, Optimism uses Geth's built-in peer-to-peer network and transaction pool to propagate transactions. The same network can also be used to propagate submitted blocks and support snap-sync.
```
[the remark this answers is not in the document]
mslipper: I can update this once we change the spec.

specs/overview.md
```diff
@@ -0,0 +1,170 @@
...
+
+- [Execution Engine](specs/exec-engine.md)
+
+Since the EE uses Geth under the hood, Optimism uses Geth's built-in peer-to-peer network and transaction pool to propagate transactions. The same network can also be used to propagate submitted blocks and support snap-sync. 
+
+Unsubmitted blocks, however, are propagated using a separate peer-to-peer network of Rollup Nodes. This is optional, however, and is provided as a convnience to lower latency for users.
+
+The below diagram illustrates how the sequencer and verifiers fit together:
+
+![Propagation](./assets/propagation.svg)
```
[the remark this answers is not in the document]
mslipper: I like having the unsafe block propagation arrow be bidirectional because it implies the existence of a peer-to-peer network that gossips data rather than a client-server network where verifiers always read from the sequencer.

specs/overview.md
mslipper: I think we should keep Execution Engine since it's the technically correct term and matches what the rest of the community uses.

specs/overview.md
mslipper: This paragraph is supposed to describe how we get fast confirmations while also preventing censorship by the sequencer. "Future epochs" here means epochs for which the window isn't finished yet. How do you recommend I describe this distinction?

mslipper:
I've updated the diagrams as per @norswap's request. I also removed the placeholder unsafe heads diagram.

@protolambda - I think the diagrams should be correct now. The arrows follow the flow of data, so it's clear that verifiers read from the L1 smart contracts rather than a submitter process.

specs/introduction.md
```diff
@@ -0,0 +1,159 @@
...
+
+Links to components mentioned in this diagram:
+
+- [L2 Output Oracle](./proposals.md#l2-output-oracle-smart-contract)
+
+## Next Steps
+
+This is a choose your own adventure. Are you interested in how a verifier works under the hood? Maybe you want to dive
+deep into the bit flippin' Fraud Proof VM? All key components have been linked at least once in this doc, so you should
+now have the context you need to dive in deeper. [The world is yours](https://www.youtube.com/watch?v=e5PnuIRnJW8)!
```
protolambda, line 159: Hello world :joy:

protolambda (commented):
Looks good, but the sequencing section is not done yet. Do we move that to a separate PR, or does it block this one?

Also, I made these diagrams for a recent presentation:

bedrock-Copy of Page-1

bedrock(2)


Something like that, maybe smaller / without the images, could help explain the node relations and symmetry with L1.

specs/overview.md
norswap: I'll post my suggested rewrite for this section below, but explaining changes:

> All subsequent blocks contain batches of transactions submitted directly to the sequencer.

Bit confusing to use the word *batch* outside of batch submission.

> All subsequent blocks contain batches of transactions submitted directly to the sequencer.

This seems to imply the first block cannot contain sequenced transactions, which it can.

> To prevent censorship, however, the sequencer must batch-submit the
blocks belonging to each epoch within a fixed number of L1 blocks.

I think it's pretty confusing in what batch submission has to do with censorship prevention here.

Otherwise I think this part is tricky because batch submission isn't really detailed before.

---

```
The rollup chain is subdivided into epochs. There is a 1:1 correspondence between L1 block numbers and epoch numbers.
For L1 block number `n`, there is a corresponding rollup epoch `n` which can only be derived a "sequencing windows" has
passed, i.e. after L1 block number `n + SEQUENCING_WINDOW_SIZE` is added to the L1 chain.

Each epoch contains at least one block. Every block in the epoch contains an L1 info transaction which contains
contextual information about L1 such as the block hash and timestamp. The first block in the epoch also contains all
deposits initiated via the `DepositFeed` contract on L1. All L2 blocks can also contain *sequenced transactions*, i.e.
transactions submitted directly to the sequencer.

Whenever the sequencer creates a new L2 block for a given epoch, it must submit it to L1 as part of a *batch*, within
the epoch's sequencing window (i.e. the batch must land before L1 block `n + SEQUENCING_WINDOW_SIZE`). These batches are
(along with the `TransactionDeposited` L1 events) what allows the derivation of the L2 chain from the L1 chain.

The sequencer does not need for a L2 block to be batch-submitted to L1 in order to build on top of it. In fact, batches
typically contain multiple L2 blocks worth of sequenced transaction. This is what enables
_fast transaction confirmations_ on the sequencer.

Since transaction batches for a given epoch can be submitted anywhere within the sequencing window, verifiers must
search all blocks within the window for transaction batches. This protects against the uncertainty of transaction
inclusion of L1. This uncertainty is also why we need the sequencing window in the first place: otherwise the sequencer
could retroactively add blocks to an old epoch, and validators wouldn't know when they can finalize an epoch.

The sequencing window also prevents censorship by the sequencer: deposits made on a given L1 block will be included in
the L2 chain at worst after `SEQUENCING_WINDOW_SIZE` L1 blocks have passed.
```

specs/introduction.md
```diff
@@ -0,0 +1,159 @@
...
+
+### Sequencers
+
+The sequencer is the primary block producer. There may be one sequencer **or** many using a consensus protocol. For
+1.0.0, there is just one sequencer.  In general, specifications may use "the sequencer" to be a stand-in term for the
+consensus protocol operated by multiple sequencers.
+
+The sequencer:
+
+1. Accepts user off-chain transactions via the mempool
```
norswap, line 100: I'd nix "via the mempool" for now.

mslipper:
@protolambda I say we document sequencing as a separate PR.

specs/overview.md
mslipper: I applied your rewrite.

specs/overview.md
```diff
@@ -0,0 +1,205 @@
...
+
+The swimlane diagram below visualizes the process:
+
+![Engine API](./assets/engine.svg)
+
+### Sequencing
+
+> **TODO:** Flesh out this section.
+
+![Sequencing 1](./assets/safe-unsafe-heads.svg)
```
trianglesphere, line 205: This is a dead link

trianglesphere (approved):
@mslipper there's one dead link that needs to be fixed, but once that's fixed, it's good to go.

specs/overview.md
mslipper: Derp, I meant to remove that section altogether. Fixed.

PR #299, opened 2022-03-28
opnode: Extend sync algorithm to SafeHead

trianglesphere:
**Description**
The sync start algorithm has been modified to find both a SafeL2Head
and UnsafeL2Head. The UnsafeL2Head is the first L2 block that has
an L1 origin that is canonical in the L1 chain.

The FindL2Heads function first finds the UnsafeL2Head, and then continues
to walk back the L2 chain until it finds and L2 block that can be completely
derived from the sequence window that has not been changed in the
reorg (i.e. the last block of the window was the reorg base).

The sync algorithm previously only found the UnsafeHead.

**Additional context**
This PR also makes some small changes to timestamp handling the state loop.


protolambda (commented):
Can you update the spec to explain the changes in the implementation?

Generally I think of "safe head" in L2 purely based on inclusion of the batch in L1, not based on the L1 origin information in the L2 block. I.e. a L2 block can go from unsafe to safe without the L2 changing at all, if L1 includes it. Tracking the safe-head could be completely separate from the sync.

opnode/rollup/sync/start.go
```diff
@@ -59,45 +58,72 @@ var TooDeepReorgErr = errors.New("reorg is too deep")
...
-		return eth.L2BlockRef{}, fmt.Errorf("failed to fetch L2 head: %w", err)
+		return false, err
 	}
+	return canonical.Hash == block.Hash, nil
+}
+
+// FindL2Heads walks back from the supplied L2 blocks and finds the unsafe and safe L2 heads.
+// Unsafe Head: The first L2 block who's L1 Origin is canonical (determined via block by number).
+// Safe Head: Walk back 1 sequence window of epochs from the Unsafe Head.
+// Returns: `unsafeHead`, `safeHead`, `error`
```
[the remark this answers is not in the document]
trianglesphere: done

trianglesphere:
> Can you update the spec to explain the changes in the implementation?

Will do.


> Generally I think of "safe head" in L2 purely based on inclusion of the batch in L1, not based on the L1 origin information in the L2 block. I.e. a L2 block can go from unsafe to safe without the L2 changing at all, if L1 includes it. Tracking the safe-head could be completely separate from the sync.

I think that the definition is still there. I think that the purpose of the sync function has slightly changed: it's essentially only used on startup and during L1 reorgs.

The L1 Origin information is serving as an interesting trick to validate the end of the batch. Take an L2 block. When we derive it we have and L1 window. We reference the start of the window with the L1 Origin, but we only reference the end of the window by block number. If there is a L1 reorg that occurs after the L1 Origin, but before the end of the L1 window, we must recompute the L2 block as the sequence window has changed. One approach would to cary metadata about the end of the sequence window (and I think we should do this at some point), but we cannot commit to the end of the sequence window (as the sequencer is unable to do so). Finding the unsafe head (based on L1 origin) is one approach to finding the "reorg base" of L1. By finding an L1 block that we know has not changed, we can be sure that any previously computed sequence window that ends before or at that L1 block will not change.

Note: I had another approach which would try to find the reorg base in L1 and supply it to this algorithm and use that as the `isCanonical` check.

trianglesphere:
@protolambda I added specs + addressed comments

specs/rollup-node.md
```diff
@@ -373,43 +373,38 @@ The [previous section on L2 chain derivation][l2-chain-derivation] assumes linea
...
-  requires a re-genesis of the L2 chain to fix (i.e. creating a new genesis configuration) (\*)
-- Otherwise, if either `currentL1` does not exist, or `currentL1 != refL1`, set `refL2` to `parentL2` and restart this
-  algorithm from step 2.
-  - Note: if `currentL1` does not exist, it means we are in a re-org to a shorter L1 chain.
-  - Note: as an optimization, we can cache `currentL1` and reuse it as the next value of `nextRefL1` to avoid an
-        extra lookup.
+ancestor, and can re-derive the L2 chain from that L1 block and onward.
+
+The rollup node maintains two heads of the L2 Chain: the unsafe head (often called head) and the safe head.
+Each L2 block has an L1 Origin block that it references in the L1 Info deposit transaction.
```
norswap, line 388: I think it's called "L1 Attributes" deposit transaction (check glossary).

specs/rollup-node.md
```diff
@@ -373,43 +373,38 @@ The [previous section on L2 chain derivation][l2-chain-derivation] assumes linea
...
+The unsafe head is the L2 block with the highest number who's L1 Origin block is canonical in the new L1 chain.
+The safe head is the L2 block that was derived from a sequencing window that did not change with the reorg.
```
norswap, line 390: s/who's/whose

norswap: Proto raised this concern, but does this mean that validators never accept blocks from the sequencer if they haven't seen the L1 block it builds on?

norswap: >  The safe head is the L2 block that was derived from a sequencing window that did not change with the reorg.

Not a fan of this formulation. I would say something like:
> The safe head is the the last L2 block of the last epoch whose sequencing window is complete (i.e. the epoch with number `L1Head.number - SEQUENCING_WINDOW_SIZE`).

specs/rollup-node.md
```diff
@@ -373,43 +373,38 @@ The [previous section on L2 chain derivation][l2-chain-derivation] assumes linea
...
+is a full sequencing window behind the L1 Origin number of the unsafe head.
+
+The purpose of this is to ensure that if the sequencing window for a L2 block has changed since it was derived, that
+the L2 block is re-derived. The start of the sequencing window for a L2 block is committed to in the L1 Origin; however
+the end of the window is based on a block number (note that it cannot be done any other way). As such we must find the
+highest L1 block that is common to both L1 chains (the reorg base) and the safe head is the L2 block who's sequencing
+window ends at the reorg base. Any L2 block after that will have a different sequencing window.
+
+Walking the L2 chain until we find a canonical L1 block as the L1 origin is a straightforward way to determining the
+reorg base and it also serves as an implicit commitment to the end of the sequencing window for any L2 block.
```
norswap, line 405: I find this paragraph from "The start of the sequencing window" and onwards to be very confusing.

> The start of the sequencing window for a L2 block is committed to in the L1 Origin

I guess you mean that the L1 origin IS the first block of the sequencing window for the epoch with the same number?

> the end of the window is based on a block number

Do you just mean that the end of the window is L1Origin + SEQUENCING_WINDOW_SIZE?

> (note that it cannot be done any other way)

If you're not going to explain why that's just unnecessary teasing :D

norswap: > As such we must find the highest L1 block that is common to both L1 chains (the reorg base)

I see what you mean but "both L1 chains" is a bit weird. And I got confused because of the sentence structure: "both L1 chains AND ...."

> and the safe head is the L2 block who's sequencing window ends at the reorg base.

Drop the "and" and make this a new sentence:

> The safe head is the last L2 block of the epoch whose sequencing window ends at the reorg base.

Note that this is sort of already said above.

> Walking the L2 chain until we find a canonical L1 block as the L1 origin is a straightforward way to determining the
reorg base and it also serves as an implicit commitment to the end of the sequencing window for any L2 block.

This repeats what has been said a bit before, but more confusingly, so I would just drop it.

opnode/rollup/sync/start.go
```diff
@@ -59,45 +58,71 @@ var TooDeepReorgErr = errors.New("reorg is too deep")
...
+	var prevL1OriginHash common.Hash
+	// Walk L2 chain from start until L1Origin matches l1Base. Bail out early of when at genesis.
+	for n := start; ; {
+		// Check if l1Origin is canonical when we get to a new epoch
+		if prevL1OriginHash != n.L1Origin.Hash {
+			if ok, err := isCanonical(ctx, l1, n.L1Origin); err != nil {
+				return eth.L2BlockRef{}, eth.L2BlockRef{}, err
+			} else if ok {
+				unsafeHead = n
+				break
```
protolambda, line 84: Scenario:
- L2 chain is `abcde`
- L1 chain is `ABCDE`
- `start` is defined as the absolute L2 tip: `e`, an unsafe block derived from `D` as origin and included in `E`, already synced
- the local view of the L1 chain is only at `C`: so retrieval of `D` and `E` return `ethereum.NotFound`

In this case I would want to see `unsafeHead = e`, and `safeHead = c`.

But if we do what this PR currently does, then the `isCanonical` fails because `n.L1Origin = C` and `isCanonical` fails or returns false.

Instead, I think we should just remember the `l2UnsafeHead`, and only return something different if we find that the tip of L1 is not part of the any of the L1-origins that we traversed when walking back the L2 chain to the common L1 point.

This way we don't rewind the L2 chain suddenly:
- on restarts and such
- when we receive L1 blocks with a delay and don't do the extension-shortcut
- when we call the `findL2Heads` on e.g. receiving a new L2 block from an unsafe channel that is not a simple extension, but that doesn't reorg all unsafe blocks either.

Ideally we:
- call this function at any time to get the true sync starting point
- can stay on the latest L2 chain, even with reorgs, even if e.g. a L1 node is temporarily offline.

specs/rollup-node.md
trianglesphere: Yep. L1 Origin got transferred over from names I chose in the code

specs/rollup-node.md
trianglesphere: > I guess you mean that the L1 origin IS the first block of the sequencing window for the epoch with the same number?

yes

> end of the window is L1Origin + SEQUENCING_WINDOW_SIZE

(and that it is by number, not by hash while the start is by hash)

> If you're not going to explain why that's just unnecessary teasing :D

It's impossible for the sequencer to know the end of the sequencing window by hash as it's a block in the future (and they must include a transaction that changes the hash of the block)

trianglesphere: Will go through and clean this up

norswap:
Would love to see a summary of what the "new" system is once this is considered ready-to-go. And also if any terms / definitions / assumptions we used previously are changing, I'd like to see that mentionned as well.

I feel in general we should get to more agreement on design (doesnt *not* need to be a complete / formal / high-polish spec) before we start implementing. Just about like getting on the same page about the design of what we're building. I've been struggling to understand some of the recent development from short PR descriptions and diffs to a (rapidly evolving) codebase. Something to think about after the audit!

opnode/rollup/sync/start.go
trianglesphere: I went through and updated this with a new test for it.

> on restarts and such

I think that we need to reorg the safe L2 chain on restarts because we don't track the end of the sequence window in way that can be checked at a later point.

> when we receive L1 blocks with a delay and don't do the extension-shortcut

This is case really only matters when we miss an L1 block. What happens is that the safe head / unsafe head should no-op change.

> when we call the findL2Heads on e.g. receiving a new L2 block from an unsafe channel that is not a simple extension, but that doesn't reorg all unsafe blocks either.

I'm not sure the use case here

trianglesphere:
A brief summary of the changes would be as follows:
- I realized that even a short L1 reorg requires us to reorg the safe L2 chain because of the issue with the sequencing window. I think that it unlikely to change the L2 safe chain, but until we have more information about how the L2 block was constructed, it must be done
- I actually very rarely call the sync start algorithm (only at startup and during L1 reorgs). Otherwise we keep building off of our local view
- Note that verifiers actually should run the same epoch verification logic rather than re-executing L2 blocks on an L1 reorg (keep L2 unsafe head the same during a short L1 reorg, but roll back the safe L2 head and then quickly verify that there are no changes).

specs/rollup-node.md
```diff
@@ -373,43 +373,41 @@ The [previous section on L2 chain derivation][l2-chain-derivation] assumes linea
...
+
+The rollup node maintains two heads of the L2 Chain: the unsafe head (often called head) and the safe head.
+Each L2 block has an L1 Attributes block that it references in the L1 Info deposit transaction.
+The unsafe head is the head of the L2 chain. It's L1 Attributes should be canonical or potentially extending the
+canonical chain (if the rollup node has not yet seen the L1 block that it is based upon).
+The safe head is the the last L2 block of the last epoch whose sequencing window is complete
+(i.e. the epoch with number `L1Head.number` - `SEQUENCING_WINDOW_SIZE`).
+
+Determining the unsafe head in the case of an L1 reorg is easy. Start with the current L2 Head and walk the L1 chain
+until the L1 Attributes of an L2 block is canonical or not found (and ahead of the current L1 head).
```
protolambda, line 395: Spec writer note: never write "is easy", there is always a chance (small or not) of an oversight in the spec itself.

> until the L1 Attributes of an L2 block is canonical or not found (and ahead of the current L1 head).

Walking back until the L1 attributes are not-found is wrong, the "ahead of the current L1 head" is critical in that case. E.g. you could have a L1-attributes that cannot be found, that builds on top of a L1 chain that is not canonical.

opnode/rollup/sync/start.go
```diff
@@ -59,45 +60,94 @@ var TooDeepReorgErr = errors.New("reorg is too deep")
...
+
+	// Walk L2 chain from start until L1Origin matches l1Base. Bail out early of when at genesis.
+	for n := start; ; {
+		// Check if l1Origin is canonical when we get to a new epoch
+		if prevL1OriginHash != n.L1Origin.Hash {
+			if ok, err := isCanonical(ctx, l1, n.L1Origin); err != nil {
+				return eth.L2BlockRef{}, eth.L2BlockRef{}, err
+			} else if ok {
+				unsafeHead = n
+				break
```
protolambda, line 99: The `unsafeHead` should only change when the L1 information of that block is *not* canonical. It should stay the same as long as the L1 info is not-found. Changing it here when the L1 info is canonical could result in unnecessary reorgs.

specs/rollup-node.md
```diff
@@ -373,43 +373,41 @@ The [previous section on L2 chain derivation][l2-chain-derivation] assumes linea
...
+
+To determine the safe head, the straightforward approach is to walk back from the unsafe head until the L1 Attributes
+number is a full sequencing window behind the L1 Attributes number of the unsafe head.
```
protolambda, line 398: The L2 chain can be a full sequencing window ahead of the L1 head if the L1 source is stuck in sync, with multiple different L1 attributes until we finally find one that we do have on L1 as *head*.

protolambda (changes requested):
It still looks like the unsafe head is determined incorrectly in some cases, and/or your definition of "unsafeL2Head" is different than mine.

I expect "unsafeL2Head" to be the very end of the L2 chain, and can use any L1 information, as long as the chain strictly extends what could be derived from the canonical L1 chain head. 

Tracking the first L2 block with a canonical L1 block does not seem useful to me, we should be tracking the L2 block that we can build on regardless of lagging L1 information.

opnode/rollup/sync/start.go
trianglesphere: This is why I don't like named returns in go. See line 130 for how this returns correctly if the start is ahead of L1

specs/rollup-node.md
```diff
@@ -373,43 +373,44 @@ The [previous section on L2 chain derivation][l2-chain-derivation] assumes linea
...
+(i.e. the epoch with number `L1Head.number` - `SEQUENCING_WINDOW_SIZE`).
+
+To find the unsafe block, start with the current L2 Head and walk the L1 chain
+until the L1 Attributes of an L2 block is canonical or not found (and ahead of the current L1 head).
+
+To find the "latest block", start with the unsafe block and walk back until the L1 Attributes of the L2 block
+is canonical in the L1 chain.
+
+To determine the safe head, walk back from the latest block until the L1 Attributes
+number is a full sequencing window behind the L1 Attributes number of the latest block.
```
protolambda, line 401: We can improve this a little by doing it like this:
1. Set "unsafe head" to equal the l2 head we retrieved, just as default
2. Set "latest block" to equal the l2 head we retrieved, also just as default
3. walk back L2, and stop until `block.l1Origin` is found AND canonical, and update "latest block" to this `block`. And don't override "unsafe head" if it's not found, but do override it when `block.l1Origin` does not match the canonical L1 block at that height.

This:
- Ensures there values are set correctly, even if there's no walking at all (i.e. L2 chain is at genesis)
- Catches a case where the unsafe-head can be determined to get reorged out, even though we don't have any L1 data for that l2 block height, by looking for a reorg between the unsafe head and the latest head.

protolambda (commented):
Looks good, thank you for all the iterations. A few more comments, but generally looks good now

opnode/rollup/driver/state.go
```diff
@@ -182,43 +203,40 @@ func (s *state) loop() {
...
-					s.log.Error("Could not get fetch L2 head when trying to handle a re-org", "err", err)
-					continue
-				}
-				nextL2Head, err := sync.FindSafeL2Head(ctx, l2Head.ID(), s.l1, s.l2, &s.Config.Genesis)
-				if err != nil {
-					s.log.Error("Could not get new safe L2 head when trying to handle a re-org", "err", err)
+					s.log.Error("Could not get new unsafe L2 head when trying to handle a re-org", "err", err)
 					continue
 				}
+				// TODO: Fork choice update
```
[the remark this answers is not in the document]
trianglesphere: It doesn't break, but we need to extend the API.

opnode/test/setup.go
```diff
@@ -151,8 +151,9 @@ func initializeGenesis(cfg *systemConfig) {
 		Difficulty: common.Big1,
 		GasLimit:   5000000,
 		Nonce:      4660,
-		Timestamp:  genesisTimestamp,
-		BaseFee:    big.NewInt(7),
+		// Set the L2 timestamps to be before the L1 timestamps. This is to match the current implementation of timestamps.
```
trianglesphere, line 154:
```suggestion
		// Set the L2 timestamps to be before the timestamp of the L1 genesis. This is to match the current implementation of timestamps.
```

protolambda: Think this is worth cleaning up later. The genesis L2 block is "derived" from the anchor L1 block I'd say, and equal genesis timestamps should be possible, just like it should be possible to match the timestamp of the L1 block you start a new epoch of L2 blocks with. Not a big deal in this test though, and can be separated into later timestamp work

protolambda (approved):
Looks good, nice work!

PR #301, opened 2022-03-29
opnode: refactor RPC server / config, prototype batch RPC endpoint (work in progress)

protolambda:
This PR implements the RPC server for batch-submitter.

Changes:
- Clean up output submitter test mock, use mocking lib
- Move WithdrawalContractAddress to the rollup node chain config, where the other system-critical addresses are defined. This address should never suddenly change, it's not a CLI parameter.
- Update e2e test with config changes
- Get rid of rpc l2 client bindings that were used in output submitter. Instead introduce a read-only l2 source, and move the l2 logic into the l2 package. Next goal is to unify better with the other l2 client bindings, to de-duplicate the readonly methods with the read/write bindings, and cache things.
- Extend l2 client bindings with block ref getters and block getter
- Method to convert L2 block to batch

### RPC design for batch submitter

Based on discussion with @cfromknecht :

Light-weight batch-submitter, focused on managing its own transactions. The only stateful thing it has is that it remembers the *last* (not even each individual) block hash of each of the batches it submitted.
We can host the state in some reliable/redundant/fast external database in production, but should have a default for testing that just stores this state locally (literally only a list of recent block hashes, about ~20 KB for a worst-case reorg of 20 min of L2 data that is all sequenced but never submitter/confirmed).

The batch-submitter can then ask the rollup-node for batches, and give any arbitrary list of block-hashes. The first common hash with the canonical L2 chain will be the resumption point (i.e. all L2 blocks with higher block height will be included in the batches). The rollup-node returns an encoded bundle of batches, possibly including compression etc. The result should be built to be smaller than the requested size .

TODOs:
- [ ] Fix batch rpc endpoint to respect parameter of desired max size
- [ ] More sanity limits
- [ ] Testing
- [ ] Clean up the L2 source. It should have caching like the new L1 downloader. And unify more of the RPC methods with the engine API source, and type things better (e.g. no string "blockTag")

We can remove the prototype batch-submitter from driver-loop once we integrate the external batch-submitter in the E2E test.
Similarly I think we should integrate the withdrawals process (just like we do deposits), and the output-root submitter.

### DIFF
--- a/opnode/test/system_test.go
+++ b/opnode/test/system_test.go
@@ -172,9 +172,14 @@ func TestSystemE2E(t *testing.T) {
 			SeqWindowSize:        2,
 			L1ChainID:            big.NewInt(900),
 			// TODO pick defaults
-			FeeRecipientAddress: common.Address{0xff, 0x01},
-			BatchInboxAddress:   common.Address{0xff, 0x02},
-			BatchSenderAddress:  submitterAddress,
+			FeeRecipientAddress:       common.Address{0xff, 0x01},
+			BatchInboxAddress:         common.Address{0xff, 0x02},
+			BatchSenderAddress:        submitterAddress,
+			WithdrawalContractAddress: common.Address{0xff, 0x03}, // TODO: pre-deploy withdrawal contract
+		},
+		RPC: rollupNode.RPCConfig{
+			ListenAddr: "localhost",
+			ListenPort: cfg.l2Sequencer.nodeConfig.HTTPPort,
 		},
 	}
 	node, err := rollupNode.New(context.Background(), nodeCfg, testlog.Logger(t, log.LvlError), "")
@@ -188,7 +193,7 @@ func TestSystemE2E(t *testing.T) {
 	sequenceCfg := &rollupNode.Config{
 		L1NodeAddr:    endpoint(cfg.l1.nodeConfig),
 		L2EngineAddrs: []string{endpoint(cfg.l2Sequencer.nodeConfig)},
-		L2NodeAddr:    endpoint(cfg.l2Verifier.nodeConfig),
+		L2NodeAddr:    endpoint(cfg.l2Sequencer.nodeConfig),
 		Rollup: rollup.Config{
 			Genesis: rollup.Genesis{
 				L1:     l1GenesisID,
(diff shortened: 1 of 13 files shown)


opnode/test/system_test.go
```diff
@@ -172,9 +172,14 @@ func TestSystemE2E(t *testing.T) {
...
-			BatchInboxAddress:   common.Address{0xff, 0x02},
-			BatchSenderAddress:  submitterAddress,
+			FeeRecipientAddress:       common.Address{0xff, 0x01},
+			BatchInboxAddress:         common.Address{0xff, 0x02},
+			BatchSenderAddress:        submitterAddress,
+			WithdrawalContractAddress: common.Address{0xff, 0x03}, // TODO: pre-deploy withdrawal contract
+		},
+		RPC: rollupNode.RPCConfig{
+			ListenAddr: "localhost",
+			ListenPort: cfg.l2Sequencer.nodeConfig.HTTPPort,
```
cfromknecht, line 182: collides with verifier geth port

https://github.com/ethereum-optimism/optimistic-specs/pull/312/commits/0cde57573942fd8355bcefd66c7be287d0521fa0

opnode/test/system_test.go
```diff
@@ -200,12 +205,17 @@ func TestSystemE2E(t *testing.T) {
...
+			FeeRecipientAddress:       common.Address{0xff, 0x01},
+			BatchInboxAddress:         common.Address{0xff, 0x02},
+			BatchSenderAddress:        submitterAddress,
+			WithdrawalContractAddress: common.Address{0xff, 0x03}, // TODO: pre-deploy withdrawal contract
 		},
 		Sequencer:        true,
 		SubmitterPrivKey: bssPrivKey,
+		RPC: rollupNode.RPCConfig{
+			ListenAddr: "localhost",
+			ListenPort: cfg.l2Sequencer.nodeConfig.HTTPPort,
```
cfromknecht, line 217: collides with sequencer geth port

https://github.com/ethereum-optimism/optimistic-specs/pull/312/commits/0cde57573942fd8355bcefd66c7be287d0521fa0

cfromknecht (commented):
Overall changes LGTM, I was able to get everything hooked up in #312. These are mostly just minor things I found when trying to do the integration

PR #302, opened 2022-03-29
Clean up L1Block test file

maurelian:
Assorted minor improvements to forge tests. 

1. Some cosmetics changes in test files. 
2. Added gas benchmarking and activated solc optimizer


packages/contracts/.gas-snapshot
```diff
@@ -0,0 +1,16 @@
...
+testCannot_appendFutureTimestamp() (gas: 4220)
+testCannot_appendOutputIfNotSequencer() (gas: 3491)
+testCannot_appendUnexpectedTimestamp() (gas: 4435)
+testCannot_computePreHistoricalL2BlockNumber() (gas: 2182)
+test_appendingAnotherOutput() (gas: 27894)
+test_computeL2BlockNumber() (gas: 5105)
+test_getL2Output() (gas: 4621)
+test_latestBlockTimestamp() (gas: 1169)
+test_nextTimestamp() (gas: 1572)
+test_constructor() (gas: 6912)
```
maurelian, line 16: This is a forge file that we want checked in to see when the cost of a given test changes.

packages/contracts/contracts/test/L1Block.t.sol
```diff
@@ -1,22 +1,20 @@
+//SPDX-License-Identifier: MIT
 pragma solidity 0.8.10;
 
 import { DSTest } from "../../lib/ds-test/src/test.sol";
+import { Vm } from "../../lib/forge-std/src/Vm.sol";
```
maurelian, line 5: This change matches the boiler plate used in https://github.com/ethereum-optimism/optimistic-specs/blob/main/packages/contracts/contracts/test/L2OutputOracle.t.sol

packages/contracts/package.json
```diff
@@ -8,12 +8,13 @@
     "hardhat": "^2.7.1"
   },
   "scripts": {
-    "build:forge": "forge build --root . -c contracts --out forge-artifacts",
```
maurelian, line 11: have to use this flag because forge doesn't seem to actually recognize the foundry.toml file. I think because it's confused by the dir structure of the repo.

packages/contracts/package.json
```diff
@@ -8,12 +8,13 @@
...
+    "build:forge": "forge build --optimize --root . -c contracts --out forge-artifacts",
```
maurelian, line 11: have to use these flags because forge doesn't seem to actually recognize the foundry.toml file. I think because it's confused by the dir structure of the repo.

Inphi: seems to work for me without adding the flags (using the latest version of forge):
```bash
/o/o/p/contracts > forge config | rg '(src|out)'
extra_output = []
extra_output_files = []
out = 'forge-artifacts'
    'ds-test/=lib/ds-test/src/',
    'forge-std/=lib/forge-std/src/',
src = 'contracts'
```
Maybe it's an issue with an older version of forge.

maurelian: LOL, turned out I'd installed a very old version of foundry using cargo, which was taking precedence over the foundryup version in my path. We can get rid of all this stuff now.

Inphi: ouch. foundryup should detect that.

Inphi (approved):

PR #303, opened 2022-03-29
l2os: Submit l2 output root via opnode

Inphi:
Using the opnode's `optimism_outputAtBlock` RPC, we can submit l2 output roots.

### DIFF
--- a/l2os/drivers/l2output/driver.go
+++ b/l2os/drivers/l2output/driver.go
@@ -8,24 +8,29 @@ import (
 	"strings"
 
 	"github.com/ethereum-optimism/optimistic-specs/l2os/bindings/l2oo"
+	"github.com/ethereum-optimism/optimistic-specs/opnode/l2"
 	"github.com/ethereum/go-ethereum/accounts/abi"
 	"github.com/ethereum/go-ethereum/accounts/abi/bind"
 	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/common/hexutil"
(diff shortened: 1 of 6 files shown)


Inphi:
> Some dial is failing in e2e, please check the endpoint configurations

@protolambda Tests should be good now.

l2os/drivers/l2output/driver.go
```diff
@@ -241,3 +243,17 @@ func (d *Driver) SendTransaction(
 
 	return d.cfg.L1Client.SendTransaction(ctx, tx)
 }
+
+func (d *Driver) outputRootAtBlock(ctx context.Context, blockNum *big.Int) (l2.Bytes32, error) {
```
cfromknecht, line 247: small nit: perhaps it's worth exporting this method (and passing in `RollupClient`) so that we can use this directly in the e2e tests?

down the road, and depending on how many rpcs we end up with, it could make sense to wrap the `rpc.Client` with a dedicated struct that exposes a method for each endpoint

l2os/flags/flags.go
```diff
@@ -25,6 +25,12 @@ var (
 		Required: true,
 		EnvVar:   "L2_ETH_RPC",
 	}
+	RollupRpc = cli.StringFlag{
```
cfromknecht, line 28: nit: add `Flag` suffix to variable name

cfromknecht (commented):
@Inphi changes look good to me, just a few nits 🙌

protolambda (approved):
LGTM

PR #304, opened 2022-03-30
Migrate DepositFeed spec to forge test contract

Inphi:
This is nice because:
* it's consistent with the existing contract testing
* smaller test spec - easier to read and maintain
* no need for the DummyContract to simulate contract calls


packages/contracts/contracts/test/DepositFeed.t.sol
```diff
@@ -0,0 +1,126 @@
...
+
+    // Test: depositTransaction should increase its eth balance when an EOA deposits a contract creation with ETH
+    function test_depositTransaction_withEthValueAndEOAContractCreation() external {
+        // EOA emulation
+        vm.prank(address(this), address(this));
+
+        vm.expectEmit(true, true, false, true);
+        emit TransactionDeposited(address(this), ZERO_ADDRESS, NON_ZERO_VALUE, ZERO_VALUE, NON_ZERO_GASLIMIT, true, '0x');
+
+        uint256 balanceBefore = address(df).balance;
```
maurelian, line 110: Since you're using the `setup()` function, which is basically like `beforeEach()`, I think that this will always be zero, so the line could be removed and would clean up the assertion as well (same for subsequent test).

maurelian (changes requested):
This is great! Really nice to get rid of that Dummy contract, and speeds things up a lot. 

In order to approve I have two requests: 
1. See if you can remove the `balanceBefore` query per my inline comment. 
2. Update the package.json scripts so that the hardhat tests don't also run (I'd keep `test:hh` for now, but change `test` so that it ony runs `test:forge`.

maurelian (approved):

PR #307, opened 2022-03-30
Add bytecode

maurelian:
These changes cause an invalid genesis file error?

image


norswap:
@maurelian can we call the contract Withdrawer and not Withdrawor?

ops/Dockerfile.l2
```diff
@@ -1,15 +1,16 @@
...
 
 RUN apk add --no-cache jq
 
 COPY entrypoint.sh /entrypoint.sh
 COPY genesis-l2.json /genesis.json
 
-RUN cat /genesis.json | jq ". | .alloc.\"4242424242424242424242424242424242424242\".code = \"$L1_BLOCK_INFO_BYTECODE\"" | tee /genesis-with-bytecode.json && \
```
Inphi, line 10: Need to also update the corresponding address in the genesis file - 
https://github.com/ethereum-optimism/optimistic-specs/blob/6697542cefd6155e2a7a3a153d10f3d0bc9bba9e/ops/genesis-l2.json#L797-L800

maurelian: Right, I have a fix for that locally actually.

maurelian: This can be closed actually, and I can't figure out how to do so from my phone

Issue #308, opened 2022-03-30
Subcomponent lifecycle managements


trianglesphere:
We have several sub-components that exist as go-routines / separate threads. We need to close them on application end, but right now it's not clear which component closes what.

In addition, we don't properly wait for things to be closed:
- https://github.com/ethereum-optimism/optimistic-specs/pull/275#discussion_r839034686
- https://github.com/ethereum-optimism/optimistic-specs/blob/95834ed431755f868cd93f016f420436fda38273/opnode/rollup/driver/state.go#L150 (need to wait for bss to be shut)

PR #309, opened 2022-03-31
opnode: Reconcile epochs and handle reorgs

trianglesphere:
**Description**

This PR does two things:
1. Better handle L2 reorgs
2. Induces L2 reorgs when then derived from L1 chain does not match the sequencer chain.

Both verifiers and sequencers run a derivation process and either insert epochs or verify that the attributes derived from L1 match the L2 block at that height. If there is a mismatch, L2 is reorg'd.

**Miscellaneous Changes**
- Switch SeqWindow to uint64 in FindL2Heads function
- Mute the `rollup/driver` tests
- Splits up state loop into sub-steps to be more readable.

**Testing**
I modified the batch submitted to see that reorgs where being handled correctly (see `jg/harden_seq_v3_archive`)
I did notice that there was an "Impossible reorg" error in the e2e test. This is an error reported by the execution engine and it seems to occur when fork choice updated is called twice with the same block.

The e2e test is passing despite the fact that the geth node logic is incorrect. I think what is happening is that the block that is produced is the same as what was there (because the attributes are in fact the same despite it reporting that they are not the same - we have bad info from L2 geth). I then think this is the root cause of the "Impossible reorg" error.

I have not done a deep reorg of L1 and that should be done (to test that verifiers also handle reorgs).
A new e2e test will also have to be designed for this - it involves modifying the batch submitter and/or the L1 chain.

**TODO**

- [x] Update this to the right version of op-geth (tests should not have passed)
- [x] Understand why tests now pass. I needed to update geth to return the mint amount correctly.
- [ ] Testing plan
- [x] Fix #299 and merge it (did fix + rebase)


opnode/rollup/driver/state.go
```diff
@@ -86,51 +114,160 @@ func (s *state) l1WindowBufEnd() eth.BlockID {
...
+		return eth.L1BlockRef{}, err
+	}
+	// State update
+	s.l2Head = newUnsafeL2Head
+	s.log.Info("Sequenced new l2 block", "l2Head", s.l2Head, "l1Origin", s.l2Head.L1Origin)
+	//Submit batch
+	go func() {
+		_, err := s.bss.Submit(&s.Config, []*derive.BatchData{batch}) // TODO: submit multiple batches
+		if err != nil {
+			s.log.Error("Error submitting batch", "err", err)
```
protolambda, line 224: This log is unsafe and will make the thing panic if the logger is closed before the go routine runs. (happened in older version during testing once for me). It's going to change with the new batch-submitter though. Maybe just add a comment for the audit that this batch-submission is going to change with #312 and that issue #308 is open to improve the life cycle of the node?

opnode/rollup/driver/step.go
```diff
@@ -172,24 +177,99 @@ func (d *outputImpl) step(ctx context.Context, l2Head eth.L2BlockRef, l2Finalize
...
+		fc.HeadBlockHash = lastHead.Hash
+		fc.SafeBlockHash = lastSafeHead.Hash
+	}
+
+	return lastHead, lastSafeHead, didReorg, nil
+}
+
+// attributesMatchBlock checks if the L2 attributes pre-inputs match the output
+// nil if it is a match. If err is not nil, the error contains the reason for the mismatch
+func attributesMatchBlock(attrs *l2.PayloadAttributes, parentHash common.Hash, block *types.Block) error {
```
protolambda, line 213: nit: we can improve the errors by inserting the different values with `%s <> %s` (calls `common.Hash.String()`, and e.g. adding the tx index `i` to know which tx failed to unmarshal.

protolambda (commented):
Some nits before I'd merge, and we definitely need more testing/polishing of the rollup node this week, but happy we can get this merged for first release/audit, nice work.

opnode/rollup/driver/state.go
trianglesphere: Noted issue 308

opnode/rollup/driver/step.go
trianglesphere: Added `fmt.Errorf` calls here.

protolambda (approved):
LGTM

PR #310, opened 2022-04-04
Add withdrawal relaying and verification

maurelian:
**Description**

This PR currently implements withdrawal verification, but needs some minor refactoring to:

1. [ ] Combine the Deposit Feed and Withdrawal Verification into one contract on L1. 
2. [ ] Bring the specs into alignment with the final implementation. 
3. [ ] Address minor feedback on #253 (which this PR is replacing).


norswap:
answering this commit message: https://github.com/ethereum-optimism/optimistic-specs/commit/aa363f14d824fef322913bf8ddd3874a63d214e8

Haha ouch on the submodule!
I'm not sure what `yarn add` does? These trie contracts are in their own npm package and we'd just pull that dependency?

Honestly, I don't think copying things over is that bad?

It's like having a pinned version in the repo, we can then later go see in the old monorepo if things have changed and import them. Also, this is only a temporary situation, since the old monorepo will disappear when we transition (or rather, when we merge this one in the old one, since this seems to be the privileged way to do it, and I made my peace with that).

In fact, if this repo were a branch in the monorepo, we would have exactly the same issue. Tthough in that case it would admittedly be a lot easier to merge in changes to the lib — assuming we're religious about not mixing changes to the trie library with other changes.

Any thoughts on changing this by a copy?

mslipper:
If the only thing we're using the monorepo for is the merkle trie, I agree with @norswap that we should just copy that over. It'll reduce the number of third-party deps we have to manage, and will be closer to what we end up putting in the monorepo anyway.

maurelian:
For this PR, I'd like to just stick with the submodules. The Trie and its deps would add about 1700 lines to the diff.

.gitmodules
```diff
@@ -1,9 +1,16 @@
 [submodule "packages/contracts/lib/ds-test"]
 	path = packages/contracts/lib/ds-test
 	url = https://github.com/dapphub/ds-test
+    ignore = dirty
```
tynes, line 4: tabs vs spaces issue here. You set `ignore = dirty` for debugging so that you could make changes to the files without git picking up on the changes? If so, I think that we don't want that set for the long term

packages/contracts/contracts/L1/Lib_WithdrawalVerifier.sol
```diff
@@ -0,0 +1,57 @@
+//SPDX-License-Identifier: MIT
+pragma solidity 0.8.10;
+
```
tynes, line 3: This file generally does not have docstrings

packages/contracts/contracts/L1/Lib_WithdrawalVerifier.sol
```diff
@@ -0,0 +1,57 @@
...
+import {
+    Lib_SecureMerkleTrie
+} from "../../lib/optimism/packages/contracts/contracts/libraries/trie/Lib_SecureMerkleTrie.sol";
+
+/**
+ * @title WithdrawalVerifier
+ */
+library WithdrawalVerifier {
+    struct OutputRootProof {
+        uint256 timestamp;
```
tynes, line 14: Why is the timestamp on this struct if it is unused?

.github/workflows/solidity-ci.yml
```diff
@@ -32,11 +32,12 @@ jobs:
             cache: 'yarn'
       - run: yarn install
 
-      - name: Build with HH
-        run: yarn build:hh
+      # We don't currently have any hardhat tests that should run in this context
```
tynes, line 35: I think this should be deleted instead of commented out

ops/Dockerfile.opnode
```diff
@@ -7,10 +7,16 @@ COPY go.mod /app/go.mod
 COPY go.sum /app/go.sum
 
 WORKDIR /app
+COPY ./ops/opnode-entrypoint.sh /app/opnode-entrypoint.sh
+COPY ./ops/rollup.json /app/rollup.json
+
 COPY ./opnode /app/opnode
 RUN make opnode
 
 FROM alpine:3.13
```
tynes, line 16: Can you update this to `alpine:3.15` so that we use the same image that we use in the monorepo as well?

ops/Dockerfile.opnode
```diff
@@ -7,10 +7,16 @@ COPY go.mod /app/go.mod
 COPY go.sum /app/go.sum
```
tynes, line 7: Can you update the builder image to be `golang:1.18.0-alpine3.15` ?

packages/contracts/contracts/L1/OptimismPortal.sol
```diff
@@ -0,0 +1,107 @@
...
+
+    /// @notice Address of the L2OutputOracle.
+    L2OutputOracle public immutable L2_ORACLE;
+
+    /**
+     * @notice Public variable which can be used to read the address of the L2 account which
+     * initated the withdrawal. Can also be used to determine whether or not execution is occuring
+     * downstream of a call to finalizeWithdrawalTransaction().
+     */
+    address public l2Sender = DEFAULT_L2_SENDER;
```
tynes, line 26: This would be a good usecase of transient memory

packages/contracts/contracts/L1/OptimismPortal.sol
```diff
@@ -0,0 +1,107 @@
...
+        address _sender,
+        address _target,
+        uint256 _value,
+        uint256 _gasLimit,
+        bytes calldata _data,
+        WithdrawalVerifier.OutputRootProof calldata _outputRootProof,
+        bytes calldata _withdrawalProof
+    ) external {
+        // Check that the timestamp is 7 days old.
+        require(_outputRootProof.timestamp <= block.timestamp - FINALIZATION_WINDOW, "Too soon");
```
tynes, line 70: Should we be using custom errors instead of `require` statements?

packages/contracts/contracts/L2/Withdrawer.sol
```diff
@@ -0,0 +1,77 @@
...
+        emit WithdrawalInitiated(nonce, msg.sender, _target, msg.value, _gasLimit, _data);
+    }
+
+    /**
+     * @notice Removes all ETH held in this contract from the state, by deploying a contract which
+     * immediately self destructs.
+     * For simplicity, this call is not incentivized as it costs very little to run.
+     * Inspired by https://etherscan.io/address/0xb69fba56b2e67e7dda61c8aa057886a8d1468575#code
+     */
+    function burn() external {
```
tynes, line 60: Will this contract be behind a delegatecall proxy? In the future, selfdestruct will likely just transfer all of the value and not delete the code associated with the account

packages/contracts/contracts/test/OptimismPortal.t.sol
```diff
@@ -0,0 +1,190 @@
...
+    bytes32 stateRoot = keccak256(abi.encode(1));
+    bytes32 withdrawerStorageRoot = 0xb8576230d94535779ec872748df80a094fcad002a8fc2b37c5b8fe250b384be6; // eth_getProof (storageHash)
+    bytes32 latestBlockhash = keccak256(abi.encode(2));
+
+    // This proof was generated by running
+    // make devnet-up
+    // Then in another terminal
+    // packages/contracts/test/withdrawer.spec.ts
+    // Invalid large internal hash
+    bytes withdrawalProof =
```
tynes, line 78: I think a more ideal way to do this is to have a simple script and use the ffi cheatcode to create the proof, see https://github.com/gakonst/lootloose/blob/7b639efe97836155a6a6fc626bf1018d4f8b2495/src/test/LootLooseMeta.t.sol#L98

mslipper:
One important note: `0x42...14` is an existing system address. We shouldn't use it as the address for the withdrawer contract.

packages/contracts/contracts/L1/DepositFeed.sol
```diff
@@ -59,9 +53,7 @@ contract DepositFeed {
         address from = msg.sender;
```
tynes, line 53: I wonder if a ternary operator here saves gas

packages/contracts/contracts/L1/OptimismPortal.sol
```diff
@@ -0,0 +1,95 @@
...
+    }
+
+    /**
+     * Accepts value so that users can send ETH directly to this contract and
+     * have the funds be deposited to their address on L2.
+     * Note: this is intended as a convenience function for EOAs. Contracts should call the
+     * depositTransaction() function directly.
+     */
+    receive() external payable {
+        depositTransaction(msg.sender, msg.value, 30000, false, bytes(""));
```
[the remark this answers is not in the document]
tynes: This doesn't inherit the `receive` function from `DepositFeed`?

packages/contracts/test/withdrawer.spec.ts
```diff
@@ -0,0 +1,213 @@
...
+
+const NON_ZERO_ADDRESS = '0x' + '11'.repeat(20)
+const NON_ZERO_GASLIMIT = BigNumber.from(50_000)
+const NON_ZERO_DATA = '0x' + '11'.repeat(42)
+
+if (!process.env.PRIVATE_KEY) {
+  throw new Error('You must define PRIVATE_KEY in your environment.')
+}
+
+const encodeWithdrawal = (args: {
```
tynes, line 29: Longer term, we should make a `class Withdrawal` that has `encode` and `decode` methods on it

packages/contracts/contracts/L2/Withdrawer.sol
maurelian: I think it's likely that it will. Although this _should_ selfdestruct the created contract, not this contract itself, however the test I just added is failing and says that the deployed contract has no balance...

specs/withdrawals.md
```diff
@@ -0,0 +1,204 @@
...
+        bytes message
+    );
+
+    function initiateWithdrawal(
+        address target,
+        bytes message,
+        uint256 gasLimit
+    ) external payable;
+
+    function burn();
```
tynes, line 119: Is this supposed to be public?

specs/withdrawals.md
```diff
@@ -0,0 +1,204 @@
...
+  [L2 output commitment construction](./proposals.md#l2-output-commitment-construction).
+
+1. A `WithdrawalMessageInclusionProof` is an MPT proof encoded in a struct as follows:
+
+   ```js
+   struct WithdrawalMessageInclusionProof {
+       WithdrawalMessage message;
+       bytes32 l2withdrawalsRoot;
+       bytes memory _key;      // storage key of the withdrawal message commitment
+       bytes memory _value;    // Always bytes32(1) (boolean validity status of the message)
```
tynes, line 198: I feel like we don't need to include a constant in this struct

ops/Dockerfile.opnode
tynes, line 16: This still needs to be updated to `alpine:3.15` unless it doesn't work

tynes: Then caching will work better with the other monorepo

mslipper: Done

packages/contracts/contracts/L1/OptimismPortal.sol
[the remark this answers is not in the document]
tynes: I picked it to cover the 21k intrinsic gas + a buffer but i'm not 100% sure that is sane, somebody more familiar with the go code should chime in

packages/contracts/contracts/L1/OptimismPortal.sol
```diff
@@ -0,0 +1,121 @@
...
+        require(
+            _timestamp <= block.timestamp - FINALIZATION_WINDOW,
+            "Finalization window has not yet passed."
+        );
+
+        // Get the output root and verify that the withdrawer contract's storage root is contained
+        // in it.
+        bytes32 outputRoot = L2_ORACLE.getL2Output(_timestamp);
+        require(
+            WithdrawalVerifier._verifyWithdrawerStorageRoot(outputRoot, _outputRootProof) == true,
```
tynes, line 90: I feel like a slightly better abstraction would be to use the library to compute the root and then check the equality here. The library is `internal` so it gets inlined anyways

packages/contracts/contracts/L1/OptimismPortal.sol
```diff
@@ -0,0 +1,121 @@
...
+        uint256 _value,
+        uint256 _gasLimit,
+        bytes calldata _data,
+        uint256 _timestamp,
+        WithdrawalVerifier.OutputRootProof calldata _outputRootProof,
+        bytes calldata _withdrawalProof
+    ) external {
+        // Check that the timestamp is 7 days old.
+        require(
+            _timestamp <= block.timestamp - FINALIZATION_WINDOW,
```
tynes, line 82: I believe this can be `unchecked` because `L2_ORACLE.getL2Output` + the `require` below

packages/contracts/contracts/L1/OptimismPortal.sol
```diff
@@ -0,0 +1,121 @@
...
+/**
+ * @title OptimismPortal
+ * @notice The OptimismPortal is a contract on L1 used to deposit and withdraw between L2 and L1.
+ */
+contract OptimismPortal is DepositFeed {
+    /// @notice Emitted when a withdrawal is finalized
+    event WithdrawalFinalized(bytes32 indexed);
+
+    /// @notice Value used to reset the l2Sender, this is more efficient than setting it to zero.
+    address internal constant DEFAULT_L2_SENDER = 0x000000000000000000000000000000000000dEaD;
```
tynes, line 22: This could use https://github.com/ethereum-optimism/optimism/blob/83ce16a00bb4100f504c5726aa5362e92077a758/packages/contracts/contracts/libraries/constants/Lib_DefaultValues.sol#L11

packages/contracts/contracts/L2/Withdrawer.sol
```diff
@@ -0,0 +1,93 @@
...
+        if (msg.sender != tx.origin) {
+            from = AddressAliasHelper.undoL1ToL2Alias(msg.sender);
+        }
+        bytes32 withdrawalHash = keccak256(
+            abi.encode(nonce, msg.sender, _target, msg.value, _gasLimit, _data)
+        );
+        withdrawals[withdrawalHash] = true;
+
+        emit WithdrawalInitiated(nonce, from, _target, msg.value, _gasLimit, _data);
+        nonce++;
```
tynes, line 66: This can be `unchecked`

packages/contracts/contracts/test/DepositFeed.t.sol
```diff
@@ -148,31 +147,11 @@ contract DepositFeedTest is DSTest {
         assertEq(address(df).balance, NON_ZERO_VALUE);
     }
 
-    // Test: depositTransaction should increase its eth balance when an EOA deposits a transaction with ETH
```
tynes, line 151: Any reason to delete this test?

packages/contracts/contracts/L1/OptimismPortal.sol
```diff
@@ -0,0 +1,121 @@
...
+import { L2OutputOracle } from "./L2OutputOracle.sol";
+
+/* Library Imports */
+import { WithdrawalVerifier } from "./Lib_WithdrawalVerifier.sol";
+
+/**
+ * @title OptimismPortal
+ * @notice The OptimismPortal is a contract on L1 used to deposit and withdraw between L2 and L1.
+ */
+contract OptimismPortal is DepositFeed {
```
tynes, line 17: I lean towards this contract using custom errors rather than reverts

packages/contracts/contracts/L1/OptimismPortal.sol
```diff
@@ -0,0 +1,95 @@
...
+import { WithdrawalVerifier } from "./Lib_WithdrawalVerifier.sol";
+import { L2OutputOracle } from "./L2OutputOracle.sol";
+
+contract OptimismPortal is DepositFeed {
+    event WithdrawalFinalized(bytes32 indexed);
+
+    // Value used to reset the l2Sender. This is more gas efficient that setting it to zero.
+    address internal constant DEFAULT_L2_SENDER = 0x000000000000000000000000000000000000dEaD;
+
+    uint256 public immutable FINALIZATION_WINDOW;
```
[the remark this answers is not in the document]
maurelian: Actually I don't think WINDOW is the appropriate term. Going to change to Finalization Period.

packages/contracts/contracts/L1/OptimismPortal.sol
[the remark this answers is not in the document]
maurelian: > This doesn't inherit the receive function from DepositFeed?

I relocated it here, because... well, it felt to me like a receive function should be visible in the child contract, ie. the `DepositFeed` is the low level logic, the `OptimismPortal` is more business logic. I could go either way tho.

packages/contracts/contracts/L1/Lib_WithdrawalVerifier.sol
```diff
@@ -0,0 +1,57 @@
...
+        bytes32 _withdrawalHash,
+        bytes32 _withdrawerStorageRoot,
+        bytes calldata _withdrawalProof
+    ) internal pure returns (bool) {
+        bytes32 storageKey = keccak256(
+            abi.encode(
+                _withdrawalHash,
+                uint256(1) // The withdrawals mapping is at the second slot in the layout
+            )
+        );
```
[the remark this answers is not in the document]
maurelian: Yes, though to be precise `M` there is just the ordering of all global storage variables (whether mapping or otherwise). So since the withdrawer has the [`nonce`](https://github.com/ethereum-optimism/optimistic-specs/blob/6839d522620483f10baf5427e055ac0627ced68d/packages/contracts/contracts/L2/Withdrawer.sol#L16) in the `0`th slot, this is at slot `1`. 
[Ref](https://docs.soliditylang.org/en/v0.8.4/internals/layout_in_storage.html#mappings-and-dynamic-arrays).

packages/contracts/contracts/L1/OptimismPortal.sol
```diff
@@ -0,0 +1,95 @@
...
+
+        // Make the call.
+        l2Sender = _sender;
+        _target.call{ value: _value, gas: _gasLimit }(_data);
+        l2Sender = DEFAULT_L2_SENDER;
+
+        // All withdrawals are immediately finalized. If the ability to replay a transaction is
+        // required, that support can be provided in external contracts.
+        finalizedWithdrawals[withdrawalHash] = true;
+        emit WithdrawalFinalized(withdrawalHash);
```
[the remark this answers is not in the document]
maurelian: We will update the messenger contracts to interface with the Portal and Withdrawer.

packages/contracts/contracts/L1/OptimismPortal.sol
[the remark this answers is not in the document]
tynes: True, that makes sense. Not super opinionated as both will be on etherscan when verified

specs/withdrawals.md
maurelian: yeah, specs need updating, it's not in the impl.

packages/contracts/contracts/L1/OptimismPortal.sol
maurelian: Agree that's cleaner.

packages/contracts/contracts/test/DepositFeed.t.sol
maurelian: Because I moved `receive()` to the Portal contract.

packages/contracts/contracts/L1/OptimismPortal.sol
maurelian: It could but for a single var, IMO it just adds one more place for the reader to jump to.

packages/contracts/contracts/test/OptimismPortal.t.sol
maurelian: I agree this is ugly, but good enough for now.

specs/glossary.md
```diff
@@ -286,6 +288,26 @@ cf. [Deposits Specification](deposits.md)
...
 ------------------------------------------------------------------------------------------------------------------------
 
+# Withdrawals
+
+[withdrawals]: glossary.md#withdrawals
+
+In general, a withdrawal is a transaction sent from L2 to L1 that may transfer of data and/or value.
+
+This term *withdrawal* is somewhat ambiguous as these "transactions" exist at multiple levels. In order to differentiate
+ between the L1 and L2 components of a withdrawal we introduce the following terms:
```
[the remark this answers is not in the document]
norswap: Not actually resolved? Not super important.

specs/withdrawals.md
```diff
@@ -0,0 +1,204 @@
...
+[deposit-tx-type]: glossary.md#deposited-transaction-type
+
+[g-withdrawal]: glossary.md#withdrawal
+[g-mpt]: glossary.md#merkle-patricia-trie
+[g-relayer]: glossary.md#withdrawals
+[g-execution-engine]: glossary.md#execution-engine
+**Table of Contents**
+
+[Withdrawals][g-withdrawal] are cross domain transactions which are initiated on L2, and finalized by a transaction
+executed on L1. They may be used to transfer data and/or ETH from L1 to L2.
```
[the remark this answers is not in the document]
norswap: None of these two things have been resolved.

specs/withdrawals.md
norswap: external according to the source file

norswap:
@mslipper You marked a bunch of things as resolved that weren't. I think there might have been some confusion here as @maurelian pushed updates to the spec but didn't address existing comments there, which you then assumed were resolved? Or maybe you have some local changes you forgot to push? I've unresolved the involved spec comments.

specs/withdrawals.md
```diff
@@ -0,0 +1,181 @@
...
+<!-- END doctoc generated TOC please keep comment here to allow auto update -->
+
+## Withdrawal Flow
+
+We first describe the end to end flow of initiating and finalizing a withdrawal:
+
+### On L2
+
+An L2 account sends a withdrawal message (and possibly also ETH) to the `Withdrawer` predeploy contract.
+   This is a very simple contract that stores the a hash of the withdrawal data.
```
norswap, line 57: extraneous spaces

norswap: I would use "withdrawal transaction data (nonce, destination, call data, etc)" instead of "withdrawal data".

packages/contracts/contracts/L1/OptimismPortal.sol
```diff
@@ -0,0 +1,157 @@
...
+        // Make the call.
+        _target.call{ value: _value, gas: _gasLimit }(_data);
+        l2Sender = DEFAULT_L2_SENDER;
+
+        // All withdrawals are immediately finalized. If the ability to replay a transaction is
+        // required, that support can be provided in external contracts.
+        finalizedWithdrawals[withdrawalHash] = true;
+        emit WithdrawalFinalized(withdrawalHash);
+    }
+}
```
norswap, line 157: For symmetry's sake, should we have a `Withdrawals` L1 contract and just make `OptimismPortal` extend both `DepositFeed` and `Withdrawals`?

I would also add a top comment in this file saying that these two facets need to live in the same contracts because we need to withdraw the ETH from the place where it was deposited.

specs/withdrawals.md
```diff
@@ -0,0 +1,181 @@
...
+
+### On L1
+
+1. A [relayer][g-relayer] submits the required inputs (see below) to the `OptimismPortal` contract. The relayer need
+   not be the same entity which initiated the withdrawal on L2.
+2. The `OptimismPortal` contract retrieves the output root from the `OutputOracle`'s `l2Outputs()` function, and
+   performs the remainder of the verification process internally.
```
norswap, line 64: This doesn't specify which output root is retrieved. Maybe specify in (1) that one of the input is a timestamp for which an L2 output root exists that commits to the withdrawal as registered on L2? Then say that we retrieve the output corresponding to the timestamp.

i.e. 

```
1. A [relayer][g-relayer] submits the required inputs to the `OptimismPortal` contract. The relayer need
   not be the same entity which initiated the withdrawal on L2.
   
   These inputs include the withdrawal transaction data, inclusion proofs, and a timestamp. The timestamp
   must be one for which an L2 output root exists, which commits to the withdrawal as registered on L2.
   More details are given below.

2. The `OptimismPortal` contract retrieves the output root for the given timestamp from the `OutputOracle`'s
   `l2Outputs()` function, and performs the remainder of the verification process internally.
 ```

File: specs/glossary.md (hunk 25 above)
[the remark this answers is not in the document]
Comment (maurelian) [reply specs/glossary.md:298]: Oh, I didn't understand the suggestion because I was focused on the end of the line. Will fix.

File: packages/contracts/contracts/L1/OptimismPortal.sol [hunk 30]
```diff
@@ -0,0 +1,95 @@
...
+    /**
+     * Finalizes a withdrawal transaction.
+     * @param _nonce Nonce for the provided message.
+     * @param _sender Message sender address on L2.
+     * @param _target Target address on L1.
+     * @param _data Data to send to the target.
+     * @param _gasLimit Gas to be forwarded to the target.
+     * @param _outputRootProof Inclusion proof of the withdrawer contracts storage root.
+     * @param _withdrawalProof Inclusion proof for the given withdrawal in the withdrawer contract.
+     */
```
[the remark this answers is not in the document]
maurelian: Good catch. Surprised the compiler didn't complain.

packages/contracts/contracts/L1/Lib_WithdrawalVerifier.sol
maurelian: Good callout, removed.

packages/contracts/contracts/L1/OptimismPortal.sol
maurelian: Sure, implemented that change.
Please double check that I've correctly inverted the comparisons in this commit: 548be0f61a9109e1fd951baebccd768fd69f1d0d

specs/withdrawals.md
```diff
@@ -0,0 +1,181 @@
...
+## Withdrawal Verification and Finalization
+
+The following inputs are required to verify and finalize a withdrawal:
+
+- Withdrawal transaction data:
+  - `nonce`: Nonce for the provided message.
+  - `sender`: Message sender address on L2.
+  - `target`: Target address on L1.
+  - `data`: Data to send to the target.
+  - `gasLimit`: Gas to be forwarded to the target.
```
norswap, line 133: This misses the value field.

specs/withdrawals.md
```diff
@@ -0,0 +1,181 @@
...
+  - `outputRootProof`: The 4 values which are used to derive the output root.
+  - `withdrawalProof`: An inclusion proof for the given withdrawal in the withdrawer contract.
+
+These inputs must satisfy the following conditions:
+
+1. The `timestamp` is at least `FINALIZATION_PERIOD` seconds old.
+1. `OutputOracle.l2Outputs(timestamp)` returns a non-zero value `l2Output`.
+2. The `l2Output` value is equal to the hash of the components of the `outputRootProof`
+1. The `withdrawalProof` is a valid inclusion proof demonstrating that a hash of the Withdrawal transaction data
+   is contained in the storage of the Withdrawer contract on L2.
```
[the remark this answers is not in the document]
norswap: This is missing a validity condition regarding `outputRootProof`.

tynes:
It looks like some of slipper's changes got reverted

.github/workflows/solidity-ci.yml
tynes, line 35: These should be deleted

.gitmodules
tynes, line 4: These ignore blocks should be deleted

specs/withdrawals.md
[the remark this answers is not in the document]
maurelian: > This is missing a validity condition regarding `outputRootProof`.

I think it's covered by line 143, how about if I rewrite it to emphasize the proof rather than the output?

```diff
- 1. The `l2Output` value is equal to the hash of the components of the `outputRootProof`
+ 1. The keccak256 hash of the `outputRootProof` values is equal to the `l2Output`.
```

specs/withdrawals.md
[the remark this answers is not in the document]
maurelian: > @mslipper You marked a bunch of things as resolved that weren't. I think there might have been some confusion here as @maurelian pushed updates to the spec but didn't address existing comments there, which you then assumed were resolved? Or maybe you have some local changes you forgot to push? I've unresolved the involved spec comments.

Yeah, we had a miscommunication and I over-wrote his changes. 🤦‍♀️

specs/withdrawals.md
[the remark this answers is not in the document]
norswap: Oh sorry, you're right. I think it's fine as-is, so up to you :)

norswap (approved):

packages/contracts/contracts/L1/OptimismPortal.sol
```diff
@@ -0,0 +1,158 @@
...
+
+        // Get the output root.
+        bytes32 outputRoot = L2_ORACLE.getL2Output(_timestamp);
+
+        // Verify that the output root can be generated with the elements in the proof.
+        if (outputRoot != WithdrawalVerifier._deriveOutputRoot(_outputRootProof)) {
+            revert InvalidOutputRootProof();
+        }
+
+        // Verify that the hash of the withdrawal transaction's arguments are included in the
```
tynes, line 128: Computing the `withdrawalHash` could go into the `WithdrawalVerifier` library but not super opinionated

packages/contracts/contracts/test/Withdrawer.t.sol
```diff
@@ -0,0 +1,112 @@
...
+    function test_burn() external {
+        assertEq(address(wd).balance, NON_ZERO_VALUE);
+        vm.expectEmit(true, false, false, false);
+        emit WithdrawerBalanceBurnt(NON_ZERO_VALUE);
+        wd.burn();
+
+        // The Withdrawer should have no balance
+        assertEq(address(wd).balance, 0);
+    }
+
```
tynes, line 111: nit extra newline

packages/contracts/test/withdrawer.spec.ts
```diff
@@ -0,0 +1,213 @@
...
+      )
+
+      // Deploy the WithdrawalVerifier with a 0 second finalization delay.
+      withdrawalVerifier = await (
+        await new WithdrawalVerifier__factory(l1Signer)
+      ).deploy(l2OutputOracle.address, 0)
+
+      // create an output root that we can prove against only the storage root matters
+      // for our purposes.
+      const outputRoot = ethers.utils.keccak256(
```
tynes, line 168: This kind of thing should be turned into a helper function. There should be a ts lib that has the same methods as the solidity libs used

tynes (approved):

PR #311, opened 2022-04-04
docs: fix typo in deposits.md

Inphi:


specs/deposits.md
```diff
@@ -234,7 +234,7 @@ The deposit contract handles two special cases:
 
 [address-aliasing]: #address-aliasing
 
-If the caller is not a contract, the address will be transformed by adding
```
maurelian, line 237: That's not a typo. :)

Impl here: 

https://github.com/ethereum-optimism/optimistic-specs/blob/e6db45f5ff024d23e78c37ae2370114d975bb2d6/packages/contracts/contracts/L1/DepositFeed.sol#L51-L55

[Background here](https://github.com/ethereum-optimism/optimism/discussions/1480).

Inphi: according to the other thread:
> This aliasing ONLY applies for contracts sending messages to L2. If an EOA sends a message to L2 then the address is NOT aliased. Here's the relevant code from the CTC

Whereas the documentation here is stating the opposite.

maurelian (approved):
You're right.

specs/deposits.md
maurelian: It seems I'm struggling with negation today. 🤦

PR #312, opened 2022-04-04
feat: add standalone BSS service

cfromknecht:
**Description**
This PR builds on #301, adding an external batch submitter in place of the simple, in-process BSS. The new batch submitter operates by continuously querying the `optimism_getBatchBundle` RPC added in #301, and attempting to submit the bundle constructed by the Sequencer. 

In order to properly handle reorgs, the BSS maintains a small database containing identifiers for each of the most recently published batches. The batches are identified by the final L2 block hash in each batch. When querying `optimism_getBatchBundle`, the BSS supplies the list of historic L2 block hashes, which the Sequencer can use to determine a common ancestor from which to generate the next bundle in the event that a reorg is detected. Currently this database consists of a simple JSON file to facilitate local development/testing, but would be replaced with a distributed variant for production workloads for fault tolerance.

The service is currently integrated into the existing l2os binary, as I was unsure if we wanted to create a third binary or potentially just rename the existing one to account for both subservices. The current iteration passes the E2E tests.

Open Questions:
 - Do we want the bss and l2os to be in the same process?
 - In order to satisfy the e2e, the poll interval must be made very low to ensure that transactions submitted to the sequencer are promptly included in batches. Is this a nuance of the e2e setup or is this something that would carry over to production?

**Metadata**
- Fixes ENG-2071


protolambda:
> Do we want the bss and l2os to be in the same process?

I would strongly prefer two separate processes. Preferably two separate binaries too. Separation of concerns is great to have long-term, as well as for security. E.g. we can keep withdrawals working with just the output-root submitter, even if batches and batch-submitting are completely broken.

> In order to satisfy the e2e, the poll interval must be made very low to ensure that transactions submitted to the sequencer are promptly included in batches. Is this a nuance of the e2e setup or is this something that would carry over to production?

I think this is because the e2e uses a 2 second L1 block time, and a 1 second L2 block time, and a relatively short sequencing window. Great fast feedback for testing, but indeed requires integrations to also act faster.

cfromknecht:
@protolambda I've successfully split out the batch submitter into a separate binary. There is still some shared logic between them that we could clean up (taking the form of the bss-core pkg in the monorepo), but this PR was getting quite large as it is. There is quite a bit of redundancy in splitting the two, but I think it's workable.

The first 4 commits after yours I think could all be squashed down, but I've left them separate so that you can review the rationale. I'm still in the process of getting everything to pass the devnet tests, but this is likely a configuration error as the E2E unit tests pass.

> we can keep withdrawals working with just the output-root submitter, even if batches and batch-submitting are completely broken.

Wouldn't this present a dangerous situation? Allowing users to withdraw funds that could be reorged away if the batches miss the sequencing window?

> I think this is because the e2e uses a 2 second L1 block time, and a 1 second L2 block time, and a relatively short sequencing window. Great fast feedback for testing, but indeed requires integrations to also act faster.

Got it, the short L1 block time i think is likely the culprit then.

protolambda:
> Wouldn't this present a dangerous situation? Allowing users to withdraw funds that could be reorged away if the batches miss the sequencing window?

Good point, if they miss the sequencing window then we wouldn't want to submit them as proposals. I guess the inverse holds true though: if we're failing to make proposals that doesn't mean we should stop submitting batches.

bss/batch_submitter.go
```diff
@@ -0,0 +1,220 @@
...
+
+		log.Info("Starting L2 Output Submitter")
+
+		if err := batchSubmitter.Start(); err != nil {
+			log.Error("Unable to start L2 Output Submitter", "error", err)
+			return err
+		}
+		defer batchSubmitter.Stop()
+
+		log.Info("L2 Output Submitter started")
```
protolambda, line 55: The logs here still refer to an "output submitter", while this is now a separate batch-submitter binary.

bss/config.go
```diff
@@ -0,0 +1,85 @@
...
+	SafeAbortNonceTooLowCount uint64
+
+	// ResubmissionTimeout is time we will wait before resubmitting a
+	// transaction.
+	ResubmissionTimeout time.Duration
+
+	// Mnemonic is the HD seed used to derive the wallet private keys for both
+	// the sequence and proposer. Must be used in conjunction with
+	// SequencerHDPath and ProposerHDPath.
+	Mnemonic string
```
protolambda, line 43: I would really like us to not use mnemonics for this type of thing. They should be strictly for backup purposes, and span multiple accounts. We don't want one leaked key to leak all of the keys (if we even use the same mnemonic for the multiple services). It's fine for now, but should be addressed in a next PR before production.

bss/db/history_db.go
```diff
@@ -0,0 +1,136 @@
...
+	"encoding/json"
+	"io/ioutil"
+	"math/big"
+	"os"
+
+	"github.com/ethereum/go-ethereum/common"
+)
+
+type L2BlockEntry struct {
+	BlockNumber *big.Int    `json:"block_number"`
```
protolambda, line 13: Using `uint64` or `hexutil.Uint64` would also be fine, I don't think the block number will ever go over 52 or 64 bits.

opnode/l2/source.go
```diff
@@ -246,3 +248,61 @@ func PayloadToBlockRef(payload *ExecutionPayload, genesis *rollup.Genesis) (eth.
...
 }
+
+type ReadOnlySource struct {
+	rpc     *rpc.Client       // raw RPC client. Used for methods that do not already have bindings
+	client  *ethclient.Client // go-ethereum's wrapper around the rpc client for the eth namespace
+	genesis *rollup.Genesis
+	log     log.Logger
+}
+
+func NewReadOnlySource(ll2Node *rpc.Client, genesis *rollup.Genesis, log log.Logger) (*ReadOnlySource, error) {
```
protolambda, line 259: is `ll2` (two L) intentional? Typo? (might be my own typo)

opnode/node/api.go
```diff
@@ -74,3 +90,105 @@ func toBlockNumArg(number rpc.BlockNumber) string {
 	}
 	return hexutil.EncodeUint64(uint64(number.Int64()))
 }
+
+type BatchBundleRequest struct {
+	// L2History is a list of L2 blocks that are already in-flight or confirmed.
+	// The rollup-node then finds the common point, and responds with that point as PrevL2BlockHash and PrevL2BlockNum.
+	// The L2 history is read in order of the provided hashes, which may contain arbitrary gaps and skips.
+	// The first common hash will be the continuation point.
+	// A batch-submitter may search the history using gaps to find a common point even with deep reorgs.
```
protolambda, line 99: I don't think the batch-submitter does a new request with different (longer or older) history when an error is returned by the rollup node that no common chain was found with previous hashes. Can you create an issue to test deep reorgs + implement this in the batch submitter?

opnode/node/api.go
```diff
@@ -74,3 +90,105 @@ func toBlockNumArg(number rpc.BlockNumber) string {
...
+		if err != nil {
+			return nil, fmt.Errorf("failed to convert L2 block %d (%s) to batch: %v", i, l2Block.Hash(), err)
+		}
+		lastL2Hash = l2Block.Hash()
+		lastL2BlockNum = l2Block.Number().Uint64()
+		if batch == nil { // empty block, nothing to submit as batch
+			continue
+		}
+		batches = append(batches, batch)
+		// TODO: estimate size of all batches so far (including compressing them together, if we reached max size yet)
```
protolambda, line 174: TODO: we're should not append the batch, and break this loop, once the total size gets too large.

protolambda (commented):
This is looking great! A few minor things to address before we can merge though. And you can squash those commits you mentioned if you want to.

opnode/l2/source.go
cfromknecht: looks like it's a typo in `Source` as well, fixed.

opnode/node/api.go
cfromknecht: Currently I think the only accurate way to do this is to encode the full set with each loop iteration. Is there a better way to estimate the size of each incremental batch into a running total? Do you think its sufficient to just add up the length of `rlp.Encode(batch)`?

bss/config.go
tynes: We need a tool that can take a mnemonic and then produce encrypted JSON keystores from derived keys

tynes: In practice each service uses their own mnemonic and the service that needs 2 keys derives both from the same mnemonic. If two unrelated keys were configured and the service was compromised, the security model would be exactly the same

cfromknecht:
@protolambda size constraints are now properly enforced in https://github.com/ethereum-optimism/optimistic-specs/pull/312/commits/e2c6db59af04af5872893de1d9cdbb0a7a18050e

Still seeing the same withdrawal error in the devnet integration tests, will continue to debug.

bss/db/history_db.go
```diff
@@ -0,0 +1,133 @@
...
+
+	var entries = make([]L2BlockEntry, 0, len(h.Entries))
+	for _, e := range h.Entries {
+		// Remove entries that are behind our lookback.
+		if e.BlockNumber < minBlockNumber {
+			continue
+		}
+
+		// Prune entries that occur after the new entry.
+		if e.BlockNumber >= entry.BlockNumber {
```
protolambda, line 34: This may break during reorgs I think. Isn't the idea to just keep a history of the last N submitted blocks, regardless of block number? Then the opnode can receive that history (in order from new to old), and find the first common canonical block, to then continue batching from.

opnode/node/api.go
```diff
@@ -74,3 +90,174 @@ func toBlockNumArg(number rpc.BlockNumber) string {
...
+		}
+
+		bundleSize := uint64(len(buf.Bytes()))
+
+		// Sanity check the bundle size respects the desired maximum. If we have
+		// exceeded the max size, prune the last block. This is very unlikely to
+		// occur since our initial greedy estimate has a very small, bounded
+		// error tolerance, so simply remove the last block and try again.
+		if bundleSize > uint64(req.MaxSize) {
+			batches = batches[:len(batches)-1]
```
protolambda, line 227: `lastL2Hash` and `lastL2BlockNum` would be wrong after pruning this last batch

bss/sequencer/driver.go
```diff
@@ -0,0 +1,205 @@
...
+			L2History: history.Ancestors(),
+			MinSize:   hexutil.Uint64(d.cfg.MinL1TxSize),
+			MaxSize:   hexutil.Uint64(d.cfg.MaxL1TxSize),
+		},
+	)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	// Bundle is not available yet, return the tip of our known chain.
```
protolambda, line 89: The comment is confusing. It returns the `tip+1`, not the tip. Is that `+1` correct?

opnode/node/api.go
```diff
@@ -74,3 +90,174 @@ func toBlockNumArg(number rpc.BlockNumber) string {
...
+		}
+		if canonBlock.Hash == h {
+			log.Info("Found common canonical block")
+			// found a common canonical block!
+			found = eth.BlockID{Hash: canonBlock.Hash, Number: canonBlock.Number}
+			break
+		}
+		log.Info("Found block, but not canonical")
+	}
+	if found == (eth.BlockID{}) { // none of the L2 history could be found. At least
```
protolambda, line 153: incomplete comment

opnode/node/api.go
cfromknecht: indeed, this one of the remaining issues i was planning to tackle after the replay issue is resolved. locally i have panic in this conditional to make sure it's not the cause

opnode/node/api.go
protolambda: The current thing looks ok to me. And this is out of protocol consensus anyway. If we want we can optimize this later.

opnode/test/system_test.go
```diff
@@ -207,12 +217,44 @@ func TestSystemE2E(t *testing.T) {
...
+	}()
+
 	cfg := defaultSystemConfig(t)
 
 	sys, err := cfg.start()
 	require.Nil(t, err, "Error starting up system")
 	defer sys.Close()
 
+	// Batch Submitter
+	batchSubmitter, err := bss.NewBatchSubmitter(bss.Config{
```
protolambda, line 235: We should move this to the test setup, so we run the batch submitter in all E2E tests, instead of just this one.

protolambda:
#392 is almost merged. I think you can then rebase onto `main` and that would also fix the merge conflicts with the L1 block info bindings and gas snapshot.

bss/sequencer/driver.go
```diff
@@ -0,0 +1,205 @@
...
+	// There is nothing to be done if the rollup returns a last block hash equal
+	// to the previous block hash. Return identical start and end block heights
+	// to signal that there is no work to be done.
+	start := big.NewInt(int64(batchResp.PrevL2BlockNum) + 1)
+	if batchResp.LastL2BlockHash == batchResp.PrevL2BlockHash {
+		return start, start, nil
+	}
+
+	d.currentBatch = batchResp
+	end := big.NewInt(int64(batchResp.LastL2BlockNum))
```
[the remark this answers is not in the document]
protolambda: Comment of the function says the `end` is exclusive. Either the comment needs to be changed, or the `+1` is missing.

opnode/test/system_test.go
protolambda: Still think this needs to be done before merge. Can't remove the old batch submitter from the opnode if we don't run the new external one in every e2e test.

cfromknecht: I moved over both the batch submitter and the l2 output submitter to the harness so they run on every test case

bss/sequencer/driver.go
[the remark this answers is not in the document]
cfromknecht: Added the +1 to make `end` exclusive 👍

cfromknecht:
@protolambda alrighty, all comments have been addressed, rebased over main, and build is back to green!

protolambda:
Not sure if something fragile changed in the devnet setup, but locally the `opnode` is not starting correctly. It fails with:
```
INFO [05-04|12:42:00.644] Initializing Rollup Node 
INFO [05-04|12:42:00.646] Starting rollup node 
INFO [05-04|12:42:00.646] Starting OpNode 
INFO [05-04|12:42:00.646] Fetching rollup starting point 
INFO [05-04|12:42:00.646] Attaching execution engine(s) 
ERROR[05-04|12:42:00.648] Could not start a rollup node            err="failed to parse L1 info deposit tx from L2 block: data is unexpected length: 84"
ERROR[05-04|12:42:00.648] Unable to start rollup node              error="failed to parse L1 info deposit tx from L2 block: data is unexpected length: 84"
CRIT [05-04|12:42:00.648] Application failed                       message="failed to parse L1 info deposit tx from L2 block: data is unexpected length: 84"
```

Which seems related to the recent changes in the L1 info deposit tx format. Maybe it's pulling in old code due to your changes in the docker files?

It used to check for a `4+8+8+32+32=84` input (see [here](https://github.com/ethereum-optimism/optimistic-specs/blob/273c8abddee294dc61bedf6b077c0b21fb9984e3/opnode/rollup/derive/invert.go#L13) but that has since changed. In the current rebase of this PR it should be checking if it's `4+32+32+32+32+32` (see [here](https://github.com/cfromknecht/optimistic-specs/blob/63e4bebf8f81c9078b0525ad8acd59ae53a41f8d/opnode/rollup/derive/l1_block_info.go#L41).

mslipper (approved):

protolambda (approved):

PR #315, opened 2022-04-04
Derivation fixes: update derivation timestamp bounds, clear up payload attributes

protolambda:
This implements the timestamp derivation design as discussed in #251, going for the proposal of @trianglesphere [here](https://github.com/ethereum-optimism/optimistic-specs/discussions/251#discussioncomment-2487856), with max-fill definition of @norswap [here](https://github.com/ethereum-optimism/optimistic-specs/discussions/251#discussioncomment-2491917) and sequencer drift (or "diff") like in https://github.com/ethereum-optimism/optimistic-specs/pull/264 but now with the fixed bounds definition from the discussion thread.

Additionally, this cleans up the rollup-node spec:
- The `min`/`max` really affect the batch parsing early on. Other batches should be ignored right away when not within bounds.
- The L1 input and sequencing window sections are unified
- Payload attributes are not implicitly derived from parsed inputs anymore, but explicitly for a specific range of timestamps, including clarification of any gaps/trailing blocks.
- Individual payload attributes are a sub-section of this, and now utilize the `noTxPool` field that the implementation has been using for a while now. (And exec-engine spec is updated to describe the field)

TODO:
- improve testing of timestamp edge cases (both as verifier and as sequencer)


specs/rollup-node.md
```diff
@@ -126,9 +109,16 @@ The rollup reads the following data from the [sequencing window][g-sequencing-wi
     - Batches not matching filter criteria are ignored:
       - `batch.epoch == sequencing_window.epoch`, i.e. for this sequencing window
       - `(batch.timestamp - genesis_l2_timestamp) % block_time == 0`, i.e. timestamp is aligned
-      - `min_l2_timestamp < batch.timestamp < max_l2_timestamp`, i.e. timestamp is within range
+      - `min_l2_timestamp <= batch.timestamp < max_l2_timestamp`, i.e. timestamp is within range
```
trianglesphere, line 112: Norswap suggested this as an AND condition here (& I agree): `min_l2_timestamp >= l1_timestamp`

specs/rollup-node.md
```diff
@@ -192,21 +182,35 @@ To encode user-deposited transactions, refer to the following sections of the de
 - [The Deposited Transaction Type](deposits.md#the-deposited-transaction-type)
 - [User-Deposited Transactions](deposits.md#user-deposited-transactions)
 
-### Building the Payload Attributes
+### Deriving all Payload Attributes of a sequencing window
+
+A sequencing window is derived into a variable number of L2 blocks, defined by a range of timestamps:
+
+- Starting at `min_l2_timestamp`, as defined in the batch filtering.
+- Up to and including `new_head_l2_timestamp = max(highest_valid_batch_timestamp, next_l1_timestamp - l2_block_time, min_l2_timestamp)`
```
trianglesphere, line 190: Nice catch to make `next_l1_timestamp` work

specs/rollup-node.md
protolambda: `min_l2_timestamp` is defined below, not something we can handle like a user input here in the filter condition `AND` list. And it always has to be `prev_l2_timestamp + l2_block_time`, or else there would be a gap.

Do you mean to define `min_l2_timestamp >= l1_timestamp` as an invariant, that just always holds true based on the other conditions?

protolambda:
See last "TODO" commit: the E2E fails when you add the `NEXT_L1_TIMESTAMP - L2_BLOCKTIME` part to the `MAXIMUM_L2_FILL`.

Maybe that is crowding out the tx attempt with more instant early empty blocks? Not sure why E2E fails or what's wrong with this max.

specs/rollup-node.md
```diff
@@ -192,21 +181,37 @@ To encode user-deposited transactions, refer to the following sections of the de
...
 attributes][g-payload-attr] as an [expanded version][expanded-payload] of the [`PayloadAttributesV1`] object, which
-includes an additional `transactions` field.
+includes additional `transactions` and `noTxPool` fields.
 
-The object's properties must be set as follows:
+Each of the timestamped transaction lists translates to a `PayloadAttributesV1` as follows:
 
-- `timestamp` is set to the timestamp of the L1 block.
-- `random` is set to the *random* L1 block attribute
+- `timestamp` is set to the timestamp of the L2 block.
```
norswap, line 209: "the timestamp of the transaction list"

norswap (commented):
Reviewing only the spec

protolambda:
Will rebase this onto `main`

opnode/rollup/driver/state.go
```diff
@@ -173,19 +175,33 @@ func (s *state) findNextL1Origin(ctx context.Context) (eth.L1BlockRef, error) {
 	}
 	// Somehow reorg'd. Will let the state loop take care of it.
 	if curr.Hash != s.l2Head.L1Origin.Hash {
```
trianglesphere, line 177: Can cut this as we are now doing `BlockRefByHash` (was a holdover from when it was by number)

opnode/rollup/driver/state.go
```diff
@@ -173,19 +175,33 @@ func (s *state) findNextL1Origin(ctx context.Context) (eth.L1BlockRef, error) {
...
-		return eth.L1BlockRef{}, errors.New("Unknown L1Origin")
+		return eth.L1BlockRef{}, errors.New("unknown L1Origin")
+	}
+
+	nextOrigin, err := s.l1.L1BlockRefByNumber(ctx, curr.Number+1)
+	if errors.Is(err, ethereum.NotFound) {
+		// no new L1 origin found, keep the current one
+		s.log.Info("No new L1 origin, staying with current one", "l2Head", s.l2Head, "l1Origin", curr)
+		return nextOrigin, nil
```
trianglesphere, line 185:
```suggestion
		return curr, nil
```

Maybe also rename `curr` to `currentOrigin`

opnode/rollup/driver/state.go
```diff
@@ -196,13 +212,25 @@ func (s *state) createNewL2Block(ctx context.Context) (eth.L1BlockRef, error) {
...
+	// If we are behind the next L1 origin, then we should be deriving empty blocks (regular verifier work) until the L2 time catches up with L1 again.
+	if nextL2Time < nextOrigin.Time {
+		s.log.Warn("Skipping block production because the next block time is behind the next L1 Origin",
+			"l2Head", s.l2Head, "nextL2Time", nextL2Time, "l1Origin", nextOrigin, "l1OriginTime", nextOrigin.Time)
+		return eth.L1BlockRef{}, nil
+	}
+
+	// TODO: we can give ourselves less slack here, and prefer to adopt the next L1 block more eagerly,
+	// to ensure we have more time to submit the next block we produce, at the cost of sequencing less eagerly.
+	maxL2Time := nextOrigin.Time + s.Config.MaxSequencerDrift
```
trianglesphere, line 225: Ditto need at least one L2 block per epoch.

trianglesphere (changes requested):
Generally looks good. I identified two ares I think we need to fix:
1. What happens if the time between L1 blocks is larger than the max sequencer drift - are these filled in with empty blocks? If so, we need to do that on the block creation side as well. If not it breaks the 2s block time.
2. You need to ensure that at least one L2 block per epoch is constructed on the sequencer side. It's not the worst bug b/c it will be done properly on the verify side, but it's still annoying to throw out L2 blocks.


For both of these issues, if the verifier does it correctly all the time and the sequencer is correct during normal case operation, it will work during normal cases. During edge cases the verifier will cause reorgs of the unsafe portion which is annoying, but not yet safety critical.

specs/rollup-node.md
```diff
@@ -126,9 +108,16 @@ The rollup reads the following data from the [sequencing window][g-sequencing-wi
...
+        - `min_l2_timestamp = prev_l2_timestamp + l2_block_time`
+          - `prev_l2_timestamp` is the timestamp of the previous L2 block: the last block of the previous epoch,
+              or the L2 genesis block timestamp if there is no previous epoch.
+          - `l2_block_time` is a configurable parameter of the time between L2 blocks
+        - `max_l2_timestamp = max(l1_timestamp + max_sequencer_drift, min_l2_timestamp + l2_block_time)`
+          - `l1_timestamp` is the timestamp of the L1 block associated with the L2 block's epoch
+          - `max_sequencer_drift` is the most a sequencer is allowed to get ahead of L1
       - The batch is the first batch with `batch.timestamp` in this sequencing window,
-        i.e. one batch per L2 block number
+      i.e. one batch per L2 block number
```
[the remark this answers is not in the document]
protolambda: I think the formatter split the line into two and didn't get the indent right. Will fix.

specs/rollup-node.md
```diff
@@ -192,21 +181,37 @@ To encode user-deposited transactions, refer to the following sections of the de
...
-### Building the Payload Attributes
+### Deriving all Payload Attributes of a sequencing window
+
+A sequencing window is derived into a variable number of L2 blocks, defined by a range of timestamps:
+
+- Starting at `min_l2_timestamp`, as defined in the batch filtering.
+- Up to and including
+  `new_head_l2_timestamp = max(highest_valid_batch_timestamp, next_l1_timestamp - l2_block_time, min_l2_timestamp)`
+  - `highest_valid_batch_timestamp = max(batch.timestamp for batch in filtered_batches)`,
+    or `0` if no there are no `filtered_batches`.
```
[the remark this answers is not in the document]
protolambda: The batch contents are enumerated in the batch definition 50 or so lines up. But I see how usage of a `timestamp` can be confusing. I'll try to clarify.

specs/rollup-node.md
```diff
@@ -192,21 +181,37 @@ To encode user-deposited transactions, refer to the following sections of the de
...
+  - `next_l1_timestamp` is the timestamp of the next L1 block.
+
+The L2 chain is extended to `new_head_l2_timestamp`, has a fixed block time, and no more than one batch per block.
```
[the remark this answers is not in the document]
protolambda: The filtering already defines "The batch is the first batch with `batch.timestamp` in this sequencing window,". I'll remove this duplicate and clean up the wording.

specs/rollup-node.md
```diff
@@ -192,21 +181,37 @@ To encode user-deposited transactions, refer to the following sections of the de
...
+This means that every `l2_block_time` that has no batch is interpreted as one with an empty sequenced transaction-list,
+thus construing a L2 block that only contains deposit transaction(s).
```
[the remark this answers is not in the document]
protolambda: I think it's confusing to say the deposits are part of the transaction list we just got from a batch, and the description of how deposits are placed in the payload sits awkwardly in the batch derivation section. Will fix that. And good point about clarifying `l2_block_time` increments for each block.

opnode/rollup/driver/state.go
protolambda: It's a duplicate with the check in the `createNewBlock` method (this function is the only caller). I'll fix one and remove the other.

protolambda:
@trianglesphere @norswap thanks for review, implemented the suggestions + fixed other minor stuff.

@trianglesphere the maxL2Time condition during sequencer block construction is now conditional on it not being the first block of the epoch, to ensure at least 1 block is output. But I am still looking at your first point, about L1 time between blocks being larger than the sequencer drift. Not sure what the best approach is there.

trianglesphere:
> But I am still looking at your first point, about L1 time between blocks being larger than the sequencer drift. Not sure what the best approach is there.

I see two approaches here:
- Keep the 2s block time and fill them out with empty blocks (must change block derivation)
- Be willing to drop the 2s block time and have a cap on the number of L2 blocks in an epoch (must modify block time req)

I think that either is possible to implement.

protolambda:
TLDR of the solution to the sequencer drift vs L1 block time gap problem: the verifier case was already solved with the `next_l1_timestamp - l2_block_time` part of the `new_head_l2_timestamp`. The sequencer just needed to consider that, by continuing on the current epoch until it catches up to the start of the next epoch.

opnode/rollup/driver/state.go
```diff
@@ -196,24 +217,42 @@ func (s *state) createNewL2Block(ctx context.Context) (eth.L1BlockRef, error) {
...
+	}
+
+	// If the L1 origin changed this block, then we are in the first block of the epoch
+	isFirstEpochBlock := s.l2Head.L1Origin.Number != nextOrigin.Number
+
+	// We create at least 1 block per epoch. After that we have to enforce the max timestamp.
+	//
+	// TODO: we can give ourselves less slack here, and prefer to adopt the next L1 block more eagerly,
+	// to ensure we have more time to submit the next block we produce, at the cost of sequencing less eagerly.
+	maxL2Time := nextOrigin.Time + s.Config.MaxSequencerDrift
```
trianglesphere, line 236: This still have the same problem that it will not fill out blocks up to the next L1 block if it instead should be. This is not consensus critical, but still slightly odd.

trianglesphere (approved):
This code does not (as best I can tell) have correctness issues. The last little thing is that when the L1 block time is greater than the max sequencer drift it will stop producing new L2 blocks (I'm not sure that it recovers prior to the reconciliation process).

opnode/rollup/driver/state.go
protolambda: Ah, forgot we were checking it there again. We should allow block creation, the verifier will accept the blocks, will update this.

protolambda: [8393ced](https://github.com/ethereum-optimism/optimistic-specs/pull/315/commits/8393ced722e5f15f65f964e6fecad9be2b45d43c) should fix this

trianglesphere: lgtm

PR #316, opened 2022-04-05
Update l2geth

protolambda:
- Update L2 geth (`optimism-prototype` default branch now holds the new squashed commits after merging PRs https://github.com/ethereum-optimism/reference-optimistic-geth/pull/5 and https://github.com/ethereum-optimism/reference-optimistic-geth/pull/6), old commit can be found in `april4` branch of l2 geth repo)
- Fix bug in L1-info deposit tx type construction. We should be using the L2 block height for deposit tx uniquness. Otherwise L1-info deposits within an epoch will all have the same tx hash (seems nice, but a bad thing for tooling / db assumptions). See deposit tx type spec: [here](https://github.com/ethereum-optimism/optimistic-specs/blob/main/specs/deposits.md#the-deposited-transaction-type)


trianglesphere (approved):
lgtm.

I didn't realize that we needed the L2 block for the L1 info deposit tx. We definitely need to make them unique - maybe requiring the number inside the epoch would be good here as well.

PR #317, opened 2022-04-05
feat: DepositFeed receive function

maurelian:
Replaces #274 and adds: a new solidity test, a warning comment, and slither db update.


Inphi (commented):
Looks good. We should also have a test asserting that non-zero calldata is ignored when sending ETH via receive.

maurelian:
> asserting that non-zero calldata is ignored...
You mean the call succeeds but the data is not included in the event?
I'm pretty sure that sending any data at all would actually revert with `receive()`, and that we'd need to use `fallback()` instead to achieve that. But I think this is preferable.

Inphi (approved):

PR #318, opened 2022-04-06
init: integration tests

tuxcanfly:
WIP: Porting over existing integration tests


integration-tests/cache/solidity-files-cache.json
```diff
@@ -0,0 +1,1008 @@
+{
```
mslipper, line 1: This file should be ignored and removed.

integration-tests/test/rpc.spec.ts
```diff
@@ -0,0 +1,172 @@
...
+
+      await expect(factory.connect(wallet).deploy()).to.be.rejectedWith(
+        'gas required exceeds allowance'
+      )
+    })
+  })
+
+  describe('debug_traceTransaction', () => {
+    before(async function () {
+      if (!envConfig.RUN_DEBUG_TRACE_TESTS) {
```
mslipper, line 150: I think we can remove this flag, since we always want to run debug trace tests now.

integration-tests/test/shared/utils.ts
```diff
@@ -0,0 +1,169 @@
...
+  PRIVATE_KEY: str({
+    default:
+      '0xbf7604d9d3a1c7748642b1b7b05c2bd219c9faa91458b370f85e5a40f3b03af7',
+  }),
+  GAS_PRICE_ORACLE_PRIVATE_KEY: str({
+    default:
+      '0x8b3a350cf5c34c9194ca85829a2df0ec3153be0318b5e2d3348e872092edffba',
+  }),
+
+  DTL_ENQUEUE_CONFIRMATIONS: num({
```
mslipper, line 58: Remove

mslipper (changes requested):

integration-tests/test/rpc.spec.ts
```diff
@@ -0,0 +1,172 @@
...
+      const tx = {
+        ...defaultTransactionFactory(),
+        gasPrice: await gasPriceForL2(),
+        chainId: await env.l2Wallet.getChainId(),
+        data: '0x',
+        value: balance.add(ethers.utils.parseEther('1')),
+      }
+
+      await expect(env.l2Wallet.sendTransaction(tx)).to.be.rejectedWith(
+        'invalid transaction: insufficient funds for gas * price + value'
```
tynes, line 135: Some of these error messages may not be exact

integration-tests/test/shared/env.ts
```diff
@@ -0,0 +1,45 @@
...
+export class OptimismEnv {
+  // The wallets
+  l1Wallet: Wallet
+  l2Wallet: Wallet
+
+  // The providers
+  l1Provider: providers.JsonRpcProvider
+  l2Provider: providers.JsonRpcProvider
+  replicaProvider: providers.JsonRpcProvider
+  verifierProvider: providers.JsonRpcProvider
```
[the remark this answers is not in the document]
tynes: Do we want to be using the term "replica" or "verifier" throughout this codebase? I feel like we should prefer verifier over replica

integration-tests/test/shared/utils.ts
```diff
@@ -0,0 +1,169 @@
...
+    default: '0',
+  }),
+  L1_URL: str({ default: 'http://localhost:9545' }),
+  L1_POLLING_INTERVAL: num({ default: 10 }),
+
+  L2_CHAINID: num({ default: 17 }),
+  L2_GAS_PRICE: gasPriceValidator({
+    default: 'onchain',
+  }),
+  L2_URL: str({ default: 'http://localhost:8545' }),
```
tynes, line 39: I believe L2 is now at 9545 based on looking at the go based tests

integration-tests/test/shared/utils.ts
```diff
@@ -0,0 +1,169 @@
...
+
+export const gasPriceForL1 = async () => {
+  if (procEnv.L1_GAS_PRICE === ON_CHAIN_GAS_PRICE) {
+    return l1Wallet.getGasPrice()
+  }
+
+  return utils.parseUnits(procEnv.L1_GAS_PRICE, 'wei')
+}
+
+export const isHardhat = async () => {
```
tynes, line 157: We should be able to delete this

tsconfig.build.json
```diff
@@ -0,0 +1,23 @@
...
+    "target": "es2017",
+    "sourceMap": true,
+    "esModuleInterop": true,
+    "composite": true,
+    "resolveJsonModule": true,
+    "declaration": true,
+    "noImplicitAny": false,
+    "removeComments": true,
+    "noLib": false,
+    "emitDecoratorMetadata": true,
```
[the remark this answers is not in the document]
tynes: We have been doing the empty extend thing in the monorepo forever.. happy to have this cleaned up finally

integration-tests/contracts/OVMContext.sol
```diff
@@ -0,0 +1,53 @@
...
+
+/// @title OVMContext - Helper Functions
+contract OVMContext {
+    function getCurrentBlockTimestamp() public view returns (uint256 timestamp) {
+        timestamp = block.timestamp;
+    }
+
+    function getCurrentL1BlockNumber() public view returns (uint256) {
+        return iOVM_L1BlockNumber(
+            0x4200000000000000000000000000000000000013
```
tynes, line 38: I don't think this contract exists in the bedrock genesis state yet?

protolambda (changes requested):
Nice to see integration tests being ported over. Going to mark this as request-changes to avoid accidental merging though: we definitely don't want to check in `artifacts`, and please rebase/amend to prune out all that data before it is merged into `main`, otherwise we'll forever have to pull those old artifacts with every git clone (100,000+ lines of unnecessary git diff...)

integration-tests/contracts/OVMContext.sol
tuxcanfly: Will clean this up along with other OVM contracts.

integration-tests/hardhat.config.ts
```diff
@@ -0,0 +1,51 @@
...
+import '@nomiclabs/hardhat-waffle'
+import 'hardhat-gas-reporter'
+import { envConfig } from './test/shared/utils'
+
+const enableGasReport = !!process.env.ENABLE_GAS_REPORT
+
+const config: HardhatUserConfig = {
+  networks: {
+    optimism: {
+      url: process.env.L2_URL || 'http://localhost:8545',
```
tuxcanfly, line 14: This should default to `https://localhost:9545`, will update.

PR #320, opened 2022-04-06
contracts: Move withdrawal logic into abstract contract

maurelian:
**Description**

1. Moves the withdrawal verification logic out of the `OptimismPortal` into a new abstract contract `WithdrawalRelay`. 
2. Makes the `DepositFeed` abstract.
3. Moves abstract and library contracts into subdirectories.


packages/contracts/.gas-snapshot
```diff
@@ -1,12 +1,12 @@
...
+DepositFeedTest:test_depositTransaction_createWithZeroValueForContract() (gas: 20980)
 DepositFeedTest:test_depositTransaction_createWithZeroValueForEOA() (gas: 21467)
-DepositFeedTest:test_depositTransaction_withEthValueAndContractContractCreation() (gas: 28117)
-DepositFeedTest:test_depositTransaction_withEthValueAndEOAContractCreation() (gas: 25856)
-DepositFeedTest:test_depositTransaction_withEthValueFromContract() (gas: 27785)
-DepositFeedTest:test_depositTransaction_withEthValueFromEOA() (gas: 25817)
+DepositFeedTest:test_depositTransaction_withEthValueAndContractContractCreation() (gas: 28003)
+DepositFeedTest:test_depositTransaction_withEthValueAndEOAContractCreation() (gas: 25031)
+DepositFeedTest:test_depositTransaction_withEthValueFromContract() (gas: 27671)
+DepositFeedTest:test_depositTransaction_withEthValueFromEOA() (gas: 28407)
```
maurelian, line 9: I'm not sure why this test in particular got so much more expensive. Will dig into it, but for now I think we should just make sure we're checkpointing this so that we can begin to improve on it.

FWIW, the previous snapshot was pretty old, so the increases are necessarily associated with this PR.

maurelian: The call itself is still below 4500, so I think it's changes in the test file most likely:
```
Running 1 test for contracts/test/DepositFeed.t.sol:DepositFeedTest
[PASS] test_depositTransaction_withEthValueFromEOA() (gas: 28407)
Traces:
  [28407] DepositFeedTest::test_depositTransaction_withEthValueFromEOA()
    ├─ [0] VM::prank(DepositFeedTest: [0xb4c79dab8f259c7aee6e5b2aa729821864227e84], DepositFeedTest: [0xb4c79dab8f259c7aee6e5b2aa729821864227e84])
    │   └─ ← ()
    ├─ [0] VM::expectEmit(true, true, false, true)
    │   └─ ← ()
    ├─ emit TransactionDeposited(from: DepositFeedTest: [0xb4c79dab8f259c7aee6e5b2aa729821864227e84], to: 0x0000000000000000000000000000000000000001, mint: 100, value: 0, gasLimit: 50000, isCreation: false, data: 0x1111)
    ├─ [4449] Target::depositTransaction{value: 100}(0x0000000000000000000000000000000000000001, 0, 50000, false, 0x1111)
    │   ├─ emit TransactionDeposited(from: DepositFeedTest: [0xb4c79dab8f259c7aee6e5b2aa729821864227e84], to: 0x0000000000000000000000000000000000000001, mint: 100, value: 0, gasLimit: 50000, isCreation: false, data: 0x1111)
    │   └─ ← ()
    └─ ← ()
```

norswap:
Do we need to separate out the abstract contracts into a different directory? We have few enough contracts that I feel this makes navigation more unpleasant than it strictly needs to be. Same with the single library, which is already clearly highlighted with the `Lib` prefix. Certainly not a hill I'm dying on, but a small organizational preference.

maurelian:
> Do we need to separate out the abstract contracts into a different directory? We have few enough contracts that I feel this makes navigation more unpleasant than it strictly needs to be. Same with the single library, which is already clearly highlighted with the `Lib` prefix. Certainly not a hill I'm dying on, but a small organizational preference.

When auditing, I always found it very helpful to have some mechanism make it very clear which files contain the contracts that actually get deployed. The alternative would be to follow the `Lib_` prefix with an `Abs_`, but if you're not willing to die on the hill, I'm going to leave it as is.

packages/contracts/contracts/test/WithdrawalsRelay.t.sol
```diff
@@ -0,0 +1,179 @@
...
+    bytes32 withdrawerStorageRoot =
+        0xb8576230d94535779ec872748df80a094fcad002a8fc2b37c5b8fe250b384be6; // eth_getProof (storageHash)
+    bytes32 latestBlockhash = keccak256(abi.encode(2));
+
+    // This proof was generated by running
+    // make devnet-up
+    // Then in another terminal
+    // packages/contracts/test/withdrawer.spec.ts
+    // Invalid large internal hash
+    bytes withdrawalProof =
```
maurelian, line 84: I know this approach is far from ideal. Def worth doing with `--ffi` later.

maurelian (commented):
Added a few comments that will hopefully help clarify what's going on. It's unfortunate that the diff is so large, when it's almost entirely moving code around.

packages/contracts/contracts/test/WithdrawalsRelay.t.sol
```diff
@@ -0,0 +1,179 @@
...
+            sequencer
+        );
+        startingBlockTimestamp = block.timestamp;
+
+        wr = new Target(oracle, 7 days);
+    }
+
+    function setUp() external {
+        vm.warp(initTime);
+        bytes32 outputRoot = keccak256(
```
tynes, line 109: Ideally there is a library that can compute this hash, but I suppose ok for now

tynes (approved):

trianglesphere:
@mslipper this PR breaks your devent (did not notice b/c it was out of date with the base)

PR #321, opened 2022-04-06
ops: Fix devnet

mslipper:
- Fixes the devnet config to correctly generate rollup/genesis files.
- Fixes a bug that causes the rollup node to get stuck when there's a large timestamp delta between genesis and the current time, as is the case in the devnet.


ops/Dockerfile.opnode
```diff
@@ -15,6 +15,8 @@ RUN make opnode
 
 FROM alpine:3.15
 
+RUN apk add --no-cache curl jq
```
tynes, line 18: `--no-cache`

mslipper: We want no cache here, don't we?

opnode/rollup/driver/state.go
```diff
@@ -237,7 +237,7 @@ func (s *state) createNewL2Block(ctx context.Context) (eth.L1BlockRef, error) {
 		return eth.L1BlockRef{}, nil
 	}
 	// Don't produce blocks until past the L1 genesis
```
protolambda, line 239:
```suggestion
	// Don't produce blocks until at the rollup-genesis block of the L1 chain
	// note: deposits within this L1 block will be included in L2.
```
cc @trianglesphere for review. There are some interesting edge-cases here with deposits: we have to make sure to include those of the genesis block (and maybe everything before if we have a migration from previous system) are all included in L2. But based on timestamps the sequencer may try to build on the next L1 block before those deposits are processed. And the verifier will also need to do the same.

ops/rollup.json
```diff
@@ -11,11 +11,17 @@
...
 
-  "seq_window_size": 64,
+  "seq_window_size": 2,
 
-  "l1_chain_id": 900
+  "l1_chain_id": 900,
+
+  "fee_recipient_address":  "0xff00000000000000000000000000000000000002",
+
+  "batch_inbox_address":  "0xff00000000000000000000000000000000000002",
```
protolambda, line 24: It's fine if they are the same, but note that all L2 fees will effectively get burned by sending them to `0xff00000000000000000000000000000000000002`. That's fine for testing, but can't have that in production :sweat_smile: Copy-paste error or intentional?

mslipper: I meant to make it `0xff00000000000000000000000000000000000001` like in the E2E tests, but on second thought I'll make this a Hardhat default address so the fees can be recovered in the itests.

opnode/rollup/driver/state.go
trianglesphere: My understanding is that we don't want to actually read the L1 genesis when producing L2 blocks (and start after it). With this I guess it still works, but we have a weird thing where the L2 genesis should maybe point back one block.

I don't actually mind missing deposits prior to the start of the rollup

trianglesphere (approved):
I think that the new definition of L1 rollup start makes sense and we should change to it. It does have some knock on implications in how we handle the L2 start and specs definitions that we need to fix

opnode/rollup/driver/state.go
protolambda: Opened an issue to track the genesis related edge cases here: https://github.com/ethereum-optimism/optimistic-specs/issues/323

Think this PR is fine and we can address the deposit related edge-cases of genesis later.

protolambda (approved):

PR #324, opened 2022-04-06
tests: Add tests for batch requests when the parent request fails

mslipper:
This test makes sure that we do not retry when the parent request fails. We don't, but it wasn't immediately clear.

I also removed a duplicate assertion lib since we're using `testify` everywhere else.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/ethereum-optimism/optimistic-specs
 
-go 1.18
+go 1.17
 
 require (
 	github.com/ethereum/go-ethereum v1.10.16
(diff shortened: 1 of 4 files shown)


protolambda:
- go imports lint fixed
- test updated to check if error is contained in error chain, instead of hard equal check
- update to nicely wrap error with `%w` instead of `%v`
- added a comment about the error behavior. Batches as a whole should never fail unless the connection/rpc just stops working, in which case we should retry later as a caller. Only individual requests that fail are worth retrying I think. We can change this later if it turns out to be different.
- removed an old reference to "receipts": the batching is generic now, not only for receipts.

protolambda (approved):
LGTM

PR #326, opened 2022-04-06
opnode: update to latest L2 geth

protolambda:
Updates the L2 geth dependency. ~~The `go.mod` still has to point to the correct commit. Need to update the geth fork repo first.~~ Updated `go.mod` to point to the rebased commit of the geth fork. That commit is not on the default branch yet though.

The update adopts the Engine API as defined for Kiln testnet (except the authentication extension which we have yet to adopt).


go.mod
```diff
@@ -66,7 +68,7 @@ require (
 	gopkg.in/yaml.v3 v3.0.0-20210107192922-496545a6307b // indirect
 )
 
-replace github.com/ethereum/go-ethereum v1.10.16 => github.com/ethereum-optimism/reference-optimistic-geth v0.0.0-20220405004857-d5e1fc1a74bd
+//replace github.com/ethereum/go-ethereum v1.10.16 => github.com/ethereum-optimism/reference-optimistic-geth v0.0.0-20220405004857-d5e1fc1a74bd
 
 // For local debugging:
-// replace github.com/ethereum/go-ethereum v1.10.16 => ../reference-optimistic-geth
```
protolambda, line 72: CI will be failing until I've a git commit to point to. Not sure if I can point it to a non-default branch commit here

protolambda, line 72: CI will be failing until I've a git commit to point to. Not sure if I can point it to a non-default branch commit here

protolambda: commit is there, branch doesn't matter, updated mod file

trianglesphere (approved):
lgtm

PR #329, opened 2022-04-06
ci: Improve timestamp management in devnet

mslipper:
This change sets timestamps to the current timestamp on fresh devnet deploys. This avoids the devnet having to fill in a bunch of blocks on startup. The timestamp is preserved between runs; `make devnet-clean` will erase the timestamp data if you want to start anew.


tynes (approved):
Much better! Can confirm this make it much easier to develop locally

PR #330, opened 2022-04-07
specs: Readability improvements to rollup-node

maurelian:
Mostly minor edits that I think improve the flow.

### DIFF
--- a/specs/deposits.md
+++ b/specs/deposits.md
@@ -32,10 +32,10 @@ with the authorization and validation conditions on L2.
   - [L1 Attributes Depositor Account](#l1-attributes-depositor-account)
   - [L1 Attributes Predeployed Contract](#l1-attributes-predeployed-contract)
     - [L1 Attributes Predeployed Contract: Reference Implementation](#l1-attributes-predeployed-contract-reference-implementation)
-- [User-Deposited Transactions](#user-deposited-transactions)
+- [Deposited Transactions](#deposited-transactions)
   - [Deposit Contract](#deposit-contract)
     - [Address Aliasing](#address-aliasing)
-    - [Deposit Feed Contract: Reference Implementation](#deposit-feed-contract-reference-implementation)
+    - [Interface](#interface)
 
 <!-- END doctoc generated TOC please keep comment here to allow auto update -->
 
@@ -196,7 +196,7 @@ After running `yarn build` in the `packages/contracts` directory, the bytecode t
 file will be located in the `deployedBytecode` field of the build artifacts file at
 `/packages/contracts/artifacts/contracts/L2/L1Block.sol/L1Block.json`.
 
-## User-Deposited Transactions
+## Deposited Transactions
 
 [user-deposited]: #user-deposited-transactions
 
@@ -241,7 +241,7 @@ contract on L1 has the same address as a contract on L2 but doesn't have the sam
 for EOAs because they're guaranteed to have the same "code" (i.e. no code at all). This also makes
 it possible for users to interact with contracts on L2 even when the Sequencer is down.
 
-#### Deposit Feed Contract: Reference Implementation
+#### Interface
 
 A reference implementation of the Deposit Feed contract can be found in [DepositFeed.sol].
 
--- a/specs/glossary.md
+++ b/specs/glossary.md
@@ -19,6 +19,8 @@
   - [Sequencing window](#sequencing-window)
   - [Sequencing epoch](#sequencing-epoch)
   - [Sequencer batch](#sequencer-batch)
+  - [Sequencer inbox](#sequencer-inbox)
+  - [Sequencer bundle](#sequencer-bundle)
 - [Deposits](#deposits)
   - [Deposited Transaction](#deposited-transaction)
   - [L1 Attributes Deposited Transaction](#l1-attributes-deposited-transaction)
@@ -175,6 +177,20 @@ equal to the L1 block number of the first block in the window.
 
 A sequencer batch is list of L2 transactions tagged with an [`epoch`](#sequencing-epoch) and L2 block `timestamp`.
 Each L2 block can have one batch of transactions, an input for the block derivation.
+
+## Sequencer inbox
+
+[sequencer-inbox]: glossary.md#sequencer-inbox
+
+The sequencer inbox is an address on L1 with no code and no state. Sequencer batches are added when a sequencer submits
+them to this address as calldata.
+
+## Sequencer bundle
+
+[sequencer-bundle]: glossary.md#sequencer-bundle
+
+A sequencer bundle is a list of sequencer transactions which are submitted together in a single call to the sequencer
+inbox.
 
 ------------------------------------------------------------------------------------------------------------------------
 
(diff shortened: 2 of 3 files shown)


specs/deposits.md
```diff
@@ -196,7 +196,7 @@ After running `yarn build` in the `packages/contracts` directory, the bytecode t
 file will be located in the `deployedBytecode` field of the build artifacts file at
 `/packages/contracts/artifacts/contracts/L2/L1Block.sol/L1Block.json`.
 
-## User-Deposited Transactions
```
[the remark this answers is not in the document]
norswap: Yup! This is precisely defined in the glossary!
https://github.com/ethereum-optimism/optimistic-specs/blob/90e3620c126148c5c0f7023788c5d00d4025cbeb/specs/glossary.md#deposited-transaction

specs/rollup-node.md
```diff
@@ -92,40 +94,49 @@ Note that the sequencing windows overlap.
...
-      - The batch is the first batch with `batch.timestamp` in this sequencing window,
-        i.e. one batch per L2 block number.
-      - The batch only contains sequenced transactions, i.e. it must NOT contain any Deposit-type transactions.
+From the *first* block in the window only:
+
+- L1 block attributes:
+  - `blockNumber`
+  - `timestamp`
+  - `basefee`
+  - `random` (the output of the [`RANDOM` opcode][random])
```
norswap, line 103: The reason these were not in code case was that in my mind code case refers to "defined" terms.

They could be terms defined in the spec to be referred to later, or definitions in particular structures / code files.

It's not quite clear to me that this refers to a specific definition? If that's the case (e.g. the geth block structure) I think this is fine but we should link/mention the source of these definition.

specs/rollup-node.md
```diff
@@ -92,40 +94,49 @@ Note that the sequencing windows overlap.
...
+- Log entries emitted by the [Deposit Feed][[g-deposit-contract] contract, with the `TransactionDeposited` signature.
+- The `blockHeight` and `transactionIndex` of any transactions which emit `TransactionDeposited` logs (these values are
+  used to augmented the [deposited transactions][g-deposited] derived from them).
```
norswap, line 106: That's incorrect. The block height / transaction index here are from the L2 transaction index.

Agree the comment was confusing, I don't think we need to mention these values here at all.

specs/rollup-node.md
```diff
@@ -92,40 +94,49 @@ Note that the sequencing windows overlap.
...
+
+- Sequencer batches, derived from transactions that meet the following criteria:
+  - The transaction receiver (ie. `to`) is the [sequencer inbox] address.
+  - The transaction is signed (ie. `from`) by a recognized sequencer account.
+  - The calldata may contain a bundle of batches. *(calldata will be substituted with blob data in the future.)*
+
+- Batches must also meet the following filter conditions, or else will be ignored:
+  - `batch.epoch == sequencing_window.epoch`, i.e. for this sequencing window.
+  - `(batch.timestamp - genesis_l2_timestamp) % block_time == 0`, i.e. the timestamp matches with that of a valid l2
+    block.
```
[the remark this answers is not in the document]
norswap: Also change `block_time` to `l2_block_time` to match the usage in the next bullets.

Maybe move up the `l2_block_time` definition from below, though that's not super important, it should be pretty intuitive what the block time is.

specs/rollup-node.md
```diff
@@ -92,40 +94,49 @@ Note that the sequencing windows overlap.
...
+  - `min_l2_timestamp <= batch.timestamp < max_l2_timestamp`, i.e. the timestamp is within range. The parameters
+    are defined as follows:
+    - `min_l2_timestamp = prev_l2_timestamp + l2_block_time`, where:
+      - `prev_l2_timestamp` is the timestamp of the previous L2 block: the last block of the previous epoch,
+          or the L2 genesis block timestamp if there is no previous epoch.
+      - `l2_block_time` is a configurable parameter of the time between L2 blocks.
+    - `max_l2_timestamp = max(l1_timestamp + max_sequencer_drift, min_l2_timestamp + l2_block_time)`.
+      - `l1_timestamp` is the timestamp of the L1 block associated with the L2 block's epoch.
+      - `max_sequencer_drift` is the most a sequencer is allowed to get ahead of L1.
+  - The batch is the first batch with `batch.timestamp` in this sequencing window, i.e. one batch per L2 block number.
```
norswap, line 128: s/one batch per L2 block number/only one batch per L2 block will be considered, and subsequent batches for the same L2 block will be ignored./

specs/rollup-node.md
```diff
@@ -92,40 +94,49 @@ Note that the sequencing windows overlap.
...
+  - The batch only contains sequenced transactions, i.e. it must NOT contain any Deposit-type transactions.
```
norswap, line 129: Add a link for "Deposit-type transactions".

specs/rollup-node.md
```diff
@@ -196,7 +210,7 @@ Each of the derived `PayloadAttributes` starts with a L1 Attributes transaction.
 Like other derived deposits, this does not have to be batch-submitted, and exposes the required L1 information for the
 process of finding the sync starting point of the L2 chain, without requiring L2 state access.
 
-The [User-deposited] transactions are all put in the first of the derived `PayloadAttributes`,
+The [deposited][g-deposited] transactions are all put in the first of the derived `PayloadAttributes`,
```
norswap, line 213: Same comment as above, keep user-deposited. The capital letter can go.

specs/rollup-node.md
```diff
@@ -92,40 +94,49 @@ Note that the sequencing windows overlap.
...
-        i.e. one batch per L2 block number.
-      - The batch only contains sequenced transactions, i.e. it must NOT contain any Deposit-type transactions.
+From the *first* block in the window only:
+
+- L1 block attributes:
+  - `blockNumber`
+  - `timestamp`
+  - `basefee`
+  - `random` (the output of the [`RANDOM` opcode][random])
+- Log entries emitted by the [Deposit Feed][[g-deposit-contract] contract, with the `TransactionDeposited` signature.
```
norswap, line 104: Extraneous `[`

norswap (changes requested):
Ended up doing my review as loose comments.

Requested changes:
- revert to using "user-deposited" transactions
- clarify the glossary for bundle + batch
- open an issue that is "audit-critical" that we need to make sure the usage of "bundle" and "batch" is clear and consistent across the repo
- (bunch of small stuff)

specs/glossary.md
```diff
@@ -176,6 +178,20 @@ equal to the L1 block number of the first block in the window.
...
 Each L2 block can have one batch of transactions, an input for the block derivation.
 
+## Sequencer inbox
+
+[sequencer-inbox]: glossary.md#sequencer-inbox
+
+The sequencer inbox is an address on L1 with no code and no state. Sequencer batches are added when a sequencer submits
+them to this address as calldata.
+
+## Sequencer bundle
```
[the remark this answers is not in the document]
norswap: > Sequencer batch is still defined above on line 178.

Yes, but could we state explicitly say that the transaction in a batch are meant to be part of a single L2 block? It's *sort of* implied by the fact that the transactions are tagged with a L2 timestamp, but the implication is easy to miss.

> I came across this term in [rollup-node.md](https://github.com/ethereum-optimism/optimistic-specs/blob/84176052f2049ac9013c133658371504b8c3a3c7/specs/rollup-node.md#L127) and thought it should be listed here.

Yeah, this escaped my scrutiny. Still a valid concern though. Feel free to merge as-is if you feel this is out of scope, but could you then please open an issue to track this and assign it to me?

specs/rollup-node.md
norswap: What do we think about removing this? Imho this really doesn't belong here?

norswap (approved):
Approving, but please see the above unresolved issues:
- capitalization
- removing the mention of "augmented with fields" which is just awkward at this place in the spec and is also specified elsewehre
- better definining bundles vs batches (or add an issue to improve this later)

maurelian:
Closed because this is super old, and not that important.

PR #331, opened 2022-04-07
Fix address of the L1 Block Attributes contracts

maurelian:
```diff
- 0x4242424242424242424242424242424242424242
+ 0x4200000000000000000000000000000000000015
```

### DIFF
--- a/opnode/test/system_test.go
+++ b/opnode/test/system_test.go
@@ -98,7 +98,7 @@ func TestSystemE2E(t *testing.T) {
 		},
 		cliqueSigners:           []string{"m/44'/60'/0'/0/0"},
 		depositContractAddress:  "0xdeaddeaddeaddeaddeaddeaddeaddeaddead0001",
-		l1InforPredeployAddress: "0x4200000000000000000000000000000000000015",
+		l1InfoPredeployAddress: "0x4200000000000000000000000000000000000015",
 	}
 	// Create genesis & assign it to ethconfigs
 	initializeGenesis(cfg)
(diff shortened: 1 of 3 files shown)


mslipper (approved):

opnode/test/system_test.go
```diff
@@ -98,7 +98,7 @@ func TestSystemE2E(t *testing.T) {
 		},
 		cliqueSigners:           []string{"m/44'/60'/0'/0/0"},
 		depositContractAddress:  "0xdeaddeaddeaddeaddeaddeaddeaddeaddead0001",
-		l1InforPredeployAddress: "0x4242424242424242424242424242424242424242",
+		l1InforPredeployAddress: "0x4200000000000000000000000000000000000015",
```
protolambda, line 101: Infor? typo in setup type definition?

protolambda (commented):
looks good, but there's a typo in existing related test type, might as well fix that

opnode/test/system_test.go
```diff
@@ -96,9 +95,9 @@ func TestSystemE2E(t *testing.T) {
 			l2OutputHDPath:     10000000,
 			bssHDPath:          10000000,
 		},
-		cliqueSigners:           []string{"m/44'/60'/0'/0/0"},
-		depositContractAddress:  "0xdeaddeaddeaddeaddeaddeaddeaddeaddead0001",
-		l1InforPredeployAddress: "0x4242424242424242424242424242424242424242",
```
tynes, line 101: We should be importing the constants here

tynes (changes requested):
We need to be using the constants where we can so this doesn't happen again

tynes (approved):

PR #335, opened 2022-04-07
opnode: Fix L1WindowBuf management

trianglesphere:
**Description**
* opnode: Use l2 safe head instead of l2 head

Epochs should be inserted on top of the L2 Safe Head, not the
unsafe head. This was causing problems when a reorg would occur
and the node would start building in the wrong spot.

* opnode: Clear l1WindowBuf on insertEpoch failure

This is done to handle the case that the l1WindowBuf does not start
in the correct location (direct check is difficult with context).
The correct response to that is to clear the window and initialize
it from the correct block.


protolambda (approved):
LGTM

PR #341, opened 2022-04-07
opnode: use correct l1 info predeploy addr

tynes:
**Description**
Updates the predeploy address to match the specs
https://github.com/ethereum-optimism/optimistic-specs/blob/main/specs/deposits.md#l1-attributes-predeployed-contract


trianglesphere:
See https://github.com/ethereum-optimism/optimistic-specs/pull/331

PR #342, opened 2022-04-07
opnode: Fix possible panic in sequencer batch creation

trianglesphere:
Previously we assumed that every deposit would be included in the L2
block. However, the L2 execution engine removes depoists that are
not valid. As such the number of deposits submitted to the execution
engine could be greater than the number of transactions returned
causing a panic.

This fixes the issue by counting the number of deposits returned
and using that to slice the payload transactions for batch submission.
(Deposits are not submitted in batches and instead read directly from L1).

### DIFF
--- a/opnode/rollup/driver/step.go
+++ b/opnode/rollup/driver/step.go
@@ -26,6 +26,34 @@ type outputImpl struct {
 	Config rollup.Config
 }
 
+// isDepositTx checks an opaqueTx to determine if it is a Deposit Trransaction
+// It has to return an error in the case the transaction is empty
+func isDepositTx(opaqueTx l2.Data) (bool, error) {
+	if len(opaqueTx) == 0 {
+		return false, errors.New("empty transaction")
+	}
+	return opaqueTx[0] == types.DepositTxType, nil
+}
+
+// lastDeposit finds the index of last deposit at the start of the transactions.
+// It walks the transactions from the start until it finds a non-deposit tx.
+// An error is returned if any looked at transaction cannot be decoded
+func lastDeposit(txns []l2.Data) (int, error) {
+	var lastDeposit int
+	for i, tx := range txns {
+		deposit, err := isDepositTx(tx)
+		if err != nil {
+			return 0, fmt.Errorf("invalid transaction at idx %d", i)
+		}
+		if deposit {
+			lastDeposit = i
+		} else {
+			break
+		}
+	}
+	return lastDeposit, nil
+}
+
 func (d *outputImpl) createNewBlock(ctx context.Context, l2Head eth.L2BlockRef, l2SafeHead eth.BlockID, l2Finalized eth.BlockID, l1Origin eth.L1BlockRef) (eth.L2BlockRef, *derive.BatchData, error) {
 	d.log.Info("creating new block", "l2Head", l2Head)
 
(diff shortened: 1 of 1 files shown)


tynes:
Shouldn't this code be safe against adversarial inputs? It still seems like its possible to craft a payload that could cause a `panic`

opnode/rollup/driver/step.go
```diff
@@ -80,6 +78,15 @@ func (d *outputImpl) createNewBlock(ctx context.Context, l2Head eth.L2BlockRef,
 	if err != nil {
 		return l2Head, nil, fmt.Errorf("failed to extend L2 chain: %v", err)
 	}
+	depositStart := 1 // At least the L1 Info deposit
+	for i, tx := range payload.Transactions() {
+		if tx.Type() == types.DepositTxType {
+			depositStart = i + 1 // Want to exclude this deposit
+		} else {
+			// Once done with deposits, assume not more deposit transactions
+			break
```
protolambda, line 87: Running through the rest, and actually making sure there are no other deposit txs, and err if there are, would be nice. Then this bug can't happen again. (Also, we should add logging and/or metrics in the derivation function if it ever hits a deposit in a sequencer batch)

trianglesphere:
> Shouldn't this code be safe against adversarial inputs? It still seems like its possible to craft a payload that could cause a `panic`

How specifically is this still vulnerable? I do start it at 1, but that is because the rollup node sets the first transaction to be the L1 deposit info transaction. Payload is from the L2 execution engine, which is mostly trusted. The only malicious inputs are the deposits, which I think that this takes care of

tynes:
Sure we trust the L2 execution engine when there are no bugs, but a bug in the execution engine shouldn't cause a panic

trianglesphere:
> Sure we trust the L2 execution engine when there are no bugs, but a bug in the execution engine shouldn't cause a panic (out of bounds index)

Ok. I'll add a check that there is at least one deposit tx and remove the assumption that it returned transactions

opnode/rollup/driver/step.go
```diff
@@ -283,6 +291,32 @@ func (d *outputImpl) insertHeadBlock(ctx context.Context, fc l2.ForkchoiceState,
...
+		if tx.Type() == types.DepositTxType && doneWithDeposits {
+			d.log.Error("Produced an invalid block where the deposit txns are not all at the start of the block", "tx_idx", i, "tx", tx.Hash(), "lastDeposit", lastDeposit)
+			return nil, fmt.Errorf("deposit tx (%d:%s) after other tx in l2 block with prev deposit at idx %d", i, tx.Hash(), lastDeposit)
+		} else if tx.Type() == types.DepositTxType {
+			lastDeposit = i
+		} else {
+			doneWithDeposits = true
+		}
+	}
+	// TODO: Is this an error? This would likely happen on the verifier as well which means a chain halt.
```
protolambda, line 315: not all transactions may be deposits right?

opnode/rollup/driver/step.go
```diff
@@ -283,6 +291,32 @@ func (d *outputImpl) insertHeadBlock(ctx context.Context, fc l2.ForkchoiceState,
...
+	if len(payload.TransactionsField) == 0 {
+		return nil, errors.New("no transactions in returned payload")
+	}
+	txns := payload.Transactions()
+	if txns[0].Type() != types.DepositTxType {
+		return nil, fmt.Errorf("first transaction was not deposit tx. Got %v", txns[0].Type())
+	}
+	// Ensure that the deposits are first
+	var lastDeposit int
+	doneWithDeposits := false
```
protolambda, line 304: instead of adding the extra var it may be nice to just determine `lastDeposit` first (and maybe deduplicate that code with the above `createNewBlock` func) and then run a 2nd loop like `for i, tx := range txns[lastDeposit+1:] { ` to check that all remaining txs are not a deposit.

opnode/rollup/driver/step.go
trianglesphere: Right, this is most important on the sequencer side when creating an unsafe block. When inserting a safe block, it could have dropped a transaction. In general, I want to know when there is a mismatch between the attributes and payload

tynes (approved):

PR #344, opened 2022-04-07
init: integration tests

mslipper:


packages/integration-tests/test/rpc.spec.ts
```diff
@@ -0,0 +1,51 @@
...
+
+  before(async () => {
+    wallet = env.l2Wallet
+  })
+
+  it('eth_chainId', async () => {
+    const network = await env.l2Provider.getNetwork()
+    expect(network.chainId).to.equal(901)
+  })
+
```
tynes, line 23: nit: extra whitespace

packages/integration-tests/test/shared/constants.ts
```diff
@@ -0,0 +1 @@
+export const DEPOSIT_FEED_ADDR = '0xdeaddeaddeaddeaddeaddeaddeaddeaddead0001'
```
tynes, line 1: We are going to need common js stuff soon, its going to be scattered all over the place again

packages/integration-tests/test/shared/utils.ts
```diff
@@ -0,0 +1,11 @@
+import { BigNumber } from 'ethers'
+
+export const defaultTransactionFactory = () => {
```
tynes, line 3: I feel like this should ideally take each of the possible args as input in an object and then assign those args if they are passed in, otherwise use the default values, will clean up a bunch of code

packages/integration-tests/tsconfig.json
```diff
@@ -0,0 +1,19 @@
...
+    "strict": true,
+    "esModuleInterop": true,
+    "outDir": "dist",
+    "declaration": true,
+    "resolveJsonModule": true
+  },
+  "include": [
+    "./test",
+    "sync-tests/*.ts",
+    "./actor-tests/**/*.ts",
```
tynes, line 14: This looks like it has some dead code in it

packages/integration-tests/tsconfig.json
```diff
@@ -0,0 +1,19 @@
+{
+  "compilerOptions": {
+    "target": "es2018",
+    "module": "commonjs",
+    "strict": true,
+    "esModuleInterop": true,
```
tynes, line 6: Could this extend the other `tsconfig`?

packages/integration-tests/contracts/ERC20.sol
```diff
@@ -0,0 +1,72 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.9;
+
+contract ERC20 {
```
tynes, line 4: We could technically delete this and just import the oz erc20 JS npm package if we wanted to

packages/integration-tests/test/shared/constants.ts
mslipper: Yeah. I think we should do this once we've stabilized and moved everything back into the monorepo.

packages/integration-tests/tsconfig.json
mslipper: It could, but I'd rather do this once we move back to the monorepo.

packages/integration-tests/contracts/ERC20.sol
mslipper: Can we compile the NPM package with Forge without having to resort to more submodules?

tynes: `forge` has a concept of remappings in `foundry.toml` where you say the `name-of-import=path-to-file`

There is also the `--hardhat` flag

tynes: Not the biggest deal, but an easy way to remove code

tynes:
Untested locally but looks good to me

Inphi (approved):

packages/integration-tests/package.json
```diff
@@ -0,0 +1,70 @@
+{
+  "private": true,
+  "name": "@eth-optimism/integration-tests",
+  "version": "0.5.9",
+  "description": "[Optimism] Integration tests",
+  "scripts": {
+    "lint": "yarn lint:fix && yarn lint:check",
+    "lint:fix": "yarn lint:check --fix",
+    "lint:check": "eslint . --max-warnings=0",
+    "build:contracts": "forge build --root . -c contracts -o artifacts",
```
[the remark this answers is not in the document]
mslipper: Will take care of in the next PR.

PR #345, opened 2022-04-07
opnode: Fix bug in block creation

trianglesphere:
The NoTxPool flag was incorrectly set to falses in the InsertEpoch
function.


protolambda (approved):
LGTM, nice catch!

tynes (approved):

PR #347, opened 2022-04-07
feat: js deposit tx

tynes:
**Description**

- JS impl of `DepositTx` that can serialize + hash, simple test now with more tests later
- Hardhat task now lists the tx hash

### DIFF
--- a/packages/contracts/helpers/index.ts
+++ b/packages/contracts/helpers/index.ts
@@ -1,0 +1,67 @@
+import * as RLP from '@ethersproject/rlp'
+import { BigNumber, BigNumberish } from '@ethersproject/bignumber'
+import { getAddress } from '@ethersproject/address'
+import { hexConcat, stripZeros } from '@ethersproject/bytes'
+import { keccak256 } from '@ethersproject/keccak256'
+
+function formatNumber(value: BigNumberish, name: string): Uint8Array {
+  const result = stripZeros(BigNumber.from(value).toHexString())
+  if (result.length > 32) {
+    throw new Error(`invalid length for ${name}`)
+  }
+  return result
+}
+
+interface DepositTxOpts {
+  blockHeight: BigNumberish
+  transactionIndex: BigNumberish
+  from: string
+  to: string | null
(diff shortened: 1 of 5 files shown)


packages/contracts/tasks/deposits.ts
```diff
@@ -52,19 +70,51 @@ task('deposit', 'Deposits funds onto L2.')
...
-    console.log('Done.')
+    const receipt = await tx.wait()
+
+    const event = receipt.events[0]
+    if (event?.event !== 'TransactionDeposited') {
+      throw new Error('Transaction not deposited')
+    }
+
+    let found = false
+    l2Provider.on('block', async (number) => {
```
Inphi, line 95: To make this a little bit more robust, we should handle the case where a new block with the deposit is produced on l2 immediately after the TransasctionDeposited event on l1.

tynes: I agree with you that this is prone to race conditions, once https://github.com/ethereum-optimism/optimistic-specs/issues/348 is fixed then this will no longer be a problem and I'll remove this brute force check code

Inphi: Alright. Let's add a TODO here referencing the other issue so we don't forget.

packages/contracts/helpers/index.ts
```diff
@@ -0,0 +1,67 @@
...
+  const result = stripZeros(BigNumber.from(value).toHexString())
+  if (result.length > 32) {
+    throw new Error(`invalid length for ${name}`)
+  }
+  return result
+}
+
+interface DepositTxOpts {
+  blockHeight: BigNumberish
+  transactionIndex: BigNumberish
```
protolambda, line 17: don't forget we'll need to update this after #351

tynes:
New PR https://github.com/ethereum-optimism/optimistic-specs/pull/352

Will fix the merge conflict after https://github.com/ethereum-optimism/optimistic-specs/pull/351 is merged

protolambda:
Rebased #352, let's continue there.

PR #349, opened 2022-04-07
opnode: Remove panic in ExecutionPayload unmarshal

mslipper:


opnode/l2/api_test.go
```diff
@@ -0,0 +1,52 @@
...
+	require.Equal(t, "0x7", payload.BaseFeePerGas.String())
+	require.Equal(t, "0x7a0bb12e6a2c78a6996252fd691529f285f721300abe5da650f977dfcbbabf75", payload.BlockHash.String())
+	require.Equal(t, 1, len(txs))
+	require.EqualValues(t, 1, txs[0].BlockHeight())
+	require.Equal(t, &derive.L1InfoPredeployAddr, txs[0].To())
+	require.Equal(t, new(big.Int), txs[0].Mint())
+	require.EqualValues(t, 99999999, txs[0].Gas())
+	require.Equal(t, common.FromHex("0x8c89c386000000000000000100000000624f6bca00000000000000000000000000000000000000000000000000000000342770c019e19e8b3faa3b597eeb94c2f88b0d484b77c75595d80a8239447b0c2a3fc593"), txs[0].Data())
+
+	f.Close()
```
tynes, line 43: Can't defer this above?

mslipper: Same variable references two different files, so I need to close it here.

opnode/l2/api.go
```diff
@@ -137,15 +140,54 @@ func (payload *ExecutionPayload) ParentHash() common.Hash {
 }
 
 func (payload *ExecutionPayload) Transactions() types.Transactions {
-	res := make([]*types.Transaction, len(payload.TransactionsField))
-	for i, t := range payload.TransactionsField {
-		res[i] = new(types.Transaction)
-		err := res[i].UnmarshalBinary(t)
-		if err != nil {
-			panic(err)
+	return payload.unmarshaledTxs
```
protolambda, line 143: As soon as we change the `TransactionsField` this gets out of sync. We use this type not just to read from L1, but also to *write* to L2. So it seems prone to bugs, and I don't particularly like the duplication of all JSON things either.

protolambda (changes requested):
Current panic should never hit unless the L2 node is in a bad state already. Whereas the new code is prone to consistency bugs when we change the Transaction field contents. Can we do this differently, e.g. by just adding an error argument to the block interface? (and creating a tiny wrapper type that embeds the `*types.Block` in the type but overrides just the Transactions method with a nil error in return?)

mslipper:
Closing since @protolambda has an alternative solution.

PR #351, opened 2022-04-08
Better unique deposit tx hashes

### LINKED ISSUE #348: Computing the L2 Deposit Tx Hash Client Side
**Is your feature request related to a problem? Please describe.**
When a user sends a L1 -> L2 transaction, the user wants to know what the L2 transaction hash is so that they can know when the deposit goes through. Right now, the `DepositTx` has this schema:

```ts
interface DepositTxOpts {
  blockHeight: BigNumberish
  transactionIndex: BigNumberish
  from: string
  to: string | null
  mint: BigNumberish
  value: BigNumberish
  gas: BigNumberish
  data: string
}
```

This is problematic because the L2 blockheight and L2 transactionIndex are included in the transaction. This means that the user needs to brute force the L2 tx hash to know when it goes through on L2. See the hardhat task changes here: https://github.com/ethereum-optimism/optimistic-specs/pull/347/files for how this would look in the sdk

**Describe the solution you'd like**
I'd like to be able to compute the L2 tx hash locally just using the receipt of the L1 transaction. This could be accomplished by changing the struct of the `TxDeposit` to:

```ts
interface DepositTxOpts {
  sourceHash: string
  from: string
  to: string | null
  mint: BigNumberish
  value: BigNumberish
  gas: BigNumberish
  data: string
}
```
Where `sourceHash` is `keccack256(receipt.blockHash || receipt.txIndex || receipt.eventIndex)`.
The `sourceHash` is required to ensure that transaction hashes are unique on L2.
The event index is required because a single L1 transaction can create multiple L2 transactions by calling the deposit contract multiple times.

protolambda:
This implements better unique tx hashes:
- User deposit tx hashes can uniquely be identified based on their L1 event log info and inclusion block hash.
- L1 info attribute deposit tx hashes (derived by system) can uniquely be identified by their L1 block hash, and seq number (heigh diff since start of epoch on L2). (simply use 0 if you want the first deposit of some specific L1 block hash)

This updates the `go.mod` to use the commit from https://github.com/ethereum-optimism/reference-optimistic-geth/pull/9

If this PR is approved I'd like to merge the above PR first by squashing it into the main commit (to keep a 1-commit diff in geth repo), and then updating the `go.mod` here.

Fix #348


opnode/l1/receipts.go
```diff
@@ -33,10 +35,41 @@ func fetchReceipts(ctx context.Context, receiptHash common.Hash, txs types.Trans
...
+		if r.TransactionIndex != uint(i) {
+			return nil, fmt.Errorf("receipt %d has unexpected tx index %d", i, r.TransactionIndex)
+		}
+		if r.BlockNumber.Uint64() != block.Number {
+			return nil, fmt.Errorf("receipt %d has unexpected block number %d, expected %d", i, r.BlockNumber, block.Number)
+		}
+		if r.BlockHash != block.Hash {
+			return nil, fmt.Errorf("receipt %d has unexpected block hash %s, expected %s", i, r.BlockHash, block.Hash)
+		}
+		for j, log := range r.Logs {
```
tynes, line 55: https://github.com/ethereum/go-ethereum/blob/d6f49bf764c48e9415b168e701b6dc26566a675c/core/types/log.go#L55

tynes (approved):

protolambda:
Added a short test section to e2e to fetch the L2 deposit tx receipt by deriving the l2 tx hash from L1 deposit receipt information.

protolambda:
Updated the `optimism-prototype` branch on geth repo, and updated the go commit in mod file + updated `go.sum` too. Ready to merge.

tynes (approved):

PR #352, opened 2022-04-08
feat: deposit tx js helpers

tynes:
**Description**
Updates the deposit UX to work with:
- https://github.com/ethereum-optimism/reference-optimistic-geth/pull/9
- https://github.com/ethereum-optimism/optimistic-specs/pull/351

Can confirm that it works and the UX is a lot better

### DIFF
--- a/packages/contracts/tasks/deposits.ts
+++ b/packages/contracts/tasks/deposits.ts
@@ -1,14 +1,25 @@
 import { task, types } from 'hardhat/config'
 import { Contract, providers, utils, Wallet } from 'ethers'
 import dotenv from 'dotenv'
+import { DepositTx } from '../helpers/index'
 
 dotenv.config()
+
+async function sleep(ms: number) {
+  return new Promise((resolve) => setTimeout(resolve, ms))
+}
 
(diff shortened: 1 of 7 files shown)


packages/contracts/tasks/deposits.ts
```diff
@@ -52,19 +70,45 @@ task('deposit', 'Deposits funds onto L2.')
...
     console.log(`Got TX hash ${tx.hash}. Waiting...`)
-    await tx.wait()
-    console.log('Done.')
+    const receipt = await tx.wait()
+
+    const event = receipt.events[0]
+    if (event?.event !== 'TransactionDeposited') {
+      throw new Error('Transaction not deposited')
+    }
+
```
tynes, line 91: I think the best devex from a user perspective would be something like:

```js
const l2Tx = DepositTx.fromReceipt(receipt)
```
Then it could handle the boilerplate

protolambda: Maybe call it `fromL1Receipt` to avoid confusion, since deposits have a receipt on both L1 and L2

mslipper (approved):

PR #353, opened 2022-04-08
[REVIEW] Rollup Package

norswap:
This PR contains only the files in `opnode/rollup`, for review of that part of the codebase.

My initial review will focus on the sync algorithm, but feel free to use for other rollup functions.

### DIFF
--- a/opnode/rollup/sync/start.go
+++ b/opnode/rollup/sync/start.go
@@ -1,0 +1,157 @@
+// The sync package is responsible for reconciling L1 and L2.
+//
+//
+// The ethereum chain is a DAG of blocks with the root block being the genesis block.
+// At any given time, the head (or tip) of the chain can change if an offshoot of the chain
+// has a higher number. This is known as a re-organization of the canonical chain.
+// Each block points to a parent block and the node is responsible for deciding which block is the head
+// and thus the mapping from block number to canonical block.
+//
+// The optimism chain has similar properties, but also retains references to the ethereum chain.
+// Each optimism block retains a reference to an L1 block and to its parent L2 block.
+// The L2 chain node must satisfy the following validity rules
+//     1. l2block.height == l2parent.block.height + 1
+//     2. l2block.l1Origin.height >= l2block.l2parent.l1Origin.height
+//     3. l2block.l1Origin is in the canonical chain on L1
+//     4. l1_rollup_genesis is an ancestor of l2block.l1Origin
+//
+//
+// During normal operation, both the L1 and L2 canonical chains can change, due to a reorg
+// or an extension (new block).
+//     - L1 reorg
+//     - L1 extension
+//     - L2 reorg
+//     - L2 extension
+//
+// When one of these changes occurs, the rollup node needs to determine what the new L2 Heads should be.
+// In a simple extension case, the L2 head remains the same, but in the case of a re-org on L1, it needs
+// to find the unsafe and safe blocks.
+//
+// Unsafe Block: The highest L2 block. If the L1 Attributes is ahead of the L1 head, it is assumed to be valid,
+// if not, it walks back until it finds the first L2 block whose L1 Origin is canonical in the L1 chain.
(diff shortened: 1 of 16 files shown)


opnode/rollup/sync/start.go
```diff
@@ -0,0 +1,157 @@
...
+
+		}
+		// Genesis is always safe.
+		if n.Hash == genesis.L2.Hash || n.Number == genesis.L2.Number {
+			safe = eth.L2BlockRef{Hash: genesis.L2.Hash, Number: genesis.L2.Number, Time: genesis.L2Time, L1Origin: genesis.L1}
+			if l2Ahead {
+				return start, safe, nil
+			} else {
+				return latest, safe, nil
+			}
```
norswap, line 147: This is where it is all coming apart from me.

Two scenarios:
1. L1 origin of `start` is ahead of known L1 head. (`l2Ahead == true`)
2. L1 origin of `start` is behind of known L1 head. (`l2Ahead == false`)

In (1), because of the checks earlier in this file (line 99), we know that all L1 origins in the L2 chain are either ahead of the L1 head (\*), or are canonical L1 blocks.

(\*) They are "plausibly" ahead, meaning that there is an extra check (performed when we learn of new L2 blocks) to ensure that (the hash of) the L2 block's L1 origin is not canonical at a height different than the L2 block epoch in the known L1 chain. (This check should probably be mentionned somewhere in this file btw.)

In (2), because again because of line 99's canonicality check, it should always be the case that `start == latest`!
It's the logical combination of:
- The number of the L1 origin of `start` is below the head of the L1 chain (`l2Ahead == false`).
- All L1 origins in the L2 chain are ahead or canonical (line 99 check).

(NB: the `latest` in this code is defined as the highest L2 block such that its L1 block is canonical)

**Conclusions**

As far as I can tell, the behaviour of this algorithm is as follows:
- It validates `start` as the unsafe head against the presence of a L1 reorg (presence of a non-canonical L1 origin).
- If an L1 reorg occurs, it bails.
- It does correctly find the L2 safe block (i.e. the highest L2 block whose epoch's sequencing window is complete in the known L1 chain). Again, assuming no L1 reorg.

This was a bit of a shock, because my prior going into this code was that it was used to "resync" the L2 safe and unsafe block after a reorg was detected. This means that either:
- my understanding is incorrect, and we need to document this stuff better
- this code is incorrect

This needs more investigation, but a quick look at `state.go` where this is called from:
```
	// New L1 Head is not the same as the current head or a single step linear extension.
	// This could either be a long L1 extension, or a reorg. Both can be handled the same way.
	s.log.Warn("L1 Head signal indicates an L1 re-org", "old_l1_head", s.l1Head, "new_l1_head_parent", newL1Head.ParentHash, "new_l1_head", newL1Head)
	unsafeL2Head, safeL2Head, err := sync.FindL2Heads(ctx, s.l2Head, s.Config.SeqWindowSize, s.l1, s.l2, &s.Config.Genesis)
	if err != nil {
		s.log.Error("Could not get new unsafe L2 head when trying to handle a re-org", "err", err)
		return err
	}
```

It looks like a L1 reorg (assuming we've built L2 blocks on some of the L1 block re-orged away) will always result in an error here. Of course, using the old L2 head, we will encounter L1 origins that are non canonical since L1 re-orged!

norswap: Followup question to sanity check all of this: do we test L1 reorgs anywhere? Doesn't seem to be the case in `start_test.go`.

norswap, line 147: A bit dubious about this. What about the scenario where `l2ahead == true` (L1 origin of `start` is ahead of know L1 head) and there is a block known to be non-canonical block in the L2 chain?

Example: L1 head is at block `X`.  `start` has an L1 origin with number `X + 1` but its parent L2 block has an L1 origin with number `X` whose blockhash is not canonical. In this case, the algorithm still returns `start` as the unsafe L1 head, which is incorrect.

PR #354, opened 2022-04-08
opnode: don't expose types.Block to driver package, prefer l2.ExecutionPayload

protolambda:
Previously there were many issues with encoding/decoding transactions:
- first it would fetch the block
- then it would decode the txs
- then it would encode the txs back, and hash them, to compare the equality of attributes
- then it would compare them
- and then it derived the L1 info from the first block

Instead, we should hide all this from the driver, and expose methods that do the most minimal thing:
- Provide execution payloads when necessary to compare against attributes (on reorgs)
- Provide L2 block ref without doing all tx decoding/encoding work, we only need the first
- Provide a method to compute the L2-block ref from a payload, without fuly decoding the payload. We don't need to go back/forth to fully parsed txs when we're dealing with a payload that we produced with the engine api

And now that it's more encapsulated, I hope we can optimize the L2 engine api client with caching (avoid repeated requests) and more light-weight json parsing (we're still fetching decoded block data from the RPC, there should be a way to fetch a block with the raw transaction data, to avoid lots of unnecessary work) 

This also gets rid of the transactions decoding panic, and removes the need for an interface that wraps both `types.Block` and `l2.ExecutionPayload`

### DIFF
--- a/opnode/l2/source.go
+++ b/opnode/l2/source.go
@@ -36,12 +36,30 @@ func (s *Source) Close() {
 	s.rpc.Close()
 }
 
-func (s *Source) BlockByHash(ctx context.Context, hash common.Hash) (*types.Block, error) {
-	return s.client.BlockByHash(ctx, hash)
+func (s *Source) PayloadByHash(ctx context.Context, hash common.Hash) (*ExecutionPayload, error) {
+	// TODO: we really do not need to parse every single tx and block detail, keeping transactions encoded is faster.
+	block, err := s.client.BlockByHash(ctx, hash)
+	if err != nil {
+		return nil, fmt.Errorf("failed to retrieve L2 block by hash: %v", err)
+	}
+	payload, err := BlockAsPayload(block)
+	if err != nil {
+		return nil, fmt.Errorf("failed to read L2 block as payload: %w", err)
+	}
+	return payload, nil
 }
 
-func (s *Source) BlockByNumber(ctx context.Context, number *big.Int) (*types.Block, error) {
-	return s.client.BlockByNumber(ctx, number)
+func (s *Source) PayloadByNumber(ctx context.Context, number *big.Int) (*ExecutionPayload, error) {
+	// TODO: we really do not need to parse every single tx and block detail, keeping transactions encoded is faster.
+	block, err := s.client.BlockByNumber(ctx, number)
+	if err != nil {
+		return nil, fmt.Errorf("failed to retrieve L2 block by number: %v", err)
+	}
+	payload, err := BlockAsPayload(block)
+	if err != nil {
+		return nil, fmt.Errorf("failed to read L2 block as payload: %w", err)
+	}
+	return payload, nil
 }
 
 // ForkchoiceUpdate updates the forkchoice on the execution client. If attributes is not nil, the engine client will also begin building a block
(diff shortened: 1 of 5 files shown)


mslipper (approved):

opnode/l2/source.go
```diff
@@ -157,5 +175,53 @@ func (s *Source) L2BlockRefByHash(ctx context.Context, l2Hash common.Hash) (eth.
...
 		return eth.L2BlockRef{}, fmt.Errorf("failed to determine block-hash of height %v, could not get header: %w", l2Hash, err)
 	}
-	return derive.BlockReferences(block, s.genesis)
+	return blockToBlockRef(block)
+}
+
+func blockToBlockRef(block *types.Block) (eth.L2BlockRef, error) {
+	txs := block.Transactions()
+	if len(txs) == 0 {
+		return eth.L2BlockRef{}, fmt.Errorf("l2 block is missing L1 info deposit tx, block hash: %s", block.Hash())
```
trianglesphere, line 184: This won't work on the genesis block. That's why the rollup genesis was always being passed around

opnode/rollup/driver/step.go
```diff
@@ -110,7 +111,7 @@ func (d *outputImpl) insertEpoch(ctx context.Context, l2Head eth.L2BlockRef, l2S
 	epoch := rollup.Epoch(l1Input[0].Number)
 	fetchCtx, cancel := context.WithTimeout(ctx, time.Second*20)
 	defer cancel()
-	l2Info, err := d.l2.BlockByHash(fetchCtx, l2SafeHead.Hash)
+	l2Info, err := d.l2.PayloadByHash(fetchCtx, l2SafeHead.Hash)
```
trianglesphere, line 114: I don't see the difference between using a `Block` and `Payload` here.

opnode/rollup/driver/step.go
```diff
@@ -212,59 +213,51 @@ func (d *outputImpl) insertEpoch(ctx context.Context, l2Head eth.L2BlockRef, l2S
...
-	for i := range attrs.Transactions {
-		var tx types.Transaction
-		err := tx.UnmarshalBinary(attrs.Transactions[i])
-		if err != nil {
-			return fmt.Errorf("failed to decode transaction %d in attributes: %w", i, err)
-		}
-
-		if tx.Hash() != btxs[i].Hash() {
-			return fmt.Errorf("transaction %d does not match. expected: %v. got: %v", i, tx.Hash(), btxs[i].Hash())
+	for i, otx := range attrs.Transactions {
```
trianglesphere, line 229: Nice here. The previous version was also hurt by the fact that the L2 EE was returning the wrong values

trianglesphere (commented):
This works. I appreciate removing the panic from the Payload Functions. I don't see a huge reason to switch to L2 Payloads in the driver. The change that we need to make this work is to be able to derive block refs from the execution payload.

opnode/l2/source.go
```diff
@@ -157,5 +175,53 @@ func (s *Source) L2BlockRefByHash(ctx context.Context, l2Hash common.Hash) (eth.
 		// w%: wrap the error, we still need to detect if a canonical block is not found, a.k.a. end of chain.
 		return eth.L2BlockRef{}, fmt.Errorf("failed to determine block-hash of height %v, could not get header: %w", l2Hash, err)
 	}
-	return derive.BlockReferences(block, s.genesis)
+	return blockToBlockRef(block)
+}
+
```
tynes, line 180: A lot of the other functions in this file have docstrings

opnode/l2/source.go
protolambda: oh right, forgot about that, good catch

PR #355, opened 2022-04-08
Fix sync-start bug + improve sync-start comments

norswap:
This fixes the bug outlined here: https://github.com/ethereum-optimism/optimistic-specs/pull/353#discussion_r846131762, and adds a lot of comments + improves code structure to make the algorithm easier to understand.

### DIFF
--- a/opnode/rollup/sync/start.go
+++ b/opnode/rollup/sync/start.go
@@ -1,15 +1,16 @@
 // The sync package is responsible for reconciling L1 and L2.
 //
-// The ethereum chain is a DAG of blocks with the root block being the genesis block. At any given
-// time, the head (or tip) of the chain can change if an offshoot of the chain has a higher number.
-// This is known as a re-organization of the canonical chain. Each block points to a parent block
-// and the node is responsible for deciding which block is the head and thus the mapping from block
-// number to canonical block.
+// The Ethereum chain is a DAG of blocks with the root block being the genesis block. At any given
+// time, the head (or tip) of the chain can change if an offshoot/branch of the chain has a higher
+// number. This is known as a re-organization of the canonical chain. Each block points to a parent
+// block and the node is responsible for deciding which block is the head and thus the mapping from
+// block number to canonical block.
 //
-// The Optimism chain has similar properties, but also retains references to the ethereum chain.
-// Each Optimism block retains a reference to an L1 block (its "L1 origin", i.e. L1 block associated
-// with the epoch that the L2 block belongs to) and to its parent L2 block. The L2 chain node must
-// satisfy the following validity rules:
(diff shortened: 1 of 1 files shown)


norswap:
Still needs a regression test.

opnode/rollup/sync/start.go
```diff
@@ -1,47 +1,45 @@
 // The sync package is responsible for reconciling L1 and L2.
 //
+// The Ethereum chain is a DAG of blocks with the root block being the genesis block. At any given
+// time, the head (or tip) of the chain can change if an offshoot/branch of the chain has a higher
```
tynes, line 4: nit: its higher total difficulty, not number. see https://github.com/ethereum/go-ethereum/blob/6c3fea0fc9db561aafd9ec017980d0ac87516b87/core/forkchoice.go#L77

opnode/rollup/sync/start.go
```diff
@@ -62,96 +60,139 @@ var TooDeepReorgErr = errors.New("reorg is too deep")
...
-	if err != nil && !errors.Is(err, ethereum.NotFound) {
-		return false, err
-	} else if err != nil {
-		return false, nil
+// isCanonical returns the following values:
+// - `aheadOrCanonical: true if the supplied block is ahead of the known head of the L1 chain,
+//    or canonical in the L1 chain.
+// - `canonical`: true if the block is canonical in the L1 chain.
+func isAheadOrCanonical(ctx context.Context, l1 L1Chain, block eth.BlockID) (aheadOrCanonical bool, canonical bool, err error) {
+	if l1Head, err := l1.L1HeadBlockRef(ctx); err != nil {
```
trianglesphere, line 68: I don't love that we call `L1HeadBlockRef` on every iteration. I think it could probably be called once, but it may have an interaction with `BlockRefByNumber`

What could happen is that the block could be ahead of the head at one point, and then at a later point turn out to be canonical - as such I think it needs to do this. Still a bit weird to have the L1 head maybe updating underneath us in this look.

trianglesphere (approved):
Nice work. I really like the comments on the algorithm.

I think this is getting close enough to be correct even during malicious unsafe block distribution.

opnode/rollup/sync/start.go
norswap: I briefly considered using the "block not found" error as a signal that the block was ahead, but it feels a bit dirty.

I think we could get away with doing it once — both ahead blocks & canonical blocks can be the unsafe head. This might lead us to rollback too far for the safe head, but I think that's generally a safe thing to do (lol).

(Note that the block being ahead then canonical could happen in the current function too! Could be good to include a comment that explains we considered this pitfall.)

I'm not sure we should change it though. I think `l1.L1HeadBlockRef(ctx)` should always return a cached value (and if not, we should fix that), so the overhead should be neglectible.

PR #357, opened 2022-04-08
specs: Update output root derivation

maurelian:
Updates the specs to match the current implementation of the output root, ie. 

`keccak256(version, stateRoot, withdrawerStorageRoot, latestBlockhash)`

### DIFF
--- a/specs/proposals.md
+++ b/specs/proposals.md
@@ -3,6 +3,7 @@
 <!-- All glossary references in this file. -->
 
 [g-rollup-node]: glossary.md#rollup-node
+[g-mpt]: glossary.md#merkle-patricia-trie
 
 <!-- START doctoc generated TOC please keep comment here to allow auto update -->
 <!-- DON'T EDIT THIS SECTION, INSTEAD RE-RUN doctoc TO UPDATE -->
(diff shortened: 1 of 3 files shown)


specs/proposals.md
```diff
@@ -47,43 +49,31 @@ output, and the appropriate timestamp, to the [L2 Output Commitment](#l2-output-
 
 ## L2 Output Commitment Construction
 
-This merkle-structure is defined with [SSZ], a type system for merkleization and serialization, used in
-L1 (beacon-chain). However, we replace `sha256` with `keccak256` to save gas costs in the EVM.
-
-[SSZ]: https://github.com/ethereum/consensus-specs/blob/dev/ssz/simple-serialize.md
+The `output_root` is a 32 byte string defined as:
```
tynes, line 53: I think this delineate between versions where what is defined below is version 0.

The top level is `keccak256(version || payload)` and in version 0 the `payload` is `state_root || withdrawal_storage_root || latest_block`

tynes: This allows the version to meaningfully change the definition of the commitment

specs/proposals.md
```diff
@@ -47,43 +49,31 @@ output, and the appropriate timestamp, to the [L2 Output Commitment](#l2-output-
...
 
-The `history_accumulator_root` is a reserved field, elevating a storage variable of the L2 chain that maintains
-the [SSZ] merkle root of an append-only `List[Bytes32, MAX_ITEM_COUNT]` (`keccak256` [SSZ] hash-tree-root),
-where each item is defined as `keccak256(l2_block_hash ++ l2_state_root)`, one per block of the L2 chain.
-While reserved, a zeroed `Bytes32` is used instead.
-This is a work-in-progress, see [issue 181](https://github.com/ethereum-optimism/optimistic-specs/issues/181).
-`MAX_ITEM_COUNT` and/or other parameters will be defined in the withdrawals milestone.
+1. The `latest_block` (`bytes32`) is an execution-layer block of L2, represented as the
+   [`ExecutionPayload`][executionpayload] SSZ type defined in L1. There may be multiple blocks per L2 output root, only
+   the latest is presented.
```
norswap, line 74: "execution-layer block" sounds weird to me, because we don't really make this kind of distinction anywhere (distinction between what even?) and it will lead readers to wonder if there's something to it. Can we just say "L2 block"?

specs/proposals.md
```diff
@@ -47,43 +49,31 @@ output, and the appropriate timestamp, to the [L2 Output Commitment](#l2-output-
...
-defined in L1. There may be multiple blocks per L2 output root, only the latest is presented.
+1. The `state_root` (`bytes32`) is the Merkle-Patricia-Trie ([MPT][g-mpt]) root of all execution-layer accounts,
+   also found in `latest_block.state_root`: this field is frequently used and thus elevated closer to the L2 output
+   root, as opposed to retrieving it from the pre-image of the block in `latest_block`, reducing the merkle proof depth
+   and thus the cost of usage.
 
-[ExecutionPayload]: https://github.com/ethereum/consensus-specs/blob/dev/specs/bellatrix/beacon-chain.md#executionpayload
+1. The `withdrawal_storage_root` (`bytes32`) elevates the Merkle-Patricia-Trie ([MPT][g-mpt]) root of L2 Withdrawal
+   contract storage. Instead of a MPT proof to the Withdrawal contract account in the account trie, one can directly
+access the MPT storage trie root, thus reducing the verification cost of withdrawals on L1.
```
norswap, line 70: Link to withdrawal contract spec.

norswap: suggestion

> Instead of making an MPT proof for a withdrawal against the state root (proving first the storage root of the L2 withdrawal contract against the state root, then the withdrawal against that storage root), we can prove against the L2 withdrawal contract's storage root directly, thus reducing the verification cost of withdrawals on L1.

norswap (approved):
Also agree with Mark on versioning scheme.

tynes:
This needs a rebase and then the CI should work

tynes (approved):

PR #359, opened 2022-04-08
devnet: Hardcode deposit feed bytecode

mslipper:


trianglesphere (approved):
Will need to build a system to make sure this stays up to date, but good for the audit.

PR #360, opened 2022-04-08
opnode: Update and deploy contracts

trianglesphere:
This is rather than hardcoding the contract bytecode. It slows down the e2e
tests (by having to wait 1 L1 block for contracts to be ready), but it
is more accurate to what occurs (and is now required as we initialize
actions).

It also uses the new contracts. No CI for them yet.


opnode/test/setup.go
```diff
@@ -40,14 +43,30 @@ func deriveAccount(w accounts.Wallet, path string) accounts.Account {
...
+	SubmissionFrequency   *big.Int
+	L2BlockTime           *big.Int
+	GenesisL2Output       [32]byte
+	HistoricalTotalBlocks *big.Int
+}
+
+type DepositContractConfig struct {
+	L2Oracle           common.Address
+	FinalizationPeriod *big.Int
+}
```
protolambda, line 56: Would prefer `uint64` in config structs here. And should these have a `Check()` method?

opnode/test/setup.go
```diff
@@ -288,12 +319,57 @@ func (cfg SystemConfig) start() (*System, error) {
...
+	}
+
+	// Deploy contracts
+	sys.L2OOContractAddr, _, _, err = l2oo.DeployL2OutputOracle(
+		opts,
+		l1Client,
+		sys.cfg.L2OOCfg.SubmissionFrequency,
+		sys.cfg.L2OOCfg.L2BlockTime,
+		sys.cfg.L2OOCfg.GenesisL2Output,
+		sys.cfg.L2OOCfg.HistoricalTotalBlocks,
```
mslipper, line 346: This contract takes the starting block timestamp as a parameter as well.

opnode/test/system_test.go
```diff
@@ -50,9 +46,18 @@ func defaultSystemConfig(t *testing.T) SystemConfig {
 			l2OutputHDPath:     10000000,
 			bssHDPath:          10000000,
 		},
+		DepositCFG: DepositContractConfig{
+			FinalizationPeriod: big.NewInt(60 * 60 * 24),
```
mslipper, line 50: Wouldn't you want to set this to a smaller value so that you could test withdrawals?

mslipper:
I'd give this a rebase off latest `main` - there are contract changes that aren't reflected here.

opnode/test/setup.go
trianglesphere: Nice catch. Will add

opnode/test/setup.go
```diff
@@ -288,12 +319,57 @@ func (cfg SystemConfig) start() (*System, error) {
...
+		return nil, err
+	}
+
+	opts, err := bind.NewKeyedTransactorWithChainID(deployerPrivKey, cfg.L1ChainID)
+	if err != nil {
+		return nil, err
+	}
+
+	// Deploy contracts
+	sys.L2OOContractAddr, _, _, err = l2oo.DeployL2OutputOracle(
```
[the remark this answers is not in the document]
trianglesphere: My theory is no. b/c it is sent first, it has a lower nonce than the `OptimismPortal` which means that the portal should be able to reference it even if they go into the same block.

opnode/test/system_test.go
trianglesphere: Probably. I can edit it when I actually add withdrawal tests. I'm not sure how high a priority it is to be able to hand that off to auditors.

opnode/test/setup.go
trianglesphere: These are all internal and they end up as `big.Ints` so I don't see a big reason to hide that fact.

opnode/contracts/Makefile
```diff
@@ -1,56 +1,49 @@
...
+		--combined-json  abi,bin,bin-runtime \
+		--allow-paths $(CONTRACTS_PATH) \
+		-o $(temp)
+	abigen \
+		--combined-json $(temp)/combined.json \
+		--pkg l1block \
+		--type l1block \
+		--out ./l1block/l1_block_info_raw.go
+	cat $(temp)/combined.json | jq -r ".contracts | with_entries(select(.key | match(\"L1Block\")))[] | .\"bin-runtime\"" > bin/l1block_deployed.hex
+
```
tynes, line 35: There is currently nothing enforcing the version of `solc` being used which could result in non-determinism between different machines

opnode/test/setup.go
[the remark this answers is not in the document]
mslipper: Cool! Works for me.

opnode/test/system_test.go
mslipper: Fair point. We can change when we add withdrawal tests.

mslipper (approved):

opnode/contracts/Makefile
tynes: This is what foundry uses to manage versioning: https://github.com/roynalnaruto/svm-rs

Is there anyway that we can get the combined json from `forge build`? That would abstract away the `solc` version management

trianglesphere: Don't we have pragmas on the contracts? I had to switch versions of `solc` this morning to get this to work

trianglesphere: we previously did do something where we would use the `artifacts` from hardhat. My big worry with that is if it doesn't get rebuilt prior to running this so I prefer building from source here.

tynes: The pragmas must be exact matches for this to be safe, it looks like they generally are except for the output oracle: https://github.com/ethereum-optimism/optimistic-specs/blob/33071761277e642a4cf409c744916b3b4189f843/packages/contracts/contracts/L1/L2OutputOracle.sol#L2

opnode/contracts/Makefile
```diff
@@ -1,56 +1,49 @@
...
+
+bindings-optimism-oracle:
```
tynes, line 37: This name makes it sound like its meant to build the output oracle but it builds the portal

opnode/contracts/Makefile
tynes: There is also a `--solc` flag for `abigen` that lets you pass a path to `solc` and it'll compile + build the bindings. Perhaps we allow for a configurable `solc` path with just using `solc` as a default, this way anybody can build and we aren't opinionated about how the user manages the version of `solc`

opnode/contracts/Makefile
trianglesphere: fixed

opnode/contracts/Makefile
trianglesphere: Ok. I'll switch this just `0.8.10` and create an issue to look into this later.

trianglesphere: https://github.com/ethereum-optimism/optimistic-specs/issues/368

tynes: Longer term the solution should be something like `forge bind go` where foundry manages the bindings being up to date

PR #362, opened 2022-04-09
integration-tests: End-to-end withdrawals tests

mslipper:
Includes several fixes to make this work:

- Modifies the devnet to deploy the L1 contracts rather than include them in genesis.
- Changes the deployment of L2OutputOracle to support a custom start timestamp.
- Updates UserDeposits to properly filter out non-TransactionDeposited events.
- Updates the L2 output submitter to post the current L1 block with each output rather than the L2 checkpoint block
- Updates the opnode API to allow all vhosts and CORS domains.


mslipper:
Hmm.. the contract creation is OOG here, but not locally. Will look deeper over the weekend.

opnode/rollup/derive/payload_attributes.go
```diff
@@ -206,7 +206,7 @@ func UserDeposits(receipts []*types.Receipt, depositContractAddr common.Address)
 			continue
 		}
 		for _, log := range rec.Logs {
-			if log.Address == depositContractAddr {
+			if log.Address == depositContractAddr && log.Topics[0] == DepositEventABIHash {
```
mslipper, line 209: @trianglesphere: I added this check since the `OptimismPortal` contract can emit multiple different event types.

l2os/drivers/l2output/driver.go
```diff
@@ -196,7 +196,12 @@ func (d *Driver) CraftTx(
 
 	numElements := new(big.Int).Sub(start, end).Uint64()
 	log.Info(name+" checkpoint constructed", "start", start, "end", end,
-		"nonce", nonce, "blocks_committed", numElements)
+		"nonce", nonce, "blocks_committed", numElements, "checkpoint_block", nextCheckpointBlock)
+
+	header, err := d.cfg.L1Client.HeaderByNumber(ctx, nil)
```
mslipper, line 201: @cfromknecht: I believe that this is the correct logic, but I'm not sure. Would love a gut check.

cfromknecht: @mslipper I suspect we want to pass `nextCheckpointBlock` rather than `nil`, to make sure the header hash corresponds to the L2Output being submitted?

packages/contracts/contracts/L1/L2OutputOracle.sol
```diff
@@ -52,20 +52,22 @@ contract L2OutputOracle is Ownable {
...
+        uint256 _startingBlockTimestamp,
         address sequencer
     ) {
         submissionInterval = _submissionInterval;
         l2BlockTime = _l2BlockTime;
         l2Outputs[block.timestamp] = _genesisL2Output; // solhint-disable not-rely-on-time
         historicalTotalBlocks = _historicalTotalBlocks;
-        latestBlockTimestamp = block.timestamp; // solhint-disable not-rely-on-time
-        startingBlockTimestamp = block.timestamp; // solhint-disable not-rely-on-time
+        latestBlockTimestamp = _startingBlockTimestamp; // solhint-disable not-rely-on-time
```
cfromknecht, line 69: it looks we also need to update above, `l2Outputs[_startingBlockTimestamp] = _genesisL2Output;`

l2os/drivers/l2output/driver.go
mslipper: We can't pass `nextCheckpointBlock` because the sequencer will always be super far ahead of L1. Is there a different block we should be using?

packages/contracts/contracts/L1/L2OutputOracle.sol
mslipper: Fixed, nice catch.

l2os/drivers/l2output/driver.go
cfromknecht: ah i see, missed that this is an L1 block. reviewed the contract again looks good to me 👍

packages/contracts/contracts/L1/L2OutputOracle.sol
maurelian: Agree. In fact it feels like using the L1 timestamp to init the L2 timestamp was probably not a good idea in the first place.

We could include a sanity check to ensure that `_startingBlockTimestamp` is not too distant from `block.timestamp`, but I think we can handle that off-chain.

packages/integration-tests/test/000_withdrawals.spec.ts
```diff
@@ -0,0 +1,212 @@
+// Named 000 in order to run first since the output submitter
```
maurelian, line 1: I see that this file has a `000` prefix but not the others in the same folder?

trianglesphere:
> Hmm.. the contract creation is OOG here, but not locally. Will look deeper over the weekend.

Did you figure this out?

mslipper:
> > Hmm.. the contract creation is OOG here, but not locally. Will look deeper over the weekend.
> 
> Did you figure this out?

I did, yes. It's fixed now.

trianglesphere:
> > > Hmm.. the contract creation is OOG here, but not locally. Will look deeper over the weekend.
> > 
> > 
> > Did you figure this out?
> 
> I did, yes. It's fixed now.

What was the fix?

ops/devnet-up.sh
```diff
@@ -0,0 +1,146 @@
...
+#    contract deployed in step 3.
+# 7. Start the rollup driver.
+# 8. Start the L2 output submitter.
+#
+# The timestamps are critically important here, since the rollup driver will fill in
+# empty blocks if the tip of L1 lags behind the current timestamp. This can lead to
+# a perceived infinite loop. To get around this, we set the timestamp to the current
+# time in this script.
+#
+# This script is save to run multiple times. It stores state in `.devnet`, and
```
trianglesphere, line 24:
```suggestion
# This script is safe to run multiple times. It stores state in `.devnet`, and
```

opnode/rollup/derive/payload_attributes.go
```diff
@@ -77,12 +77,12 @@ func (dep *L1InfoDepositSource) SourceHash() common.Hash {
 //
 // Additionally, the event log-index and
 func UnmarshalLogEvent(ev *types.Log) (*types.DepositTx, error) {
-	if len(ev.Topics) != 3 {
-		return nil, fmt.Errorf("expected 3 event topics (event identity, indexed from, indexed to)")
-	}
 	if ev.Topics[0] != DepositEventABIHash {
 		return nil, fmt.Errorf("invalid deposit event selector: %s, expected %s", ev.Topics[0], DepositEventABIHash)
 	}
+	if len(ev.Topics) != 3 {
```
trianglesphere, line 83: Please revert the order of the checks. The length of topics may be 0 (see LOG0 opcode)

opnode/rollup/derive/payload_attributes.go
trianglesphere: You also need a check for `len(log.Topics) > 0` here. We do check that the ABI is correct in the `UnmarshalLogEvent`, but halt instead of ignoring the error thus adding the check here is important.

trianglesphere (changes requested):
There's one important issue assuming the length of the topics array which would result in a panic.
I didn't look through the tests as much

mslipper:
Fixed @trianglesphere's comments.

trianglesphere (approved):

Makefile
```diff
@@ -1,9 +1,14 @@
 COMPOSEFLAGS=-d
 ITESTS_L2_HOST=http://localhost:9545
 
-build: submodules opnode contracts
+build: submodules opnode contracts integration-tests
 .PHONY: build
 
+build-go: submodules opnode l2os
```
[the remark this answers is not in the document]
mslipper: Fixed!

.github/workflows/devnet-ci.yml
```diff
@@ -24,16 +24,17 @@ jobs:
           cache: 'yarn'
 
       - run: |
-          curl -L -o /tmp/foundry.tgz https://github.com/gakonst/foundry/releases/download/nightly-9c2469488c6872e5d17198555f7d8e1a80173151/foundry_nightly_linux_amd64.tar.gz
+          curl -L -o /tmp/foundry.tgz https://github.com/gakonst/foundry/releases/download/nightly/foundry_nightly_linux_amd64.tar.gz
           tar -xzvf /tmp/foundry.tgz
```
tynes, line 28: There is a github action for this https://github.com/onbjerg/foundry-toolchain

mslipper: Georgios just opened a PR for that, I'll rebase this once that PR is merged.

packages/contracts/deploy/000-L2OutputOracle.deploy.ts
```diff
@@ -0,0 +1,26 @@
...
+  const { deployer } = await hre.getNamedAccounts()
+
+  await deploy('L2OutputOracle', {
+    from: deployer,
+    args: [
+      15,
+      2,
+      `0x${'00'.repeat(32)}`,
+      0,
+      process.env.L2OO_STARTING_BLOCK_TIMESTAMP,
```
tynes, line 16: This could be hardened by adding a fallback value if its not set, maybe checking to see the timestamp on the latest block?

packages/integration-tests/test/000_withdrawals.spec.ts
```diff
@@ -0,0 +1,212 @@
...
+    before(async function () {
+      this.timeout(60_000)
+      recipient = Wallet.createRandom().connect(env.l2Provider)
+      withdrawer = withdrawer.connect(recipient)
+
+      logger.info('Generated new wallet', {
+        recipient: recipient.address,
+      })
+      logger.info('Depositing to new address on L2')
+      let tx = await portal.connect(env.l1Wallet).depositTransaction(
```
tynes, line 54: You should be able to send value directly to the contract without needing to call `depositTransaction`, could remove some lines of code

mslipper: I kind of like being explicit about calling the contract rather than just sending value - it makes it clear that we're calling the portal, and what parameters we're setting.

packages/contracts/deploy/000-L2OutputOracle.deploy.ts
mslipper: IMO the deployment should fail if that value isn't set rather than falling back. I can update that.

packages/integration-tests/test/000_withdrawals.spec.ts
```diff
@@ -0,0 +1,212 @@
...
+        [],
+        {
+          value: utils.parseEther('1.337'),
+        },
+      )
+      await tx.wait()
+
+      await awaitCondition(async () => {
+        const bal = await recipient.getBalance()
+        return bal.eq(tx.value)
```
tynes, line 68: This isn't stateless, ie you would need to bring down the system and bring it up again for this condition to be met. It would be better to do something like what the hardhat task does, where it does `DepositTx.fromL1Event().hash()` and then polls L2 for that tx hash

mslipper: What do you mean? This condition does get met, otherwise the tests wouldn't pass. The recipient is a random address with zero balance, so we can assert that the balance equals the transaction's value.

I can update it to use the same thing the hardhat task does, assuming those libraries are accessible to the integration tests package. Otherwise, I'd like to do that cleanup once we've moved into the monorepo. I'd rather not invest time into making code reuse work if we're just going to rip all that out in a week or two.

tynes: Ah got it - the random address part that the value is sent to prevents the problem that I was thinking of, my bad. I was thinking running the tests twice against the same state but since its a random address its fine

tynes (approved):

packages/integration-tests/test/000_withdrawals.spec.ts
mslipper: Ok sounds good!

PR #366, opened 2022-04-11
Add reverted deposit with mint test

Inphi:
This tests the edge-case that motivated https://github.com/ethereum-optimism/reference-optimistic-geth/pull/10

When a deposit is invalid, it must still be included in the L2, without state-changes, except the mint and nonce changes.


opnode/test/system_test.go
```diff
@@ -343,7 +343,66 @@ func TestSystemE2E(t *testing.T) {
...
+		URL: accounts.URL{
+			Path: transactorHDPath,
+		},
+	})
+	require.Nil(t, err)
+	opts, err := bind.NewKeyedTransactorWithChainID(ethPrivKey, cfg.L1ChainID)
+	fromAddr := crypto.PubkeyToAddress(ethPrivKey.PublicKey)
+	require.Nil(t, err)
+
+	toAddr := common.Address{0xff, 0xff}
```
trianglesphere, line 375: One nice check would to be sure that the `toAddr` balance is still 0 after the deposit (should mint if it has value, but should fail to transfer)

opnode/test/system_test.go
```diff
@@ -343,7 +343,66 @@ func TestSystemE2E(t *testing.T) {
...
+	require.Nil(t, err, "Waiting for deposit tx on L1")
+
+	reconstructedDep, err := derive.UnmarshalLogEvent(receipt.Logs[0])
+	require.NoError(t, err, "Could not reconstruct L2 Deposit")
+	tx = types.NewTx(reconstructedDep)
+	receipt, err = waitForTransaction(tx.Hash(), l2Verif, 6*time.Second)
+	require.NoError(t, err)
+	require.Equal(t, receipt.Status, types.ReceiptStatusSuccessful)
+
+	// Confirm balance
```
trianglesphere, line 397: Another check (which I think would fail right now) is to make sure that the nonce of the depositing account is bumped.

protolambda: No execution happened though, only one balance was increased (which you could think of as the system moving the funds), should we even increment the nonce of the account?

trianglesphere: I think so. Reverted transactions also bump the nonce. It's weird that some deposit txns increase the nonce but others do not.

Inphi: Indeed. This test should also check the nonce.

protolambda: Added that to the test, and implemented the nonce increment here: https://github.com/ethereum-optimism/reference-optimistic-geth/pull/10/commits/7058816e614c29c7f509f8e1ec1623ee154a22c5

go.sum
```diff
@@ -70,6 +70,7 @@ github.com/btcsuite/btcd v0.22.0-beta h1:LTDpDKUM5EeOFBPM8IXpinEcmZ6FWfNZbE3lfrf
 github.com/btcsuite/btcd v0.22.0-beta/go.mod h1:9n5ntfhhHQBIhUvlhDvD3Qg6fRUj4jkN0VB8L8svzOA=
 github.com/btcsuite/btcd/btcec/v2 v2.1.2 h1:YoYoC9J0jwfukodSBMzZYUVQ8PTiYg4BnOWiJVzTmLs=
 github.com/btcsuite/btcd/btcec/v2 v2.1.2/go.mod h1:ctjw4H1kknNJmRN4iP1R7bTQ+v3GJkZBd6mui8ZsAZE=
+github.com/btcsuite/btcd/chaincfg/chainhash v1.0.0 h1:MSskdM4/xJYcFzy0altH/C/xHopifpWzHUi1JeVI34Q=
```
Inphi, line 73: go mod tidy reported the following error:
```
github.com/ethereum-optimism/optimistic-specs/l2os imports
        github.com/miguelmota/go-ethereum-hdwallet imports
        github.com/btcsuite/btcd/chaincfg imports
        github.com/btcsuite/btcd/chaincfg/chainhash loaded from github.com/btcsuite/btcd@v0.22.0-beta,
        but go 1.16 would fail to locate it:
        ambiguous import: found package github.com/btcsuite/btcd/chaincfg/chainhash in multiple modules:
        github.com/btcsuite/btcd v0.22.0-beta (/home/inphi/go/pkg/mod/github.com/btcsuite/btcd@v0.22.0-beta/chaincfg/chainhash)
        github.com/btcsuite/btcd/chaincfg/chainhash v1.0.0 (/home/inphi/go/pkg/mod/github.com/btcsuite/btcd/chaincfg/chainhash@v1.0.0)

To proceed despite packages unresolved in go 1.16:
        go mod tidy -e
If reproducibility with go 1.16 is not needed:
        go mod tidy -compat=1.17
For other options, see:
        https://golang.org/doc/modules/pruning
```
I reran tidy using `go mod tidy -e` to resolve.

protolambda:
Changed `go.mod` to use the squashed commit with the changes instead. And also did that `go mod tidy -e`. Our Go minimum version in opnode is Go 1.17, so the go 1.16 problem doesn't matter and `-e` is fine.

protolambda (approved):
LGTM

PR #369, opened 2022-04-11
ci: Add automated docker builds for opnode and l2 output submitter

mslipper:


ops/l2os-entrypoint.sh
```diff
@@ -1,14 +0,0 @@
-#!/bin/sh
```
protolambda, line 1: I think it's nice to reduce code, and don't like the curl at startup, but why exactly does this PR remove these sh files?

ops/Dockerfile.l2os
```diff
@@ -7,17 +7,13 @@ COPY go.mod /app/go.mod
...
 
 FROM alpine:3.15
 
-RUN apk add --no-cache curl jq
-
 COPY --from=builder /app/bin/l2os /usr/local/bin
-COPY --from=builder /app/l2os-entrypoint.sh /l2os-entrypoint.sh
 
-ENTRYPOINT ["/bin/sh", "/l2os-entrypoint.sh"]
+CMD ["l2os"]
```
protolambda, line 19: nice clean up

.circleci/config.yml
```diff
@@ -0,0 +1,57 @@
...
+      - checkout
+      - setup_remote_docker:
+          version: 19.03.13
+      - run:
+          name: Build
+          command: |
+            echo "$DOCKER_PASS" | docker login -u "$DOCKER_USERNAME" --password-stdin
+            docker build -t "ethereumoptimism/<<parameters.image-name>>:<<parameters.image-tag>>" -f <<parameters.dockerfile>> <<#parameters.target>>--target <<parameters.target>><</parameters.target>> .
+            docker push "ethereumoptimism/<<parameters.image-name>>:<<parameters.image-tag>>"
+
```
protolambda, line 32: Not sure how to test this part without access, seems to be the complicated part of the PR.

.circleci/config.yml
```diff
@@ -0,0 +1,57 @@
+version: 2.1
+
+
+jobs:
+  build-dockerfile:
+    docker:
+      - image: cimg/base:2021.04
```
protolambda, line 7: should we try a newer image?

trianglesphere (commented):
A couple small Qs. Generally looks fine.

.circleci/config.yml
```diff
@@ -0,0 +1,57 @@
...
+          context:
+            - optimism
+          filters:
+            branches:
+              only:
+                - main
+          image-name: l2os
+          image-tag: develop
+          dockerfile: ops/Dockerfile.l2os
+          name: Build L2 output submitter
```
maurelian, line 57: Friendly nudge to make use of the editorConfig file we have in this repo and the monorepo. I think you're on [IntelliJ](https://www.jetbrains.com/help/idea/editorconfig.html)?

ops/l2os-entrypoint.sh
mslipper: The best practice for Docker containers is to treat them like executables. They should usually expose a single `CMD` that people can then pass env vars or CLI flags.

We violated this best practice in order to make sure that dependent services were online before starting the main executable in the container. The devnet-up script does this for us now, so we can remove these entrypoint scripts.

.circleci/config.yml
mslipper: Yeah I can upgrade this.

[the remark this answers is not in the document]
mslipper: 👍 will upgrade.

.circleci/config.yml
```diff
@@ -0,0 +1,57 @@
...
+        description: Dockerfile to use
+        type: string
+    steps:
+      - checkout
+      - setup_remote_docker:
+          version: 19.03.13
+      - run:
+          name: Build
+          command: |
+            echo "$DOCKER_PASS" | docker login -u "$DOCKER_USERNAME" --password-stdin
```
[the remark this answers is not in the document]
mslipper: The CircleCI side injects these environment variables at runtime.

They can be leaked if we run untrusted code within CircleCI, which is why we have to whitelist untrusted PRs before running them.

.circleci/config.yml
mslipper: Once this is merged I'll run the job to test this out. We've used this exact code elsewhere to push images, the only issue will be setting up the Docker context which happens in the CircleCI Console.

maurelian (approved):

PR #370, opened 2022-04-13
contracts: Output oracle improvements

maurelian:
1. Makes immutable casing consistent with other contracts in this repo.
2. Adds a check in the constructor to ensure that "Submission Interval is a multiple of L2 Block Time".

Please ignore the branch name, it does not represent the changes herein. I made this changes in the course of other work but forgot to move them to a new branch.


packages/contracts/contracts/L1/L2OutputOracle.sol
```diff
@@ -62,12 +62,17 @@ contract L2OutputOracle is Ownable {
         uint256 _startingBlockTimestamp,
         address sequencer
     ) {
-        submissionInterval = _submissionInterval;
-        l2BlockTime = _l2BlockTime;
+        require(
+            _submissionInterval % _l2BlockTime == 0,
+            "Submission Interval must be a multiple of L2 Block Time"
+        );
```
maurelian, line 68: This prevents a situation where some outputs would be useless zero values, ie. given:

```
_submissionInterval = 2
_l2BlockTime = 5
```
Then the outputs with `timestamp` of 5, 15, 25, etc would not have corresponding l2 block.

maurelian: I chose to use a string here instead of a custom error, unlike the other errors in this contract, because this one is in the constructor, so I didn't want to add clutter with another `error` declaration at the top of this contract.

mslipper (approved):

PR #371, opened 2022-04-13
style: add comments and cleanup to state.go

smartcontracts:
**Description**
Second comment+cleanup PR for state.go. Adds lots of comments and shifts
a few things around to make the code easier to read. Also fixes one bug
found in state.go.

### DIFF
--- a/opnode/rollup/driver/state.go
+++ b/opnode/rollup/driver/state.go
@@ -30,11 +30,11 @@ type state struct {
 	sequencer bool
 
 	// Connections (in/out)
-	l1Heads <-chan eth.L1BlockRef
-	l1      L1Chain
-	l2      L2Chain
-	output  outputInterface
-	bss     BatchSubmitter
+	l1HeadsCh <-chan eth.L1BlockRef
+	l1        L1Chain
+	l2        L2Chain
+	output    outputInterface
+	bss       BatchSubmitter
 
 	log  log.Logger
 	done chan struct{}
@@ -57,7 +57,7 @@ func NewState(log log.Logger, config rollup.Config, l1 L1Chain, l2 L2Chain, outp
 
 // Start starts up the state loop. The context is only for initilization.
 // The loop will have been started iff err is not nil.
-func (s *state) Start(ctx context.Context, l1Heads <-chan eth.L1BlockRef) error {
+func (s *state) Start(ctx context.Context, l1HeadsCh <-chan eth.L1BlockRef) error {
 	l1Head, err := s.l1.L1HeadBlockRef(ctx)
 	if err != nil {
 		return err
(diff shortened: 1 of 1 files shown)


opnode/rollup/driver/state.go
```diff
@@ -72,13 +72,16 @@ func (s *state) Start(ctx context.Context, l1Heads <-chan eth.L1BlockRef) error
 		// Ensure that we are on the correct chain. Note that we cannot rely on rely on the UnsafeHead being more than
 		// a sequence window behind the L1 Head and must walk back 1 sequence window as we do not track the end L1 block
 		// hash of the sequence window when we derive an L2 block.
+		// TODO: safeHead returned here is actually the latest L2 block that *could* be the safe
+		// head, but we don't actually know if it necessarily *is* the safe head because it
+		// might've come into this node via unsafe distribution (directly from the sequencer). We
+		// need to figure out how to guarantee that this head is actually safe.
```
trianglesphere, line 78: My preferred solution for this is to add metadata to blocks in the L2 execution engine that includes the block hash of the L1 block that was at the end of the sequencing window.

opnode/rollup/driver/state.go
```diff
@@ -258,41 +274,51 @@ func (s *state) createNewL2Block(ctx context.Context) error {
 // It ensures that a full sequencing window is available and updates the state as needed.
 func (s *state) handleEpoch(ctx context.Context) (bool, error) {
 	s.log.Trace("Handling epoch", "l2Head", s.l2Head, "l2SafeHead", s.l2SafeHead)
-	// Extend cached window if we do not have enough saved blocks
+
+	// Extend cached window if we do not have enough saved blocks.
 	if len(s.l1WindowBuf) < int(s.Config.SeqWindowSize) {
-		// attempt to buffer up to 2x the size of a sequence window of L1 blocks, to speed up later handleEpoch calls
+		// Attempt to buffer up to 2x the sequencing window. Anything up to the window size is good
+		// enough, but caching a few extra blocks will speed things up down the line.
```
trianglesphere, line 281:
```suggestion
		// enough, but caching more reduces the number of RPC calls.
```

opnode/rollup/driver/state.go
smartcontracts: How would that tweak be used to fix the problem?

trianglesphere: You can then check if the end of the sequencing window is still canonical in L1. The question is if the view of the chain is the same as when the L2 block was derived.

protolambda: We shouldn't have to do that, optimizing/improving the stateless solution is a lot better than adding extra storage details in the geth db and increasing the diff IMO.

norswap: Wait, I thought in our call (me, Proto & Josh) we figured that it's not too hard to get the safe head? Just walk back and get the last L2 block whose L1 origin is a full sequencing window behind the L1 head. When I reviewed, it was my impression that this was indeed what `FindL2Heads` did?

Also, I thought the above was literally the definition of the safe head. If you're getting it from unsafe block distribution BUT it can't be derived from L1, it's not the safe head, innit?

norswap:
@smartcontracts I'm interested in reviewing this when it's ready for review!

opnode/rollup/driver/state.go
trianglesphere: That is what `FindL2Heads` does. I realized though that the `SafeHead` it returns is in fact the `MaxPermissableSafeHead` - say you have a safe head behind it, you don't want to advance to that safe head.

I think as I realized this, I've started to switch my definition of `safeHead` from "could be fully derived from L1" to "is fully derived from L1". As such, it is likely, but not guaranteed, that the `safeHead` is a sequencing window behind the `head`

So in this case, we can query the L2 Execution engine for the `safeHead` and still prevent it from being advanced.

trianglesphere: > We shouldn't have to do that, optimizing/improving the stateless solution is a lot better than adding extra storage details in the geth db and increasing the diff IMO.

This problem is inherently stateful. The contents of any L2 block depends upon view of the L1 chain that the rollup node had when the L2 block was derived.
Right now, we have this implicit commitment to the view of the L1 chain in the `L1 Attributes` stored inside the L2 chain.

trianglesphere:
@smartcontracts closing b/c this never got pulled out of draft status. LMK if you want to reopen it in the monorepo

PR #372, opened 2022-04-14
Clean up withdrawals itest

maurelian:
Please review #370 first, which this extends. 

Changes added here: 
1. Makes the logging less noisy in the withdrawals spec file, and consolidates the two awaitCondition loops into a single one.
2. Uses math to predict the target output timestamp, instead of polling to see when a timestamp has been appended that is larger than the burn block.


mslipper (approved):

packages/integration-tests/test/000_withdrawals.spec.ts
```diff
@@ -15,8 +15,36 @@ import winston from 'winston'
...
+      (withdrawalTimestamp - startingBlockTimestamp)
+      / submissionInterval
+    )
+    * submissionInterval
+    + startingBlockTimestamp
+  }
+
+  return targetOutputTimestamp
+}
+
```
norswap, line 47: This seems like a useful thing that should not only live in the tests?

Issue #163, opened 2022-01-26
Milestone epic: Sequencer (unsafe) block distribution on L2 via P2P


protolambda:
TODO list for happy-case block distribution through P2P:

- [x] Initial P2P spec, configuration/flags, construction code and testing: #373
- [x] See #388
  - attach the host, and discv5 to the rollup-node struct. And `Close` them on shutdown.
- [x] See #394
  - define the block and signature serialization for P2P distribution
  - create a gossipsub instance wrapping the host, registering and subscribing to the L2 blocks topic.
  - implement the publishing of blocks on gossip
  - implement the validation of blocks on gossip
  - test the L2 blocks making a trip from sequencer to verifier, without L1
- [x] implement the peerstore feeding with a background discv5 process that randomly walks the DHT for new nodes when necessary
- ~~implement basic peer-scoring~~ (will be addressed in a later iteration)

protolambda:
Discovery landed, which wraps up this milestone. And we've been tracking the same things + new things in Linear. Closing this.

PR #373, opened 2022-04-14
P2P block distribution

protolambda:
This PR:
- [x] defines new CLI flags for many of the P2P options
- [x] defines a config for all P2P options
- [x] implements the setup of a LibP2P host and Discv5 node.
- [x] tests p2p host setup
- [x] specs the general usage of the libp2p and discv5 components

The remaining P2P work is being tracked in #163

The P2P package is not yet integrated into the opnode at this point, and the specs live in a separate document, so there are no breaking changes in this PR.


opnode/flags/p2p_flags.go
```diff
@@ -0,0 +1,223 @@
...
+		Required: false,
+		EnvVar:   p2pEnv("NO_DISCOVERY"),
+	}
+	P2PPrivPath = cli.StringFlag{
+		Name: "p2p.priv.path",
+		Usage: "Read the hex-encoded 32-byte private key for the peer ID from this txt file. Created if not already exists." +
+			"Important to persist to keep the same network identity after restarting, maintaining the previous advertised identity.",
+		Required: false,
+		Value:    "opnode_p2p_priv.txt",
+		EnvVar:   p2pEnv("PRIV_PATH"),
```
trianglesphere, line 32: Should add `TakesFile` here? I can't actually find much documentation on what the flag does.

opnode/p2p/config.go
```diff
@@ -0,0 +1,394 @@
...
+			return fmt.Errorf("failed to make %s constructor: %v", v, err)
+		}
+		conf.HostSecurity = append(conf.HostSecurity, sc)
+	}
+
+	conf.PeersLo = ctx.GlobalUint(flags.PeersLo.Name)
+	conf.PeersHi = ctx.GlobalUint(flags.PeersHi.Name)
+	conf.PeersGrace = ctx.GlobalDuration(flags.PeersGrace.Name)
+
+	// TODO parse NAT option
```
trianglesphere, line 294: Important for now or not?

opnode/p2p/host_test.go
```diff
@@ -0,0 +1,136 @@
...
+
+// Simplified p2p test, to debug/test basic libp2p things with
+func TestP2PSimple(t *testing.T) {
+	confA := TestingConfig(t)
+	confB := TestingConfig(t)
+	hostA, err := confA.Host()
+	assert.NoError(t, err, "failed to launch host A")
+	defer hostA.Close()
+	hostB, err := confB.Host()
+	assert.NoError(t, err, "failed to launch host B")
```
trianglesphere, line 64: assert vs require. I guess we ended up on require elsewhere. I don't have a strong opinion either way, but consistency is nice.

trianglesphere (commented):
There's one lint issue with the new specs.
I'm pretty unfamiliar with networking code, but what's here looks good. The PR desc should probably be pulled out to a ticket and then this could probably be merged.

specs/rollup-node-p2p.md
```diff
@@ -0,0 +1,258 @@
...
+To ensure malicious peers get scored based on application behavior the validation signals
+`ACCEPT` (valid), `IGNORE` (like inactivity) or `REJECT` (score penalties).
+
+In order of operation:
+
+- `[REJECT]` if the encoding or compression is not valid
+- `[REJECT]` if the block timestamp is older than 20 seconds in the past
+  (graceful boundary for worst-case propagation and clock skew)
+- `[REJECT]` if the block timestamp is more than 5 seconds into the future (graceful boundary for clock skew)
+- `[REJECT]` if the signature by the sequencer is not valid
```
tynes, line 223: How do we manage the set of sequencer keys? Does this live in the L1 state?

specs/rollup-node-p2p.md
```diff
@@ -0,0 +1,258 @@
...
+- `[REJECT]` if more than 5 blocks have been seen with the same block height
```
tynes, line 224: Does ETH2 reject blocks if more than 5 have been seen with the same height? It seems like a way to perma-partition the network by selectively gossiping blocks

specs/rollup-node-p2p.md
```diff
@@ -0,0 +1,258 @@
...
+
+#### Block validation
+
+To ensure malicious peers get scored based on application behavior the validation signals
+`ACCEPT` (valid), `IGNORE` (like inactivity) or `REJECT` (score penalties).
+
+In order of operation:
+
+- `[REJECT]` if the encoding or compression is not valid
+- `[REJECT]` if the block timestamp is older than 20 seconds in the past
```
tynes, line 220: If the block timestamp is older than 20 seconds in the past, how would it be possible to sync historically? This can only apply if we are at the tip, but how does one know if they are at the tip?

specs/rollup-node-p2p.md
```diff
@@ -0,0 +1,258 @@
...
+  (graceful boundary for worst-case propagation and clock skew)
+- `[REJECT]` if the block timestamp is more than 5 seconds into the future (graceful boundary for clock skew)
```
tynes, line 222: In L1, the  more than 15 seconds in the future is a policy rule and not a consensus rule, I think it may be useful to delineate between consensus and policy here

specs/rollup-node-p2p.md
```diff
@@ -0,0 +1,258 @@
...
+- `[REJECT]` if the signature by the sequencer is not valid
+- `[REJECT]` if more than 5 blocks have been seen with the same block height
+- `[IGNORE]` if a conflicting block was already seen on L1.
+  It may not be malicious due to racing between L1 confirmation and L2 propagation, but should be filtered out.
+
+The block is signed by the corresponding sequencer, to filter malicious messages.
+The sequencer model is singular but may change to multiple sequencers in the future.
+A default sequencer pubkey is distributed with rollup nodes and should be configurable.
```
tynes, line 230: If we decide to run clique consensus, we likely want multiple nodes using different keys for redundancy. If the key is hardcoded, then we will have a really bad time if the key is leaked or lost

specs/rollup-node-p2p.md
```diff
@@ -0,0 +1,258 @@
...
+
+## Gossip Topics
+
+### `blocks`
+
+The primary topic of the L2, to distribute blocks to other nodes faster than proxying through L1 would.
+
+#### Block encoding
+
+TODO: encode execution payload (SSZ or RLP), with sequencer identifier and signature.
```
tynes, line 208: Should the p2p format be any different than the L1 block format RLP encoded with the signature in the extradata?

specs/rollup-node-p2p.md
```diff
@@ -0,0 +1,258 @@
...
+The [rollup node](./rollup-node.md) has an optional peer-to-peer (P2P) network service to improve the latency between
+the view of sequencers and the rest of the network by bypassing the L1 in the happy case,
+without relying on a single centralized endpoint.
+
+This also enables faster historical sync to be bootstrapped by providing block headers to sync towards,
+and only having to compare the L2 chain inputs to the L1 data as compared to processing everything one block at a time.
+
+The rollup node will *always* prioritize L1 and reorganize to match the canonical chain.
+The L2 data retrieved via the P2P interface is strictly a speculative extension to improve the happy case performance.
+
```
tynes, line 12: Perhaps using the `unsafe` terminology would be useful here

opnode/flags/p2p_flags.go
protolambda: It's just a hint for auto-completing file arguments in the shell (they say something about fish support, but I don't use it). Will add it.

opnode/p2p/host_test.go
protolambda: I've only been using `github.com/stretchr/testify/assert` in all of my spec work so far. We also use their mock library. I think the system test uses `require` 99% to check if errors are nil with a generic `require.Nil`. And then there are like ~5 misc calls to `require`. If think we should change those others to `assert.NoError(t, err)`

specs/rollup-node-p2p.md
protolambda: Undecided yet, we could put a pubkey (or even just an address, since the pubkey can be recovered from the sig) in the L1 state and read it with an RPC call yes. Or we just make it part of the chain configuration while we don't have more than 1 sequencer yet. Maybe needs a key-replace mechanism though.

opnode/p2p/config.go
protolambda: This is an old TODO, will clean it up. Idea was to offer more advanced NAT configuration, but we already have a flag for setting the advertised external IP explicitly, and I don't think anyone cares to select PMP, UPNP or both, as long as it can be disabled/enabled.

specs/rollup-node-p2p.md
protolambda: Historic block-by-block sync wouldn't happen through gossipsub. Instead the peer should just read the data from L1. And if we want p2p historic block-by-block sync (if it's even faster than pulling the data from L1) then we can specify a req-resp protocol in libp2p (libp2p enables you to open named read/write channels with peers, if the peers negotiate the support for that channel protocol during their connection).

For long-range happy-case sync though, we only need *any recent* header. We verify that it matches the output root on L1 if we want more of a guarantee we're syncing a correct chain (otherwise have to reorg it all out again if it's wrong), and then call the engine with a new-payload call that doesn't simply extend the chain by 1 block, but is that recent header. The engine will then start snap-sync (or other type of sync) towards that header, on its own. This is a later implementation step to trigger, we just need recent blocks for now.

specs/rollup-node-p2p.md
protolambda: These are propagation rules, not consensus rules. At the top of the doc it already explains that this whole p2p stack does not affect consensus.

specs/rollup-node-p2p.md
protolambda: Yep, need some key-replacement mechanism. We can discuss what's the best approach later, for now I'm going to keep it simple.

specs/rollup-node-p2p.md
protolambda: Yes, we don't intend to use `extradata`, at all. These signatures need to only be available temporarily, they are only required against DoS/spam protection during gossip, not during sync. And yes, we can have a different key for gossip than for batch-submission.

So instead of `extradata` like Clique, we just encode the block (in whatever format, RLP is fine), and wrap it with an envelope container type that adds a signature.

Also note that we don't use the total-difficulty like clique to figure out the head of the chain. We just take the first block received from the sequencer (while there's only one sequencer this is simple, at worst there's equivocation with multiple different blocks, but we'll check against L1 anyway) and try to apply it to the engine as unsafe-block.

specs/rollup-node-p2p.md
protolambda: Eth2 can always sync blocks by slot number or by root via req-resp, gossip is only used as source to get recent blocks. 

And their gossip is stricter: only the first block of a certain `(validator_index, slot)` combination is accepted (and the `validator_index` must be a valid proposer w.r.t. the parent block).

I'm thinking we could do the same, but I'm undecided if we want to propagate reorgs as a *single* sequencer, or fall back completely to syncing blocks from L1 if we are reorging on L2. In Eth2 a single validator would never reorg itself (double proposals are slashable), but a rollup sequencer might be forced to if L1 reorgs, so the network may need some more graceful rules.

opnode/p2p/host_test.go
protolambda: Scratch that, didn't realize `require` is from `testify` too. It just terminates the test quicker, which makes sense for error handling. I'll update my tests

specs/rollup-node-p2p.md
tynes: > Instead the peer should just read the data from L1

If blob transactions are a thing, then this will not be possible. It also sounds like you are really in favor of an eip4444 like mechanism. I agree that we need to be able to snap sync via a single recent header that is greater than 1 week old (finality, knowing we are on the correct chain). I do think that it is important to be able to spin up archive nodes though which requires historical execution because no existing sync mechanism is designed specifically for archive nodes

opnode/p2p/host_test.go
```diff
@@ -0,0 +1,136 @@
...
+	noiseC, err := noiseC()
+	require.NoError(t, err)
+	tlsC, err := tlsC()
+	require.NoError(t, err)
+
+	confA := Config{
+		Priv:                (*ecdsa.PrivateKey)((pA).(*crypto.Secp256k1PrivateKey)),
+		DisableP2P:          false,
+		NoDiscovery:         true,
+		ListenIP:            net.IP{127, 0, 0, 1},
```
mslipper, line 91: Don't we want to parameterize this?

opnode/p2p/host_test.go
```diff
@@ -0,0 +1,136 @@
...
+		ListenTCPPort:       0, // bind to any available port
+		StaticPeers:         nil,
+		HostMux:             []lconf.MsMuxC{yamuxC, mplexC},
+		HostSecurity:        []lconf.MsSecC{noiseC, tlsC},
```
mslipper, line 95: For both `HostSecurity` and `HostMux`, what do you think of limiting each of these to only one option? Concerned about offering too many different configurations that have to be negotiated at connection time. We have the ability to require specific settings since the network isn't live yet.

opnode/p2p/host_test.go
protolambda: This is just a test, binding to localhost should be fine to check if it works. Generally we'd use mocknets in p2p tests to avoid binding to an actual network interface, but we do need to test this.

opnode/p2p/host_test.go
protolambda: I prefer to have at least the two most popular options. When one of them breaks or underperforms we can prioritize the other in an upgrade, and keep the old and new nodes connected.

specs/rollup-node-p2p.md
```diff
@@ -0,0 +1,292 @@
...
+Peers may be banned if their performance score is too low, or if an objectively malicious action was detected.
+
+Banned peers will be persisted to the same data-store as the peerstore records.
+
+TODO: the connection gater does currently not gate by IP address on the dial Accept-callback.
+
+#### Transport security
+
+[Libp2p-noise][libp2p-noise], `XX` handshake, with the the `secp256k1` P2P identity, as popularized in Eth2.
```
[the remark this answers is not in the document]
protolambda: It refers to the `XX` handshake pattern: https://github.com/libp2p/specs/blob/master/noise/README.md#xx

specs/rollup-node-p2p.md
```diff
@@ -0,0 +1,292 @@
...
+
+#### Block validation
+
+To ensure malicious peers get scored based on application behavior the validation signals
+`ACCEPT` (valid), `IGNORE` (like inactivity) or `REJECT` (score penalties).
+
+In order of operation:
+
+- `[REJECT]` if the encoding or compression is not valid
+- `[REJECT]` if the block timestamp is older than 20 seconds in the past
```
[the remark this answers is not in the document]
protolambda: yea, happy to adjust this type of number later on after testing.

PR #374, opened 2022-04-14
specs: Fix definition of output root

maurelian:
Changes: 

- `latest_block` -> `latest_block_hash`
- Rewrote to words about hoisting the state_root, which were necessary because of the above change.


norswap (approved):

protolambda (approved):
LGTM

PR #375, opened 2022-04-15
solidity: fail if updated gas snapshot isn't included

tynes:
A new gas snapshot can be generated by running
`forge snapshot`

**Description**
Add updated gas snapshot and CI that enforces gas snapshots to be added if there is a change


maurelian (approved):
LGTM

PR #376, opened 2022-04-16
small fixes

norswap:
Fix a broken link, a factual mistake and a weird terminology term.

### DIFF
--- a/specs/rollup-node.md
+++ b/specs/rollup-node.md
@@ -391,14 +391,16 @@ meaning the head of the L1 chain changes to a block that is not the child of the
 descendant of an ancestor of the previous head. In that case, the rollup driver must first search for the common L1
 ancestor, and can re-derive the L2 chain from that L1 block and onward.
 
-The rollup node maintains two heads of the L2 Chain: the unsafe head (often called head) and the safe head.
-Each L2 block has an L1 Attributes block that it references in the L1 Info deposit transaction.
-The unsafe head is the head of the L2 chain. It's L1 Attributes should be canonical or potentially extending the
-canonical chain (if the rollup node has not yet seen the L1 block that it is based upon).
-The safe head is the the last L2 block of the last epoch whose sequencing window is complete
-(i.e. the epoch with number `L1Head.number` - `SEQUENCING_WINDOW_SIZE`).
+The rollup node maintains two heads of the L2 Chain: the unsafe head (often called head) and the safe head. Each L2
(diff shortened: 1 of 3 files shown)


specs/rollup-node.md
```diff
@@ -391,14 +391,16 @@ meaning the head of the L1 chain changes to a block that is not the child of the
...
-The safe head is the the last L2 block of the last epoch whose sequencing window is complete
-(i.e. the epoch with number `L1Head.number` - `SEQUENCING_WINDOW_SIZE`).
-
-Steps during a reorg
+The rollup node maintains two heads of the L2 Chain: the unsafe head (often called head) and the safe head. Each L2
+block has an L1 origin block (corresponding to its epoch) that it references in the [L1 attributes deposited
+transaction][l1-attr-deposit]. The unsafe head is the head of the L2 chain. Its L1 origin block should be canonical or
+potentially extending the canonical chain (if the rollup node has not yet seen the L1 block that it is based upon). The
+safe head is the the last L2 block of the last epoch whose sequencing window is complete (i.e. the epoch with number
+`L1Head.number` - `SEQUENCING_WINDOW_SIZE`).
```
protolambda, line 399: Changes look ok, but I recommend changing the formatting to split text by sentence, not by exact line-length. It's annoying to see a whole text section re-wrap just due to a single sentence getting a little bit different length. The markdown renderer will only show text on a new line when we use a blank line anyway. This will reduce the diff in the future

protolambda (approved):
LGTM, would be nice to split lines by sentence for future diff usability though.

specs/rollup-node.md
norswap: Yeah kinda weird that github doesn't do the word highlighting in this case. I wonder if it's because the diff is > one paragraph? I might have been able to edit the hunks to avoid that.

I kinda like exact line length. You can reformat in one-keystroke in most editor, whereas the other one is a lot of fuss.
If you get word highlighting (and it's formatted to begin with), that should help a lot.

protolambda: It's not just about review-ability, but also about reducing the affected lines in git itself. Merge conflicts / rebases get a lot easier when the changes only affect lines they closely relate to.

norswap: I don't we often edit the same spec simultaneously. But I hear you, I'll change it back to single lines, doesn't matter all that much to me.

PR #377, opened 2022-04-18
contracts: Add deleteL2Output function

maurelian:
**Description**

Adds a `deleteL2Output` function to the OutputOracle, so that the sequencer can retract an erroneously appended output.

A similar function was included in the original prototype: 
https://gist.github.com/karlfloersch/02ce7938000680ce7820f39ba06bd9b1#file-staterootoracle-sol-L53-L60


packages/contracts/contracts/L1/L2OutputOracle.sol
```diff
@@ -117,6 +120,20 @@ contract L2OutputOracle is Ownable {
...
+     *  the wrong root
+     */
+    function deleteL2Output(bytes32 _l2Output) external onlyOwner {
+        bytes32 outputToDelete = l2Outputs[latestBlockTimestamp];
+        require(_l2Output == outputToDelete, "Can only delete the most recent  output.");
+        emit l2OutputDeleted(outputToDelete, latestBlockTimestamp);
+
+        delete l2Outputs[latestBlockTimestamp];
+        latestBlockTimestamp = latestBlockTimestamp - submissionInterval;
+    }
```
maurelian, line 135: Compared to the [original approach](https://gist.github.com/karlfloersch/02ce7938000680ce7820f39ba06bd9b1#file-staterootoracle-sol-L53-L60), this takes an output instead of a timestamp. I found the original a bit more error prone as there were more edge cases (ahead/behind the tip, not a multiple of the `submissionInterval`). I could also provided no arguments and just deleted the tip, but that seemed like it could lead to race conditions.

packages/contracts/contracts/test/L2OutputOracle.t.sol
```diff
@@ -226,4 +226,50 @@ contract L2OutputOracleTest is L2OutputOracle_Initializer {
...
+        vm.expectRevert("Ownable: caller is not the owner");
+        oracle.deleteL2Output(outputToDelete);
+    }
+
+    function testCannot_deleteL2Output_ifWrongOutput() external {
+        uint256 previousBlockTimestamp = oracle.latestBlockTimestamp() - submissionInterval;
+        bytes32 outputToDelete = oracle.getL2Output(previousBlockTimestamp);
+
+        vm.prank(sequencer);
+        vm.expectRevert("Can only delete the most recent  output.");
```
tynes, line 272: I notice there are 2 spaces in between `recent` and `output`

packages/contracts/contracts/L1/L2OutputOracle.sol
tynes: Generally looks good to me, this isn't a hot code path so I don't see a need for using `checked`. This function does not provide the ability to roll back an arbitrary number of outputs less than the fault proof window old, so we will either need to implement that functionality as well or put this contract behind a proxy so that functionality can be added in the future.

I think we should put this behind a delegatecall proxy for now

packages/contracts/contracts/L1/L2OutputOracle.sol
```diff
@@ -117,6 +120,20 @@ contract L2OutputOracle is Ownable {
...
     }
 
+    /**
+     * @notice Deletes the most recent output.
+     * @param _l2Output The value of the most recent output. Used to prevent erroneously deleting
+     *  the wrong root
+     */
+    function deleteL2Output(bytes32 _l2Output) external onlyOwner {
+        bytes32 outputToDelete = l2Outputs[latestBlockTimestamp];
+        require(_l2Output == outputToDelete, "Can only delete the most recent  output.");
```
tynes, line 130: nit: two spaces between recent and output and also curious why you use `require` here vs a custom error, I have a slight preference for consistency but open for using either when more appropriate

tynes:
Generally looks good to me, just a couple of nits

norswap:
A few quick questions:

- What's the use case for this?
    - Best guess: used by the challenge contract to remove all outputs following and including an output that has been proven faulty by a fault proof?
        - If that's the case, should we maybe include this logic directly in this contract?
- Has this been specced/explained?

tynes:
I agree with @norswap that its a good idea to specify this and I think as part of the spec we should define that this contract is behind a delegatecall based proxy so that its possible to upgrade away from this implementation and to another implementation that allows for fraud proof based reorgs

packages/contracts/contracts/L1/L2OutputOracle.sol
maurelian: I used `require` here simply because that's what was used in the rest of the contract, and I wanted to be consistent internally. That said, this is the only new contract still using `require`. Would you approve if I expand the scope of this PR with a new commit that introduces custom errors?

tynes: Yes I would! But if it's additional work that you don't want to do, that's cool too

packages/contracts/contracts/L1/L2OutputOracle.sol
maurelian: Since we'll only be checkpointing every 30 minutes or so, do you think we're likely to need that ability? 

Also, for this contract I'm open to a delegate proxy, but I could also imagine we just upgrade via social consensus (ie. to a new address).

maurelian:
@norswap I added specs. 
@tynes I'm going to forego adding custom errors on this PR.

tynes (approved):

tynes:
We will need to go through in a follow up PR, perhaps in a `deployment.md` or something that defines the deployment patterns of each contract, like if they are behind proxies and calling out what is owned clearly

PR #378, opened 2022-04-18
contracts: Fix reentrancy attack in WithdrawalsRelay

maurelian:
**Description**

Prevent reentrancy by observing the Checks-Effects-Interactions pattern.

I did not add a test due to the difficulty in setting one up (I'd need to generate a proof where the recipient address would be a malicious contract created for testing).


tynes (approved):

tynes:
We need some tooling for generating these kinds of proofs, it seems like it could be a `bin` distributed with the sdk, or a small go program that is a different binary but part of `opnode`

mslipper (approved):

PR #382, opened 2022-04-20
opnode: Better parsing of the TransactionDepositedEvent

trianglesphere:
This now asserts that the offset for the bytes field of the event
is the correct value rather than the previous incorrect check. This
check is not fully necessary, but good to have to validate that the
data is well formed.


protolambda (approved):
LGTM

PR #383, opened 2022-04-20
opnode: Do not drop all deposits on parsing error

trianglesphere:
Only drop the affected deposit. This does mean loss of funds, but
this is also a case that should never happen. In addition, halting
the chain means that while funds are safe, we are exposed to a
denial of service attack if it would be possible to cause a parsing
error from the deposit feed contract.

In addition, Change deposit gasLimit from uint256 to uint64

This prevents the user from setting a gas limit that the rollup node
is not able to parse.


trianglesphere:
I gotta update some more contracts stuff. Will force push when that's done

opnode/rollup/driver/step.go
```diff
@@ -89,11 +89,12 @@ func (d *outputImpl) createNewBlock(ctx context.Context, l2Head eth.L2BlockRef,
...
 	// be empty and no deposits will be derived.
-	deposits, err := derive.DeriveDeposits(receipts, d.Config.DepositContractAddress)
+	deposits, errs := derive.DeriveDeposits(receipts, d.Config.DepositContractAddress)
 	d.log.Info("Derived deposits", "deposits", deposits, "l2Parent", l2Head, "l1Origin", l1Origin)
-	if err != nil {
-		return l2Head, nil, fmt.Errorf("failed to derive deposits: %v", err)
+	for _, err := range errs {
+		d.log.Error("Failed to derive a deposit", "l1OriginHash", l1Origin.Hash, "err", err)
 	}
+	// TODO: Should we halt if len(errs) > 0? Opens up a denial of service attack, but prevents lockup of funds.
```
protolambda, line 97: Halting is better than losing funds in case of unexpected processing errors IMO. It should never happen though, so either way sounds fine, as long as we can catch this in testing. Curious what others think, but doesn't have to block this PR.

protolambda (approved):
LGTM

protolambda:
Can you update https://github.com/ethereum-optimism/optimistic-specs/blob/main/specs/deposits.md to reflect the `uint256 -> uint64 gasLimit` change?

protolambda (approved):
LGTM

PR #384, opened 2022-04-20
opnode: Fuzz manual ABI parsing

trianglesphere:
**Description**
This adds fuzzing for the following functions:
- `(*L1BlockInfo) MarshalBinary()`
- `(*L1BlockInfo) UnmarshalBinary`
- `UnmarshalLogEvent(ev *types.Log) (*types.DepositTx, error)`

The fuzzing is comparing the results to `abigen` and is designed to be flag errors that occur when the contract bindings are updated but the parsing logic is not.

It fuzz's each function for 10s during CI. Ideally we would have a separate long term fuzzing cluster, but during my testing 1s was enough (if properly configured) to catch the uint64 conversion bug in PR #383. The other (hacky) option is to preseed the corpus here.

**Metadata**
- Fixes https://github.com/ethereum-optimism/optimistic-specs/issues/405
- Fixes ENG-2197


mslipper (approved):

mslipper:
This looks good to me. Just be aware that the portal contract will be changing shortly once Mark and Kelvin wrap up the messengers.

PR #385, opened 2022-04-26
Makefile,ops: docker compose up instead of run to make devnet-down work

protolambda:
Minor fixes to makefile and devnet-up.sh, to make devnet-down work as expected. Previously the `ops_l2os_1` container wasn't being shut down because it was started separately (presumably because of env var option, but fixed that by passing along the env var in the yml environment definition).

### DIFF
--- a/ops/devnet-up.sh
+++ b/ops/devnet-up.sh
@@ -137,10 +137,8 @@ jq ". | .genesis.l1.hash = \"$(echo $L1_GENESIS | jq -r '.result.hash')\"" < ./o
 
 # Bring up everything else.
 cd ops
-echo "Bringing up rollup node..."
-docker-compose up -d opnode
-echo "Bringing up output submitter..."
-docker-compose run -e L2OO_ADDRESS="$L2OO_ADDRESS" -d --name ops-l2os-1 l2os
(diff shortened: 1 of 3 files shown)


ops/devnet-up.sh
```diff
@@ -137,10 +137,8 @@ jq ". | .genesis.l1.hash = \"$(echo $L1_GENESIS | jq -r '.result.hash')\"" < ./o
 
 # Bring up everything else.
 cd ops
-echo "Bringing up rollup node..."
-docker-compose up -d opnode
-echo "Bringing up output submitter..."
-docker-compose run -e L2OO_ADDRESS="$L2OO_ADDRESS" -d --name ops-l2os-1 l2os
+echo "Bringing up devnet..."
+L2OO_ADDRESS="$L2OO_ADDRESS" docker-compose up -d l2os
```
Inphi, line 141: The envar won't be forwarded unless it's added to the docker-compose.yml

protolambda: It's in there, see below

Inphi: you're right. my bad.

Inphi (approved):
thanks!

PR #386, opened 2022-04-26
opnode: Fix L1 Info Transactions

trianglesphere:
There are three issues being fixed:
1. Set the `from` field to the correct magic value (found by ToB)
2. Update the L2 EE to provide gas to deposits. Otherwise the
   deposit transactions immediatley out of gas and fail.
3. Update the manual ABI encoding to match what solidity expects.
   The previous version was similar to packed encodeding, but as
   such could not be parsed by solidity.

This also includes a regression test that the parsing of the L1 info
tx and what is recorded in the state for each block matches.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -165,7 +165,7 @@ require (
 	gopkg.in/yaml.v3 v3.0.0-20210107192922-496545a6307b // indirect
 )
 
-replace github.com/ethereum/go-ethereum v1.10.16 => github.com/ethereum-optimism/reference-optimistic-geth v0.0.0-20220411203319-ad60590374c8
+replace github.com/ethereum/go-ethereum v1.10.16 => github.com/ethereum-optimism/reference-optimistic-geth v0.0.0-20220427171107-d0070e0a6ead
 
 // For local debugging:
-//replace github.com/ethereum/go-ethereum v1.10.16 => ../go-ethereum
+// replace github.com/ethereum/go-ethereum v1.10.16 => ../go-ethereum
--- a/opnode/rollup/derive/invert.go
+++ b/opnode/rollup/derive/invert.go
@@ -1,26 +1,58 @@
 package derive
 
 import (
-	"encoding/binary"
+	"errors"
 	"fmt"
 	"math/big"
 
 	"github.com/ethereum/go-ethereum/common"
 )
 
+// setL1BlockValues(uint256 _number, uint256 _timestamp, uint256 _basefee, bytes32 _hash)
+type L1BlockInfo struct {
+	Number    uint64
+	Time      uint64
+	BaseFee   *big.Int
+	BlockHash common.Hash
+}
+
 // L1InfoDepositTxData is the inverse of L1InfoDeposit, to see where the L2 chain is derived from
 func L1InfoDepositTxData(data []byte) (nr uint64, time uint64, baseFee *big.Int, blockHash common.Hash, err error) {
-	if len(data) != 4+8+8+32+32 {
-		err = fmt.Errorf("data is unexpected length: %d", len(data))
-		return
+	info, err := L1InfoDepositTxDataToStruct(data)
+	return info.Number, info.Time, info.BaseFee, info.BlockHash, err
+}
+
+// L1InfoDepositTxDataToStruct is the inverse of L1InfoDeposit, to see where the L2 chain is derived from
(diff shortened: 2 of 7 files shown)


go.mod
```diff
@@ -66,7 +66,7 @@ require (
 	gopkg.in/yaml.v3 v3.0.0-20210107192922-496545a6307b // indirect
 )
 
-replace github.com/ethereum/go-ethereum v1.10.16 => github.com/ethereum-optimism/reference-optimistic-geth v0.0.0-20220411203319-ad60590374c8
+replace github.com/ethereum/go-ethereum v1.10.16 => github.com/ethereum-optimism/reference-optimistic-geth v0.0.0-20220426163645-51a763f5c46d
```
protolambda, line 69: We should merge the fix commit into the `optimism-prototype` commit, and then use that. It's already in the audit branch anyway, the fix should be in the default branch too.

protolambda (commented):
LGTM, but need to port that geth fix into the optimism-prototype default branch.

opnode/rollup/derive/invert.go
```diff
@@ -1,26 +1,58 @@
...
 import (
-	"encoding/binary"
+	"errors"
 	"fmt"
 	"math/big"
 
 	"github.com/ethereum/go-ethereum/common"
 )
 
+// setL1BlockValues(uint256 _number, uint256 _timestamp, uint256 _basefee, bytes32 _hash)
```
tynes, line 11: Will we want to add `Random` here as well? I know that is out of scope for this PR, but it seems like doing so eventually will be painful as in a lot of code to change to handle it

opnode/rollup/derive/invert.go
```diff
@@ -1,26 +1,58 @@
...
+	out := L1BlockInfo{}
+	if len(data) != 4+32+32+32+32 {
+		return out, fmt.Errorf("data is unexpected length: %d", len(data))
+	}
+
+	// Number
+	offset := 4 // Selector hash. Should check
+	number := new(big.Int)
+	number.SetBytes(data[offset : offset+32])
+	if !number.IsUint64() {
```
tynes, line 36: A malicious sequencer can cause this to return an error, what are the implications of that?

opnode/test/system_test.go
```diff
@@ -413,3 +417,134 @@ func TestMissingBatchE2E(t *testing.T) {
...
+	if err != nil {
+		return derive.L1BlockInfo{}, fmt.Errorf("failed to get block hash: %w", err)
+	}
+	out.BlockHash = common.BytesToHash(blockHashBytes[:])
+
+	return out, nil
+}
+
+func TestL1InfoContract(t *testing.T) {
+	log.Root().SetHandler(log.DiscardHandler()) // Comment this out to see geth l1/l2 logs
```
tynes, line 464: Perhaps we can put this behind a flag that turns logs on and off and use the `-args` flag of `go test`?

https://pkg.go.dev/cmd/go/internal/test

tynes: Out of scope of this PR, just a thought that could apply across the test suite

opnode/rollup/derive/invert.go
tynes: Before we get to this codepath, we would have already checked the signature on the unsafe block received via p2p. So the worst thing that could happen is a malicious sequencer halts unsafe block distribution and then the verifier would have to wait until the transactions are batch submitted. So the protection against this is governance voting the malicious sequencer out. A sequencer would not want to do this because they want economic activity on their chain and doing this would result in degraded experience for users. Does that sound correct to you as a security model?

trianglesphere: Yep. I guess at the same time the sequencer could be building the chain at the same time.
It's probably worth fixing this at the ABI level, but I didn't quite get around to it in this PR- I think we need a more comprehensive look at this as well.

opnode/rollup/derive/invert.go
protolambda: It's already part of the L2 payload-attributes, and exposed via the `RANDOM` opcode (since l2 doesn't have its own randomness beacon like L1 does post-merge)

protolambda (approved):
LGTM. And opened #392 to extend the attributes with seq-nr to fix another bug and refactor some of this code (use binary marshal/unmarshal etc.)

PR #391, opened 2022-04-28
opnode: Fix Incorrect error handling when creating an L2 block

trianglesphere:
**Description**

This is an issue ToB identified (issue #8). What happens is that
the ethereum.NotFound error would never be returned, but in general
if there is an error in the lookup, the state loop should keep
going.

This also properly wrap errors in l1/source.go.

Note: There is an argument for halting L2 block production if it cannot
find the next L1 origin; however, it should be clearly communicated with
an error.

**Metadata**
- Fixes: ENG-2127

### DIFF
--- a/opnode/l1/source.go
+++ b/opnode/l1/source.go
@@ -124,7 +124,7 @@ type Source struct {
 
 func NewSource(client RPCClient, log log.Logger, config *SourceConfig) (*Source, error) {
 	if err := config.Check(); err != nil {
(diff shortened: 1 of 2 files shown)


protolambda (commented):
Change looks good. It would be nice if we can test this, e.g. with some mock L1 source that returns an error.

mslipper:
Spoke to Proto - we're good to merge this and handle the testing separately. I'm tracking the testing in a separate Linear.

PR #392, opened 2022-04-28
bugfix: createNewBlock must use correct sequence number, add seq nr to L1 info

protolambda:
## Bug

The safeBlockHash is a confirmation thing, and not always consistent with the L1 origin. It is only an argument to help with forkchoice updates were we need to hint at confirmation safety to the engine. But we were using it to compute a "sequence number" (number of blocks since last L1 origin change), which ended up being wrong a lot of the time after the slightest irregular confirmation. The block could then still be created, but would fail to match verification (deposits were different due to different source hash, while L1 info and L2 block height were otherwise the same).

## Fix

Add the sequence-number field to the L1 block info, as previously intended with https://github.com/ethereum-optimism/optimistic-specs/pull/196

The seq-nr is now also part of the `L2BlockRef`, just like the L1 origin is.

This seq-nr is now used to derive the deposit source hash from, making the derivation deterministic again (safeBlockHash was close, but not always accurate, since it may change in other independent ways)

Additionally this refactors the L1 info into a struct, in both Go and solidity:
- in Go it helps avoid having too many return arguments
- ~~in Solidity it helps pack the calldata into a smaller amount~~. TODO: changed it to struct-typed input, but the arguments are not packed by solidity yet. Any way to get packing to work elegantly? Edit: not likely, without yul or manual bitshifts/masks etc.

This PR is based on top of #386 that fixes some of the previous ABI encoding issues


opnode/test/system_test.go
```diff
@@ -100,8 +100,8 @@ func defaultSystemConfig(t *testing.T) SystemConfig {
 			},
 		},
 		Loggers: map[string]log.Logger{
-			"verifier":  testlog.Logger(t, log.LvlError),
-			"sequencer": testlog.Logger(t, log.LvlError),
+			"verifier":  testlog.Logger(t, log.LvlInfo),
+			"sequencer": testlog.Logger(t, log.LvlInfo),
```
protolambda, line 104: I always end up changing these log levels to get info for my tests. Can we just default them to info level?

opnode/test/system_test.go
```diff
@@ -299,6 +299,8 @@ func TestSystemE2E(t *testing.T) {
 	seqBlock, err := l2Seq.BlockByNumber(context.Background(), receipt.BlockNumber)
 	require.Nil(t, err)
 	require.Equal(t, verifBlock.Hash(), seqBlock.Hash(), "Verifier and sequencer blocks not the same after including a batch tx")
+
+	time.Sleep(1 * time.Minute)
```
protolambda, line 303: will remove this before merge. Useful to keep it open a bit longer to see everything is healthy.

tynes:
> Any way to get packing to work elegantly?
Would need to use `yul` to unpack it and would lose the ability to have a nice `abigen` generated binding. I think each being a `uint64` could work

packages/contracts/contracts/L2/L1Block.sol
```diff
@@ -16,12 +16,14 @@ contract L1Block {
...
     bytes32 public hash;
+    uint256 public seqnr;
 
     function setL1BlockValues(
         uint256 _number,
         uint256 _timestamp,
         uint256 _basefee,
-        bytes32 _hash
+        bytes32 _hash,
+        uint256 _seqnr
```
tynes, line 26: Is it always expected that `seqnr` increments by 1? We could `revert` if it doesn't increment correctly but we also have `number` which we could do that on

opnode/test/system_test.go
protolambda: won't change it for now, we can revisit logging separately

packages/contracts/contracts/L2/L1Block.sol
```diff
@@ -16,12 +16,14 @@ contract L1Block {
     uint256 public timestamp;
     uint256 public basefee;
     bytes32 public hash;
+    uint256 public seqnr;
```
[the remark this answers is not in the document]
protolambda: I don't think we have to. That same randomness can already be accessed with an opcode (and it's part of the L2 block header)

opnode/rollup/driver/step.go
```diff
@@ -279,6 +286,28 @@ func attributesMatchBlock(attrs *l2.PayloadAttributes, parentHash common.Hash, b
 	}
 	for i, otx := range attrs.Transactions {
 		if expect := block.Transactions[i]; !bytes.Equal(otx, expect) {
+			//if otx[0] == types.DepositTxType && expect[0] == types.DepositTxType {
```
tynes, line 289: Did you mean to leave this dead code in?

packages/contracts/contracts/L2/L1Block.sol
tynes: Got it, you should add that explanation to the specs since that clarifies its usage quite a bit

opnode/rollup/driver/step.go
protolambda: thanks for catching that, that's some debug code, will remove

protolambda: removed it, please have a look again.

specs/deposits.md
```diff
@@ -193,6 +193,9 @@ A predeployed contract on L2 at address `0x4200000000000000000000000000000000000
 certain block variables from the corresponding L1 block in storage, so that they may be accessed
 during the execution of the subsequent deposited transactions.
 
+A `seqnr` attribute is stored along with the L1 attributes to track the L2 block distance
+to the L2 block height that the L1 attributes were first introduced.
```
protolambda, line 197: @tynes spec is already updated. Suggestions welcome if you want to word it differently.

packages/contracts/contracts/L2/L1Block.sol
protolambda: See below review comment on spec doc.

opnode/rollup/derive/l1_block_info.go
```diff
@@ -0,0 +1,70 @@
...
+	if !bytes.Equal(data[offset:offset+24], padding[:]) {
+		return fmt.Errorf("l1 info time exceeds uint64 bounds: %x", data[offset:offset+32])
+	}
+	offset += 32
+	info.BaseFee = new(big.Int).SetBytes(data[offset : offset+32])
+	offset += 32
+	info.BlockHash.SetBytes(data[offset : offset+32])
+	offset += 32
+	info.SeqNr = binary.BigEndian.Uint64(data[offset+24 : offset+32])
+	if !bytes.Equal(data[offset:offset+24], padding[:]) {
```
tynes, line 59: Why not make the abi type be a `uint64` as to prevent a malicious sequencer from causing issues here?

specs/deposits.md
tynes: I was thinking that your wording above was much more clear :)

It resets to 0 whenever the L1 origin fields change. This allows it to uniquely identify L2 blocks in the same epoch, without needing the context of the actual L2 block height.

packages/contracts/contracts/L2/L1Block.sol
protolambda, line 19: number, timestamp and seqnr could all be `uint64` instead of `uint256`. This would also help fix a concern from @tynes I think. Anyone opposed to changing the types?

opnode/rollup/derive/l1_block_info_test.go
```diff
@@ -5,6 +5,8 @@ import (
 	"math/rand"
 	"testing"
 
+	"github.com/stretchr/testify/require"
+
```
trianglesphere, line 9: We should be fuzzing the following:
- That `Marshall/Unmarshall` never panic
- That `Marshall` and `Unmarshall` match abigen

specs/deposits.md
```diff
@@ -193,6 +193,9 @@ A predeployed contract on L2 at address `0x4200000000000000000000000000000000000
 certain block variables from the corresponding L1 block in storage, so that they may be accessed
 during the execution of the subsequent deposited transactions.
 
+A `seqnr` attribute is stored along with the L1 attributes to track the L2 block distance
```
trianglesphere, line 196: I really don't like `seqnr` as the name in the specs/contract. We don't need to limit identifiers to [6 characters](https://stackoverflow.com/questions/38035628/c-why-did-ansi-only-specify-six-characters-for-the-minimum-number-of-significa). In code it's a little better, but spelling out `sequenceNumber` is not that bad.

trianglesphere (commented):
I'm in favor of the change, but I don't love the name + would like to see differential testing against `abigen`.

packages/contracts/contracts/L2/L1Block.sol
tynes: I'm into changing the types to `uint64`s and also into tightly packing the calldata but consider that a nice to have, would require a bunch of changes in tests to do so. If we really want to optimize then we don't even need a function selector to be honest and just use a `fallback` function + a solidity lib that can build the calldata easily for the tests

specs/deposits.md
norswap, line 197: I think a clearer description would be

> A `seq_number` attribute is stored along with the L1 attributes to track the position of the L2 block within its epoch (starting at 0).

If you want you can clarify:
> Said otherwise, the `seq_number` is the the L2 block distance to the L2 block height that the L1 attributes were first introduced.

norswap:
Took me a bit (and some help from Proto) to figure the issue from the description, so to clarify: we were using the safe block head instead of the L1 origin of the L2 head to derive the sequence number (the number of an L2 block within its epoch).

packages/contracts/contracts/L2/L1Block.sol
trianglesphere: TBH I wouldn't mind handwritten assembly for this contract. It seems like a lot of the solidity stuff (plus the ABI) makes the bytecode much longer than it needs to be.

opnode/rollup/derive/l1_block_info_test.go
protolambda: Going to add fuzzing as a new issue. This bugfix is important and has been stuck for long enough. We already have your e2e test to verify the minimum works.

packages/contracts/contracts/L2/L1Block.sol
protolambda: Going for `uint64` for now. We can do the fancy calldata packing and fallback function thing at a later stage if we really want to. And if we're planning to add more logic to this contract (like the history accumulator), then I do like to maintain the ability to also write some regular solidity in the contract. I guess we can combine it though.

opnode/rollup/derive/l1_block_info.go
```diff
@@ -0,0 +1,70 @@
...
+	"github.com/ethereum/go-ethereum/common"
+)
+
+// L1BlockInfo presents the information stored in a L1Block.setL1BlockValues call
+type L1BlockInfo struct {
+	Number         uint64
+	Time           uint64
+	BaseFee        *big.Int
+	BlockHash      common.Hash
+	SequenceNumber uint64
```
trianglesphere, line 18: Can you add a callout that this is a property of the L2 block unlike the rest of the properties?

tynes:
I'm getting a diff on top of this when running `make bindings` from `opnode/contracts`

protolambda:
```
abigen --version
abigen version 1.10.17-stable

forge --version
forge 0.2.0 (6130af7 2022-04-06T00:03:48.094239707+00:00)

solc --version
solc, the solidity compiler commandline interface
Version: 0.8.10+commit.fc410830.Linux.g++
```
I might be running something wrong/outdated? What do you run?
Feel free to push the corrected bindings if mine are wrong (and we should add something to CI to detect what's wrong/correct...)

tynes:
> ```
> abigen --version
> abigen version 1.10.17-stable
> 
> forge --version
> forge 0.2.0 (6130af7 2022-04-06T00:03:48.094239707+00:00)
> 
> solc --version
> solc, the solidity compiler commandline interface
> Version: 0.8.10+commit.fc410830.Linux.g++
> ```
> 
> I might be running something wrong/outdated? What do you run? Feel free to push the corrected bindings if mine are wrong (and we should add something to CI to detect what's wrong/correct...)

Josh was experiencing the same thing, a diff from me. Unsure why this is happening

tynes:
Follow up: I am using the same version of `abigen` and a similar version of forge

Opened this to address: https://github.com/ethereum-optimism/optimistic-specs/pull/406

@trianglesphere do you get a diff when running the `make` commands locally on top of this PR?

trianglesphere:
I also get a diff

```
abigen --version
abigen version 1.10.16-unstable

forge --version
forge 0.2.0 (a97090d 2022-05-03T21:27:48.936399+00:00)

solc --version
solc, the solidity compiler commandline interface
Version: 0.8.10+commit.fc410830.Darwin.appleclang
```

trianglesphere:
When I changed forge versions, I did get a new diff only in the byte code. I'm not sure exactly where though.
```
forge --version
forge 0.2.0 (6130af7 2022-05-03T23:50:03.259959+00:00)
# forge 0.2.0 (6130af7 2022-04-07T00:06:13.404059+00:00) - original date
```

tynes (approved):

PR #388, opened 2022-04-26
opnode: attach p2p host and discovery to rollup node, update e2e test

protolambda:
This PR implements the next p2p todo: attaching the p2p host and discovery service to the node.

The rollup node config can now be set up with a "prepared" p2p setup, to attach a host to it that's part of a "mock net", so we can test p2p communication in different topologies easily, without attaching to the actual localhost TCP and UDP interfaces etc.

See #163

PR #394, opened 2022-04-30
P2P gossip of blocks

protolambda:
~~Builds on top of #388~~ (now rebased on `main`)

See #163 for tracking

Changes:
- gossip subscription filter
- gossip message id function
- gossip spec topic naming format
- minor additions to rollup chain config
  - l2 chain ID
  - sequencer p2p address is part of chain config for now. We can make it replaceable (with L1 deposit maybe) and have multiple different ones in a later PR
- init gossipsub with all options, except peer scoring
- init blocks topic
- implement blocks topic validator
  - uncompress the block
  - split signature and payload
  - verify payload time etc.
  - verify signature
  - specs for validation, signature, etc.
- implement some subscriber work to move the block to the node main loop.
  - not sure yet if we want to fan out to attached drivers in the main loop, or use the geth even/subscribers code to do it for us, like we do with new L1 block heads
- update p2p spec to define p2p signing over block messages.
- implement SSZ serialization/deserialization for ExecutionPayload, with fuzz testing
- Refactor node initialization
  - No more event.Feed and subscription mess
  - Clean separation of domains (l1 client, l2 engines, rollup server, p2p signer, p2p stack)
  - Close node resources properly if node init fails
- Gossip In/Out interfaces
- Sequencer publishes blocks now
- Opnode forwards sequencer blocks to p2p stack (if enabled), to be signed and published
- Implement local p2p signer, and abstraction for later remote signer
- Add new E2E test that confirms a tx and syncs the verifier faster than L1 could
- Update devnet to enable p2p (still single op-node though)
- ~~update driver state tests with unsafe-l2 payload step type~~: driver tests use mock chain, no full payloads, not the right place to test.
- Add tracer to opnode, so we can check what's being published/received in which node during e2e testing.


protolambda:
Tests are passing, but like `main`, integration tests are running for 1h+ and so I canceled the workflows. Will focus on e2e tests for now while we fix integration tests outside of this PR.

protolambda:
Still working on more testing, but ready for review. `TestSystemMockP2P` in the Go e2e test sets up a verifier/sequencer pair, connected via a "mocknet" (test network layer), with slow L1 time and a tx that confirms super fast, thanks to the p2p block distribution :)

protolambda:
updating devnet now, broke it with some cli flag change

opnode/flags/p2p_flags.go
```diff
@@ -192,6 +192,14 @@ var (
 		Value:     "opnode_discovery_db",
 		EnvVar:    p2pEnv("DISCOVERY_PATH"),
 	}
+	SequencerP2PKeyFlag = cli.StringFlag{
```
tynes, line 195: I know that geth uses this same scheme, do you think we could use an encrypted keystore as well so that we can fully migrate to only using encrypted keystores?

opnode/l2/api.go
```diff
@@ -111,6 +115,48 @@ func (payload *ExecutionPayload) ID() eth.BlockID {
...
+
+	header := types.Header{
+		ParentHash:  payload.ParentHash,
+		UncleHash:   types.EmptyUncleHash,
+		Coinbase:    payload.FeeRecipient,
+		Root:        common.Hash(payload.StateRoot),
+		TxHash:      txHash,
+		ReceiptHash: common.Hash(payload.ReceiptsRoot),
+		Bloom:       types.Bloom(payload.LogsBloom),
+		Difficulty:  big.NewInt(0), // zeroed, proof-of-work legacy
```
tynes, line 146: Thoughts on using `common.Big0` to save an alloc here?
https://github.com/ethereum/go-ethereum/blob/86d547707965685cef732aa28c15e6811ea98408/common/big.go#L26

opnode/flags/p2p_flags.go
protolambda: The "sequencer p2p key" is the key used to sign application layer data, and should use an encrypted keystore, like the batchsubmitter and output submitter also should.

The "p2p node key" (different!) used for the node identity is much less critical, and every user has one. I think it makes sense to follow the geth (and eth2) pattern there, to generate one if it doesn't exist, and persist it as file to disk.

opnode/flags/flags.go
```diff
@@ -63,13 +63,6 @@ var (
 		EnvVar: prefixEnvVar("SEQUENCING_ENABLED"),
 	}
 
-	// TODO: move batch submitter to stand-alone process
```
trianglesphere, line 66: nit: Unrelated to p2p work.

opnode/l2/api.go
```diff
@@ -2,9 +2,13 @@
 package l2
 
 import (
+	"bytes"
 	"fmt"
+	"math/big"
 	"reflect"
 
+	"github.com/ethereum/go-ethereum/trie"
```
trianglesphere, line 10: We should probably be using `goimports` or something, but your imports here are split from the rest of the `go-ethereum` imports

opnode/l2/ssz_test.go
```diff
@@ -0,0 +1,80 @@
...
+		var payload ExecutionPayload
+		err := payload.UnmarshalSSZ(uint32(len(data)), bytes.NewReader(data))
+		if err != nil {
+			// not every input is a valid ExecutionPayload, that's ok. Should just not panic.
+			return
+		}
+	})
+}
+
+// FuzzExecutionPayloadMarshalUnmarshal checks that our SSZ encoding>decoding round trips properly
```
trianglesphere, line 24: What would it take to do differential fuzzing? i.e. compare our marshall / unmarshall against a known library

opnode/l2/ssz.go
```diff
@@ -0,0 +1,225 @@
...
+
+	buf := *payloadBufPool.Get().(*[]byte)
+	if uint32(cap(buf)) < scope {
+		buf = make([]byte, scope)
+	} else {
+		buf = buf[:scope]
+	}
+	defer payloadBufPool.Put(&buf)
+
+	offset := uint32(0)
```
trianglesphere, line 67: Is there a way to make the compiler generate this? Another option is build an API that does the copy and bumps the offset at the same time.

Basically, the amount of code that we have that looks like this is terrifying.

opnode/l2/ssz.go
```diff
@@ -0,0 +1,225 @@
...
+	copy(buf[offset:offset+32], payload.ParentHash[:])
+	offset += 32
+	copy(buf[offset:offset+32], payload.FeeRecipient[:])
+	offset += 20
```
trianglesphere, line 71: Did you switch the increase of the offset with where to copy the bytes to?

opnode/node/node.go
```diff
@@ -65,210 +80,297 @@ func New(ctx context.Context, cfg *Config, log log.Logger, appVersion string) (*
...
+		if closeErr := n.Close(); closeErr != nil {
+			return nil, multierror.Append(err, closeErr)
+		}
+		return nil, err
+	}
+	return n, nil
+}
+
+func (n *OpNode) init(ctx context.Context, cfg *Config) error {
+
```
trianglesphere, line 102: This doesn't need a newline after every block

opnode/rollup/driver/state.go
```diff
@@ -282,6 +313,39 @@ func (s *state) handleEpoch(ctx context.Context) (bool, error) {
...
+	if s.l2SafeHead.Number > uint64(payload.BlockNumber) {
+		s.log.Info("ignoring unsafe L2 execution payload, already have safe payload", "id", payload.ID())
+		return nil
+	}
+
+	if uint64(payload.Timestamp) < s.l2SafeHead.Time {
+		s.log.Warn("received payload rewinds L2 chain beyond current L2 safe head")
+	}
+
+	if uint64(payload.Timestamp) > s.l2Head.Time+s.Config.BlockTime {
```
trianglesphere, line 326: Is the intention to only advance on block time at a time? But this does allow reorgs, right?

opnode/test/system_test.go
```diff
@@ -481,7 +487,40 @@ func TestSystemMockP2P(t *testing.T) {
...
+		Gas:       21000,
+	})
+	err = l2Seq.SendTransaction(context.Background(), tx)
+	require.Nil(t, err, "Sending L2 tx to sequencer")
+
+	// Wait for tx to be mined on the L2 sequencer chain
+	receiptSeq, err := waitForTransaction(tx.Hash(), l2Seq, 3*time.Duration(cfg.RollupConfig.BlockTime)*time.Second)
+	require.Nil(t, err, "Waiting for L2 tx on sequencer")
+
+	// Wait until the block it was first included in shows up in the safe chain on the verifier
```
trianglesphere, line 519: This test does fail if p2p is disabled, but because of a timeout + how times are constructed. We need to add a TODO to make sure that when we can tell the difference between unsafe and safe blocks, we check that this is the result of the p2p network.

trianglesphere (changes requested):
I've reviewed everything but the `p2p` package. I think that this PR is doing too many orthogonal things and could be split up a bit.

One major concern that I have is the amount of handwritten byte parsing code. I don't love reflection based APIs but I do tend to trust them more than the hand written code. My preference would be code generation against a defined struct. We need test vectors as well as differential fuzzing to get more confidence in the parsing code.

protolambda:
@trianglesphere regarding the SSZ code: the alternative is to import a library, but we're only dealing with 1 type. I did implement several already, and used to maintain the SSZ specs: https://github.com/ethereum/consensus-specs/issues/2138

Specifically, we could use helper funcs of: https://github.com/protolambda/ztyp

Implemented the type too, in ZRNT (my Go implementation of eth2 specs, passing all standard consensus / ssz tests): https://github.com/protolambda/zrnt/blob/5447d45e7e85ddff746721baaca3eced8ce67a6f/eth2/beacon/common/execution.go#L304

Obviously I'm biased to use the above library, but I thought it would be better to avoid creating a dependency on my non-optimism code. The encoding/decoding is simple though, and I added fuzz tests. If you prefer the dependency on ZRNT, or maybe just ZTYP, we can do that too.

opnode/l2/api.go
protolambda: I mostly just let the IDE handle imports, and run `golangci-lint run -E asciicheck,goimports,misspell --fix`. Will clean up imports, but we should update the `linting.md` doc if it needs to be different.

opnode/l2/ssz_test.go
protolambda: Already implemented several libraries, including 2 Go libraries (and 3 different ways of representing the type structure), so there are ways... That said, fuzzing itself is good enough for initial code imho.

opnode/l2/ssz.go
protolambda: Prysm uses code-generated SSZ marshal/unmarshal code. But historically there were bugs in that, the maintainer never gave it much love, and it's protobuf specific and quite unreadable. If we want optimized code I'd rather implement it like this.

opnode/l2/ssz.go
protolambda: Not sure if I understand your question here

opnode/rollup/driver/state.go
protolambda: We should allow reorgs, but not too far in the future. This is not consensus code, so a heuristic to not try and add blocks beyond the expected timestamp seemed ok. I can remove it if you prefer that blocks received on p2p can trigger more.

opnode/flags/flags.go
protolambda: It's an outdated comment that needed to be removed, and I wanted to avoid confusion with the sequencer p2p key by cleaning up any unused references to the batch submitter key.

opnode/p2p/gossip.go
```diff
@@ -0,0 +1,403 @@
...
+// BuildMsgIdFn builds a generic message ID function for gossipsub that can handle compressed payloads,
+// mirroring the eth2 p2p gossip spec.
+func BuildMsgIdFn(cfg *rollup.Config) pubsub.MsgIdFunction {
+	return func(pmsg *pb.Message) string {
+		valid := false
+		var data []byte
+		// If it's a valid compressed snappy data, then hash the uncompressed contents.
+		// The validator can throw away the message later when recognized as invalid,
+		// and the unique hash helps detect duplicates.
+		dLen, err := snappy.DecodedLen(pmsg.Data)
```
mslipper, line 70: Is this immune to zip-bomb like attacks?

opnode/p2p/gossip.go
```diff
@@ -0,0 +1,403 @@
...
+	if err != nil {
+		panic(fmt.Errorf("failed to set up block height LRU cache: %v", err))
+	}
+
+	return func(ctx context.Context, id peer.ID, message *pubsub.Message) pubsub.ValidationResult {
+		// [REJECT] if the compression is not valid
+		outLen, err := snappy.DecodedLen(message.Data)
+		if err != nil {
+			log.Warn("invalid snappy compression length data", "err", err, "peer", id)
+			return pubsub.ValidationReject
```
mslipper, line 175: What happens under the hood when you return these validation values?

opnode/p2p/gossip.go
```diff
@@ -0,0 +1,403 @@
...
+		log.Debug(msg, "result", validationResultString(res), "from", src)
+		return res
+	}
+}
+
+func BuildBlocksValidator(log log.Logger, cfg *rollup.Config) pubsub.ValidatorEx {
+
+	// Seen block hashes per block height
+	// uint64 -> []common.Hash
+	blockHeightLRU, err := lru.New(100)
```
mslipper, line 165: Is this LRU shared among all peers, or just one? If it's shared among multiple peers, isn't there a race condition in which multiple peers could try and append to the `seen` slice at the same time? Or does the LRU copy it on the way out?

opnode/p2p/gossip.go
protolambda: See the `dLen <= MaxGossipSize` below. It detects the length before actually allocating or uncompressing it. If the length is too large it will handle it like a message with invalid compression: the data stays as-is, and the message ID is computed over the data with a different domain (to avoid collisions with valid compressed data). The validation pipeline will later take care of the message being too large / compressed wrong.

opnode/p2p/gossip.go
protolambda: Good catch, the LRU is shared and can be concurrently called by at most 4 validator routines, so there's a race condition indeed. See [0fca724](https://github.com/ethereum-optimism/optimistic-specs/pull/394/commits/0fca7248b28f90350bfe99d45ab3ae3e4fbc72ed) with fix + related misc improvements.

opnode/rollup/driver/state.go
protolambda: Thought about this more and think it's fine to remove the heuristics here, as the execution engine will handle them anyway, and the heuristics were not that nice. Removed in commit [a069290](https://github.com/ethereum-optimism/optimistic-specs/pull/394/commits/a069290a999cea7290ce35aad32d3fd200cffcd6), and added a comment about reorg/sync behavior.

protolambda:
Integration tests failed on a withdrawal test, but passed in another job. Re-running the failed case now. The withdrawal issue is outside of the scope of this PR

mslipper (approved):

.gitignore
```diff
@@ -10,4 +10,7 @@ bin
 integration-tests/cache/
 integration-tests/artifacts/
 **/*.env
-.devnet
\ No newline at end of file
+.devnet
+
+# Ignore local fuzzing results
+**/testdata/fuzz/
```
tynes, line 16: If I understand correctly, failures are added to the corpus which then seed future tests and they are held at this directory. Wouldn't we want to keep those seeds?

protolambda: Maybe, I'm not so sure if we want it here in the source-code, or in a separate repo focused on fuzzing maybe. cc @trianglesphere do you want to check these files into git, if there are any?

opnode/p2p/signer.go
```diff
@@ -0,0 +1,100 @@
...
+
+type SignerSetup interface {
+	SetupSigner(ctx context.Context) (Signer, error)
+}
+
+// LoadSignerSetup loads a configuration for a Signer to be set up later
+func LoadSignerSetup(ctx *cli.Context) (SignerSetup, error) {
+	keyFile := ctx.GlobalString(flags.SequencerP2PKeyFlag.Name)
+	if keyFile != "" {
+		// Mnemonics are bad because they leak *all* keys when they leak.
```
tynes, line 86: Mnemonics are good because its easier to offline backup words compared to a hex string. There are safe and unsafe ways to use any type of key

opnode/l2/ssz.go
trianglesphere: The line `copy(buf[offset:offset+32], payload.FeeRecipient[:])` is correct because `copy` copies the min of `len(src), len(dst)`. You're copying 20 bytes into a 32 byte buffer.

.gitignore
trianglesphere: I don't think they should be added to git. It'd probably be good practice to add them with `f.Add(...)` at the start.
The long term solution is a fuzzing cluster.

opnode/p2p/gossip.go
```diff
@@ -0,0 +1,427 @@
...
+	}
+}
+
+func BuildGlobalGossipParams(cfg *rollup.Config) pubsub.GossipSubParams {
+	params := pubsub.DefaultGossipSubParams()
+	params.D = 8                                      // topic stable mesh target count
+	params.Dlo = 6                                    // topic stable mesh low watermark
+	params.Dhi = 12                                   // topic stable mesh high watermark
+	params.Dlazy = 6                                  // gossip target
+	params.HeartbeatInterval = 500 * time.Millisecond // frequency of heartbeat, seconds
```
trianglesphere, line 106: Doc comment does not match the actual interval.

opnode/p2p/gossip.go
```diff
@@ -0,0 +1,427 @@
...
+			return pubsub.ValidationReject
+		}
+		*res = data // if we ended up growing the slice capacity, fine, keep the larger one.
+
+		// message starts with compact-encoding secp256k1 encoded signature
+		signatureBytes, payloadBytes := data[:65], data[65:]
+
+		// [REJECT] if the block encoding is not valid
+		var payload l2.ExecutionPayload
+		if err := payload.UnmarshalSSZ(uint32(len(payloadBytes)), bytes.NewReader(payloadBytes)); err != nil {
```
trianglesphere, line 220: We do a `[]byte -> reader -> []byte -> copy to dest` decoding here because of the fact that we Marshall with readers/writers. It'd probably be more performant to do a reader based API on the decoding side.

mslipper:
This is a tricky one. We should evaluate personal projects the same way we evaluate any other library. After taking a look at Proto's library, here's where I stand:

1. The library is large, and does a lot more than we need. The implementation here is only ~200 lines of mostly byte offset math. This is much easier to reason about.
2. Adopting the library will create an additional dependency on Proto, and Proto is already pretty tapped.
3. The implementation here is so small we can include it in our upcoming security audits.
4. The implementation here is well tested, and fuzzed.

Over the long term I agree we should maintain an SSZ library. That library could be Proto's. However, I don't think we should make that a blocker for Bedrock.

opnode/rollup/driver/state.go
```diff
@@ -282,6 +313,31 @@ func (s *state) handleEpoch(ctx context.Context) (bool, error) {
 
 }
 
+func (s *state) handleUnsafeL2Payload(ctx context.Context, payload *l2.ExecutionPayload) error {
+	if s.l2SafeHead.Number > uint64(payload.BlockNumber) {
+		s.log.Info("ignoring unsafe L2 execution payload, already have safe payload", "id", payload.ID())
+		return nil
+	}
+
+	// Not that the payload may cause reorgs. The l2SafeHead may get out of sync because of this.
```
trianglesphere, line 322:
```suggestion
	// Note that the payload may cause reorgs. The l2SafeHead may get out of sync because of this.
```

opnode/rollup/driver/state.go
```diff
@@ -282,6 +313,31 @@ func (s *state) handleEpoch(ctx context.Context) (bool, error) {
...
+	// The engine should never reorg past the finalized block hash however.
```
trianglesphere, line 323: We actually (correctly I believe), don't allow reorgs past the safe block head.

trianglesphere (commented):
I found a couple more small things.

After looking through everything, I'm fine keeping this ssz implementation, but I do want to draw attention to line 70 of ssz.go - I believe it should copying from the offset to offset + 20 instead of offset + 32. I doesn't currently cause problem, but I don't love that it's there. That's the only blocker remains from my previous review.

opnode/p2p/gossip.go
protolambda: The doc comment is copy-paste from the spec. Spec says `0.5`, but Go uses `time.Duration` instead. Will fix the comment

opnode/p2p/gossip.go
protolambda: All the underlying `[]byte` are pooled and recycled, I don't think it's that bad with performance. Repeated `Read` calls to an interface may even be slower.

opnode/rollup/driver/state.go
protolambda: The L1 safe head itself can still reorg, so even if L2 reorgs due to a block from p2p I don't think there's much of a difference here.

protolambda:
Addressed the comments + fixed that ssz thing (my bad, copy range was larger than it should have been, although effectively the same behavior, thus not catched in test)

opnode/rollup/driver/state.go
trianglesphere: In processBlock, we reject payloads behind the safe block head. You're right that the engine will not got reorg past the finalized hash, but this comment is still a little misleading about what exact reorgs it can cause.

protolambda: If you're referring to the remaining `s.l2SafeHead.Number > uint64(payload.BlockNumber)` heuristic in `handleUnsafeL2Payload`: a reorg past the safe block may still happen using a block newer than the safe-head if the common ancestor is before the safe head. So I think the comment is accurate that the `l2SafeHead` can get out of sync. The `processBlock` itself doesn't do any checks like that.

trianglesphere: You're right that is is on a different chain. I don't think that we should allow that reorg, but that might take more design to do.

trianglesphere (approved):
Approving this, it LGTM.

Follow ups for later:
- Fuzzing cluster / regression tests based on fuzz results
- Unsafe block distribution causing reorgs past the safe head (probably incorrect): https://github.com/ethereum-optimism/optimistic-specs/pull/394#discussion_r869603281
- SSZ encoding/decoding - library / reader based approach over offsets / increasing confidence in this code.

protolambda:
@trianglesphere I created issues for the follow ups:
- #424
- #425
- #426

PR #397, opened 2022-05-02
specs: define max L2 tx size policy

tynes:
**Description**
Defines a policy for max L2 transaction size. This is important to prevent liveliness failures where users submit
transactions that are too large to be included in a batch.


specs/rollup-node.md
```diff
@@ -146,6 +146,45 @@ Batch contents:
...
+to submit transactions to L2 that make an L1 transaction including it
+be invalid. The only way to do so would be to send a transaction to L2
+that is so large that it would use more than the L1 block's gas limit
+as part of [intrinsic gas][intrinsic-gas]. As of [London][london],
+a zero byte of calldata costs 4 gas while a non zero byte of calldata
+costs 16 gas and the [block gas limit is 30 million][block-size].
+An entire block's worth of gas could be used by a transaction with
+7.5 megabytes of zero byte calldata. Including a transaction
+that uses more than the entire block gas limit as part of it's
+intrinsic gas would cause a liveness failure of L2.
```
tynes, line 160: Previously there was a concept of "force inclusion period" where if a deposit was not ingested in some time, it could be forcefully ingested by a user. This functionality was never really built out, but it would enable a fault proof in the case of a liveliness failure like mentioned here. What is the security model around a sequencer including a L2 tx that is too large to be included in a L1 block?

trianglesphere:
Hmm, this is a good point that has much further reaching implications that we need to figure out.
Currently, all transactions for a single block are submitted in one batch, and multiple batches can be submitted at once. As such, the total size of the transactions in a L2 block needs to be limited.

If there is a transactions that is too large, the problem is not a malicious sequencer, it is a sequencer that keeps the too large transaction around in the mempool. What will happen is that the batch submission will fail, and then the sequencer will re-org. If the sequencer tries to keep including it, it will keep reorging and DOS the transactions that it includes

specs/rollup-node.md
```diff
@@ -146,6 +146,45 @@ Batch contents:
...
+need a direct communication channel with the block
+producer to ensure that the batch can be included in a block
+because it will not be sent via p2p. This max size
+is defined to prevent denial of service attacks on the p2p
+network.
+
+As of [London][london], the policy is set to 128kb.
+The sequencer should not accept transactions larger
+than 126kb to give a 2kb safety buffer. This may
+change pending adoption of [EIP-4844][eip-4844].
```
norswap, line 180: Nit, but can we make the width of that text consistent with the rest of the document?

specs/rollup-node.md
```diff
@@ -146,6 +146,45 @@ Batch contents:
...
+costs 16 gas and the [block gas limit is 30 million][block-size].
+An entire block's worth of gas could be used by a transaction with
+7.5 megabytes of zero byte calldata. Including a transaction
+that uses more than the entire block gas limit as part of it's
+intrinsic gas would cause a liveness failure of L2.
+
+```python
+block_gas_limit = 30_000_000
+max_tx_size = block_gas_limit / 4 # 7500000.0
+```
```
norswap, line 165: Would related this to the paragraph above in some way. As presented, I thought it was a prescription on the max tx size rather than the calculation backing the previous paragraph's point.

norswap:
Agree with @trianglesphere.

One easy thing to do is to define a max size and enforcing it as part of the validation process.

But that doesn't solve all the problems — a batch could still exceed max size by including a lot of transactions within the size limit.

We need to either:
- Make the sequencer size aware so that it never builds blocks bigger than the size limit. Seems like that could expose some DOS opportunity though.
- Enable the batch submitter to split a batch (i.e. all tx for a L2 block) between multiple L1 tx. Probably this is the solution we need.

specs/rollup-node.md
tynes: Sure, sorry about that, will make sure the style follows the rest of the doc

specs/rollup-node.md
tynes: I'm going to delete this code snippet, its not particularly useful

tynes:
> Enable the batch submitter to split a batch (i.e. all tx for a L2 block) between multiple L1 tx. Probably this is the solution we need.

I agree that this is the solution that we need

trianglesphere:
@tynes we're planning to implement splitting L2 transactions/blocks over multiple batches/L1 txns (not sure of the exact names yet). Are you good to close this PR?

tynes:
> @tynes we're planning to implement splitting L2 transactions/blocks over multiple batches/L1 txns (not sure of the exact names yet). Are you good to close this PR?

Yes I'm ok with closing this

PR #400, opened 2022-05-03
contracts: don't unalias on L2->L1

tynes:
Co-authored-by: smartcontracts <kelvin@optimism.io>

**Description**
Don't alias when doing a L2 to L1 message

cc @smartcontracts


norswap:
The rationale being that this is not really necessary because we can't control the sender on L1 like we can on L2, right?

tynes:
> 

To get the message sender on a L2 to L1 transaction, you need to call the cross domain messenger. We don't need to alias on the way out, `msg.sender` will be the messenger/withdrawer depending on which path you take while withdrawing.

PR #401, opened 2022-05-03
specs: Add deposit gas metering spec

trianglesphere:
**Description**
Add specs for the new deposit fee model.

Implementation is done in the following PRs:
- Tracking guaranteed gas on L2: https://github.com/ethereum-optimism/reference-optimistic-geth/pull/21
- Versioning deposits: https://github.com/ethereum-optimism/reference-optimistic-geth/pull/19
- Solidity Fee Market: https://github.com/ethereum-optimism/optimism/pull/2575
- Adjust L1 Info Tx Gas: https://github.com/ethereum-optimism/optimism/pull/2582

**Metadata**
- Fixes ENG-2109


specs/deposit_gas_metering.md
```diff
@@ -0,0 +1,150 @@
...
+    curr_number := block.number
+    curr_bought_gas := 0
+    pack_and_store(curr_basefee, curr_number, curr_bought_gas)
+
+}
+
+curr_bought_gas += required_gas
+require(curr_bought_gas <= GUARANTEED_GAS_LIMIT)
+require(curr_bought_gas <= SANITY_GAS_LIMIT)
+gas_cost = requested_gas * curr_basefee
```
protolambda, line 109: This looks different than the original pseudocode I wrote. The pack_and_store should be at the end, and always run. Otherwise the bought_gas never increases between the deposits in the same block.

trianglesphere: Nice catch. I did update it try to figure out wtf eip 1559 is doing with the basefee calc.

specs/deposit_gas_metering.md
```diff
@@ -100,14 +100,15 @@ if curr_num != block.number {
...
     pack_and_store(curr_basefee, curr_number, curr_bought_gas)
-
 }
 
 curr_bought_gas += required_gas
 require(curr_bought_gas <= GUARANTEED_GAS_LIMIT)
 require(curr_bought_gas <= SANITY_GAS_LIMIT)
 gas_cost = requested_gas * curr_basefee
 
+store(curr_bought_gas) # May have to be a pack and store as well.
```
protolambda, line 110: if you don't return in the above if statement, then the store and pack thing runs twice. Just move that line to the end and it would work. We don't want to run a single `store` here, because we're packing the data.

trianglesphere: Yea, but what about the revert?
TBH I don't really consider the pseudocode that important - particularly with the details of how to deal with storing the data.
The basefee calculations are much more important to specify exactly (including how to handle integer divisions)

protolambda: If we revert the deposit then we don't consume actual gas anyway. The effect (storing the update, and the log) should be at the very end of the logic.

protolambda: Also, I liked the python style of the old pseudocode I wrote better. It makes it easier to compare against the EIP-1559 spec: https://github.com/ethereum/EIPs/blob/master/EIPS/eip-1559.md

trianglesphere: That's a difference in taste - I think the python specs are not actually good as specs.

I'm also looking for some to explain why if `parent_gas_used > parent_gas_target` the basefee_delta cannot be 0 while on the flipside of that condition, it can be.
Another fun fact is that go-ethereum clamps the new basefee at zero (i.e. won't let it go below that, but I can't find a reference to that in the spec).

protolambda: I think in the original pseudocode I tried to not have the 0 case at all by adding an if check to both arms. But yes, definitely don't want to change the basefee if the gas matches target exactly

trianglesphere: Different thing. Your code is exactly the same structure as 1559 w/ renames

```
if used == target:
    basefee_delta = 0
elif used > target:
    basefee_delta = .....
    basefee_delta = max(basefee_delta, 1) # Only in this arm
elif used < target:
   basefee_delta = ....
```

protolambda: Yep, because we want EIP-1559. We should link to the EIP in our spec. And I like less verbose names

protolambda:
For reference. My pseudocode proposal previously posted in notion:
```python
BASE_FEE_MAX_CHANGE_DENOMINATOR = 8

prev_basefee, prev_num, prev_bought_gas = load_and_unpack_storage()

gas_target_limit, gas_sanity_limit = load_and_unpack_storage2()


gas_cost = requested_gas * prev_basefee

now_basefee = prev_basefee
now_num = block.number
now_bought_gas = prev_bought_gas + requested_gas
# update only if we are in a new block
if now_num != prev_num:
  now_bought_gas = requested_gas
  # update basefee if we are using more or less than the target
  if prev_bought_gas < gas_target:
      gas_used_delta = prev_bought_gas - gas_target
			base_fee_per_gas_delta = max(prev_basefee * gas_used_delta // gas_target // BASE_FEE_MAX_CHANGE_DENOMINATOR, 1)
			now_basefee = prev_basefee + base_fee_per_gas_delta
  elif prev_bought_gas > gas_target:
      gas_used_delta = gas_target - prev_bought_gas
			base_fee_per_gas_delta = prev_basefee * gas_used_delta // gas_target // BASE_FEE_MAX_CHANGE_DENOMINATOR
			now_basefee = prev_basefee - base_fee_per_gas_delta

  # TODO: we can reduce now_basefee if now_num > prev_num by multiple blocks,
  #   since that means we had some rest on L2
  
  # optional: maybe reduce gas cost a little bit for spending gas on above math work

require(now_bought_gas < gas_sanity_limit)  # limit how much can be bought per L1 block

required_lockup = mint + (requested_gas * now_basefee)
require(msg.value >= required_lockup)

pack_and_store(now_basefee, now_num, now_bought_gas)
```

specs/deposit_gas_metering.md
tynes: I have a preference for python style code in specs too, it feels like its less characters and easier to read

tynes: What is the sanity limit used for here? Its read from the state but unused?

specs/deposit_gas_metering.md
```diff
@@ -0,0 +1,153 @@
...
+- `gasPrice`
+- `additionalGas`
+
+`guaranteedGas` and `additionalGas` must fit into a 64 bit unsigned integer.  
+`gasPrice` must fit into a 256 bit unsigned integer.
+
+### Guaranteed Gas
+
+The **guaranteed gas** is composed of a gas stipend and any additional guaranteed
+gas the user would like to buy on L1.
```
norswap, line 34: Since the term "stipend" has caused confusion in the past, I would precise here that the gas stipend is L2 gas given for free.

specs/deposit_gas_metering.md
```diff
@@ -0,0 +1,153 @@
...
+bought. To reduce PGAs if the deposit mechanism is congested, we also implement
+an EIP-1559 style fee market with the following pseudo code:
+
+```text
+BASE_FEE_MAX_CHANGE_DENOMINATOR = 8
+ELASTICITY_MULTIPLIER = 2
+
+curr_basefee: u128, curr_num: u64, curr_bought_gas: u64 = load_and_unpack_storage()
+GUARANTEED_GAS_LIMIT: u64, SANITY_GAS_LIMIT: u64 = load_and_unpack_storage2()
+gas_target = GUARANTEED_GAS_LIMIT // ELASTICITY_MULTIPLIER
```
norswap, line 83: What are those `load_and_unpack_storage()` for? (Feels like related to ABI?) Does it really need to be part of the spec (and if so, can it be explained better?)

norswap: What is the `SANITY_GAS_LIMIT`?

specs/deposit_gas_metering.md
```diff
@@ -0,0 +1,153 @@
...
+    }
+    curr_basefee := new_basefee
+    curr_number := block.number
+    curr_bought_gas := 0
+   
+}
+
+curr_bought_gas += required_gas
+require(curr_bought_gas <= GUARANTEED_GAS_LIMIT)
+require(curr_bought_gas <= SANITY_GAS_LIMIT)
```
norswap, line 109: `require(curr_bought_gas <= min(GUARANTEED_GAS_LIMIT, SANITY_GAS_LIMIT)`

specs/deposit_gas_metering.md
norswap: Must also explain what `curr_bought_gas` is (i.e. requested guaranteed gas on top of stipend).

specs/deposit_gas_metering.md
```diff
@@ -0,0 +1,153 @@
...
+
+### Rationale for Guaranteed vs Additional Gas
+
+Need to have this additional gas mechanism to enable gas refunds.  
+Guaranteed gas cannot be refunded without opening up a DOS vector.
+
+### Rationale for burning L1 Gas
+
+If we burn Eth (or collect it), we need to add the payable selector everywhere.
+Adding it everywhere is not feasible and really bad UX.
```
norswap, line 125: nit, but we should standardize on "ETH" everywhere, not "eth" or "Eth"

norswap: Also, the current plan is to have both, meaning we need to duplicate code paths between a `payable` version and one that is not `payable`. So in a certain sense it's feasible.

The problem is that we don't want to require contracts that are making deposit to define all their functions as payable themselves!

specs/deposit_gas_metering.md
```diff
@@ -0,0 +1,153 @@
...
+
+When processing L2 blocks, do the following:
+
+1. Sum up all guaranteed gas in deposits.
+2. Subtract the guaranteed gas from the gas pool.
+3. Don't later subtract guaranteed gas when processing deposits individually.
+
+The reason for subtracting out guaranteed gas at the start is to ensure that we have
+enough gas for all deposits in the presence of additional gas.
+
```
norswap, line 139: Super confused by this. This sounds like we have a limit `additional_gas <= gas_pool_size - guaranteed_gas`.

This doesn't really make sense to me, because say that the total request additional gas is more than that:

1. What actually happens? Is the additional gas not purchased? Then we need to order the transactions by `gasPrice` before processing them.
     - What happens in the case the `gasPrice` in deposits are much lower than the gas prices in normal L2 tx, do we still prioritze deposits?
2. Why don't we let additional gas use 100% of `gas_pool_size`? In the current scheme it looks like we would cut it off at `gas_pool_size - guaranteed_gas` BUT THEN let normal L2 tx use up to `guaranteed_gas`.

specs/deposit_gas_metering.md
```diff
@@ -0,0 +1,153 @@
...
+### Additional Gas On L2
+
+Additional gas is only bought given the following conditions:
+
+- `gasPrice` >= the L2 gas price (however it is set)
+- There is enough `additionalGas` remaining in the gas pool
```
norswap, line 145: What does this mean? Isn't it just enough gas remaining in the gas pool, since additional gas is just normal gas?

specs/deposit_gas_metering.md
```diff
@@ -0,0 +1,153 @@
...
+3. Don't later subtract guaranteed gas when processing deposits individually.
+
+The reason for subtracting out guaranteed gas at the start is to ensure that we have
+enough gas for all deposits in the presence of additional gas.
+
+### Additional Gas On L2
+
+Additional gas is only bought given the following conditions:
+
+- `gasPrice` >= the L2 gas price (however it is set)
```
norswap, line 144: Do we need to split this into max_basefee + tip? cf. comment above on ordering by `gasPrice` 

The benefit of a max basefee is that you can specify a high one if you want to get speedy inclusion (and not overthink your parameters) and usually you won't get charged the whole amount — if there happens to be high congestion, you will simply drop out due to low tip.

The other benefit is that this is how normal gas works, and additional gas is supposed to be normal gas. This might potentially be easier to implement also (??)

specs/deposit_gas_metering.md
```diff
@@ -0,0 +1,153 @@
...
+- There is enough `additionalGas` remaining in the gas pool
+- The account has enough balance to cover `gasPrice * additionalGas`
```
norswap, line 146: So cf. comment above, this has the interesting property that suddenly additional gas works differently from normal gas which only pays current block basefee + tip.

specs/deposit_gas_metering.md
```diff
@@ -0,0 +1,153 @@
...
+
+If `additionalGas` is bought, the `gasLimit` of the transaction is
+`guaranteedGas + additionalGas` else it is `guaranteedGas`
```
norswap, line 149: Missing period at the end.

specs/deposit_gas_metering.md
```diff
@@ -0,0 +1,153 @@
...
+
+When the transaction is done, if `gasUsed > guaranteedGas` a refund of
+`(gasLimit - (gasUsed - guaranteedGas)) * gasPrice` is sent to
+`from`. It is important to not refund any `guaranteedGas`
```
norswap, line 153: This will refund `(gasLimit - gasUsed + guaranteedGas) * gasPrice` which is not what we want (you're refunding unused gas and guaranteed gas!).

The refund should be `gasLimit - max(gasUsed, guaranteedGas)` (and no conditions needed anymore). This ensures we don't get a refund if used less gas the guaranteed gas.

norswap (changes requested):
I think fleshing out how deposits are being processed on L2 wrt fees needs a bit of thought — feels like there a few big issues there currently.

specs/deposit_gas_metering.md
trianglesphere: It still needs to pay for it. In a pre 1559 you're right that users could set `gasPrice` to 0. Post 1559 on L2 the `gasPrice` must be greater than the `base fee`

specs/deposit_gas_metering.md
trianglesphere: I don't think so. Sequencers have to include deposits no matter what.

specs/deposit_gas_metering.md
trianglesphere: In practice the additional gas works like a legacy transaction

specs/deposit_gas_metering.md
norswap: I meant it could have been "there is enough gas remaining in the gas pool" instead of "enough additionalGas remaining in the gas pool".

specs/deposit_gas_metering.md
norswap: We discussed it on the call, gasPrice seems simpler, esp. this will not be used to order transactions (though there is a proposal to use something else for that, see below).

specs/deposit_gas_metering.md
```diff
@@ -0,0 +1,153 @@
...
+  - [Rationale for burning L1 Gas](#rationale-for-burning-l1-gas)
+- [Deposits on L2](#deposits-on-l2)
+  - [Guaranteed Gas On L2](#guaranteed-gas-on-l2)
+  - [Additional Gas On L2](#additional-gas-on-l2)
+
+<!-- END doctoc generated TOC please keep comment here to allow auto update -->
+
+## Deposits on L1
+
+L1 Deposits are augmented with the following fields:
```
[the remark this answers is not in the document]
trianglesphere: No longer augmenting this

specs/deposit_gas_metering.md
```diff
@@ -0,0 +1,153 @@
...
+
+If a gas stipend is provided, we only buy the amount of guaranteed gas
+in excess of the gas stipend.
+
+### Additional Gas
+
+Users can request additional gas for their L2 transaction. This gas is
+refundable is less than the gas limit is used. However, if there is not
+enough available gas on L2 or the gas price not high enough, the additional
+gas will not be provided on L2.
```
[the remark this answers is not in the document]
trianglesphere: cut

specs/deposit_gas_metering.md
```diff
@@ -0,0 +1,153 @@
...
+
+Additional gas is bought with the following:
+
+- `gasPrice`: L2 gas price to buy.
```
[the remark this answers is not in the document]
trianglesphere: cut

specs/deposit_gas_metering.md
```diff
@@ -0,0 +1,153 @@
...
+- `additionalGas`: Amount of additional gas requested
+
+The additional gas will be bought with the `from` account's balance. We do
+not require that the user mint enough L2 eth to cover the additional gas fee,
+but do suggest that tooling checks that the user has enough balance for it.
```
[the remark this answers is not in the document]
trianglesphere: cut

specs/deposit_gas_metering.md
trianglesphere: Fixing both. Switching to ETH

specs/deposit_gas_metering.md
trianglesphere: I guess the language here is weird. It's also no longer relevant.
The intention was to ensure that we always had enough gas for guaranteed gas, and then as we processed each deposit, we may or may not have enough additional gas.

specs/deposit_gas_metering.md
trianglesphere: this got cut

specs/guaranteed-gas-market.md
```diff
@@ -0,0 +1,143 @@
...
+
+Because there is some cost to submitting the transaction and updating the basefee,
+we provide transactions with a small amount of free gas.
+
+If the user requests more `guaranteedGas` than the `gasStipend`, that gas will
+be bought with L1 ETH via a gas burn or by buying it directly. If they request
+less gas than the stipend, they will not be charged.
+
+We provide XX,XXX amount of gas in the stipend. TODO - should this be price
+rather than gas based??
```
norswap, line 37: To consider:
- we want to guarantee that this amount is always enough to register the tx on L2 for replayability purposes
- but at the same time, this means the guaranteed gas could be "cheaper" than guaranteed gas or even regular L2 gas, meaning it could be optimal to make a lot of deposits
     - however, taking advantage of this requires some kind of L2 setup where a long computation can be split over many smaller one, which is not usually how things work

Given that, I think gas-based is the way to go.

specs/guaranteed-gas-market.md
```diff
@@ -0,0 +1,143 @@
...
+  
+  # optional: maybe reduce gas cost a little bit for spending gas on above math work
+
+require(now_bought_gas < gas_sanity_limit)  # limit how much can be bought per L1 block
+
+required_lockup = mint + (requested_gas * now_basefee)
+require(msg.value >= required_lockup)
+
+pack_and_store(now_basefee, now_num, now_bought_gas)
+```
```
norswap, line 131: Didn't re-review this, but is it normal there are two copies of the pseudocode? 

Same comment as before, not really sure what `pack_and_store` is.

specs/guaranteed-gas-market.md
```diff
@@ -0,0 +1,143 @@
...
+      now_basefee = prev_basefee + base_fee_per_gas_delta
+  elif prev_bought_gas > gas_target:
+      gas_used_delta = gas_target - prev_bought_gas
+      base_fee_per_gas_delta = prev_basefee * gas_used_delta // gas_target // BASE_FEE_MAX_CHANGE_DENOMINATOR
+      now_basefee = prev_basefee - base_fee_per_gas_delta
+
+  # TODO: we can reduce now_basefee if now_num > prev_num by multiple blocks,
+  #   since that means we had some rest on L2
+  
+  # optional: maybe reduce gas cost a little bit for spending gas on above math work
```
norswap, line 123: That work is factored in the guaranteed gas you get "for free".

specs/guaranteed-gas-market.md
trianglesphere: Yea, I lean to gas based because it makes more sense. The problem is that it mucks up the gas market. Somewhere in the 50k-100k gas range feels right.

specs/guaranteed-gas-market.md
trianglesphere: We'll decide on one copy before merging. `pack_and_store` is about storing multiple values in a single storage slot (as part of solidity).

norswap: What's the difference between both copies?

specs/guaranteed-gas-market.md
```diff
@@ -0,0 +1,143 @@
...
+
+require(now_bought_gas < gas_sanity_limit)  # limit how much can be bought per L1 block
```
tynes, line 125: There is a difference between the 2 specs on this line here

specs/guaranteed-gas-market.md
```diff
@@ -0,0 +1,143 @@
...
+    }
+    curr_basefee := new_basefee
+    curr_number := block.number
+    curr_bought_gas := 0
+   
+}
+
+curr_bought_gas += required_gas
+require(curr_bought_gas <= min(GUARANTEED_GAS_LIMIT, SANITY_GAS_LIMIT)
+gas_cost = requested_gas * curr_basefee
```
tynes, line 87: `gas_cost` is missing `mint`, is this required?

specs/guaranteed-gas-market.md
```diff
@@ -0,0 +1,143 @@
...
+        new_basefee   := curr_basefee - basefee_delta
+    }
+    curr_basefee := new_basefee
+    curr_number := block.number
+    curr_bought_gas := 0
+   
+}
+
+curr_bought_gas += required_gas
+require(curr_bought_gas <= min(GUARANTEED_GAS_LIMIT, SANITY_GAS_LIMIT)
```
tynes, line 86: missing trailing `)`

specs/guaranteed-gas-market.md
trianglesphere: I don't think I understand the question. If you mean `gas_cost` + `mint` >= `msg.value`, note that we set `mint` to be `msg.value`.

specs/guaranteed-gas-market.md
```diff
@@ -0,0 +1,143 @@
...
+
+now_basefee = prev_basefee
+now_num = block.number
+now_bought_gas = prev_bought_gas + requested_gas
+# update only if we are in a new block
+if now_num != prev_num:
+  now_bought_gas = requested_gas
+  # update basefee if we are using more or less than the target
+  if prev_bought_gas < gas_target:
+      gas_used_delta = prev_bought_gas - gas_target
```
tynes, line 112: There is an underflow here here, if `prev_bought_gas` is smaller then it must be on the right hand side of the subtraction operator

specs/guaranteed-gas-market.md
```diff
@@ -0,0 +1,143 @@
...
+BASE_FEE_MAX_CHANGE_DENOMINATOR = 8
+ELASTICITY_MULTIPLIER = 2
+
+curr_basefee: u128, curr_num: u64, curr_bought_gas: u64 = load_and_unpack_storage()
+GUARANTEED_GAS_LIMIT: u64, SANITY_GAS_LIMIT: u64 = load_and_unpack_storage2()
+gas_target = GUARANTEED_GAS_LIMIT // ELASTICITY_MULTIPLIER
+
+# // implies floor division, however because gas_delta is always positive, it is the same as truncating (aka round to 0) division
+# If first deposit of this block, calculate the new basefee and store other info as well.
+if curr_num != block.number {
```
tynes, line 65: I don't know if i follow this because if we do the updates of the base fee inside of this `if curr_num != block.number` then that means only the first deposit of each block can influence the deposit basefee

trianglesphere: That is correct - the fee is updated once per block at the start of the block. I just realized we may need to adjust this algorithm to handle skipped blocks.

We are intentionally moving away from the design where the fee increases for each tx in the block.

specs/guaranteed-gas-market.md
tynes: There are some slight differences in the logic, need to reconcile which is the desired behavior. I have a preference for the python, as I find it easier to read

specs/guaranteed-gas-market.md
```diff
@@ -0,0 +1,134 @@
...
+
+### Exponent Based Fee Reduction
+
+When there are stretches where no deposits are executed on L1, the basefee should be decaying, but
+is not. If there is the case that the basefee spiked, this mechanism is needed to enable a more
+accurate decay. It uses exponentiation to run in constant (relative to the number of missed blocks)
+gas.
+
+With the current elasticty and change denominator values, if `n` is greater than 42, the
+exponention will overflow. TODO: Use exponent/logarithm library for this.
```
protolambda, line 121: If `n > 42` then there have been no deposits for a long period of time `42 * 12 = 504 seconds = 8.4 minutes`. At that point we may as well reset the base fee for deposits to a default minimum value. No need for special libraries that way.

specs/deposits.md
```diff
@@ -50,9 +50,11 @@ transaction types:
 1. They are derived from Layer 1 blocks, and must be included as part of the protocol.
 2. They do not include signature validation (see [User-Deposited Transactions][user-deposited]
    for the rationale).
+3. They buy their L2 gas on L1 and as such, the L2 gas is not refundable.
 
-We define a new [EIP-2718] compatible transaction type with the prefix `0x7E`, and the following
-fields (rlp encoded in the order they appear here):
+We define a new [EIP-2718] compatible transaction type with the prefix `0x7E`, and then a versioned
+byte sequence. The first version has `0x00` as the version byte and then ahs the following fields
```
protolambda, line 56:
```suggestion
byte sequence. The first version has `0x00` as the version byte and then as the following fields
```
typo?

trianglesphere: yep. TY

specs/guaranteed-gas-market.md
trianglesphere: `(7/8)^42 ~1e-3` which is a large reduction in basefee, but it doesn't feel large enough to reset it to the minimum. It is something that I thought about.

trianglesphere: Definitely an option. This isn't fully decided yet

specs/guaranteed-gas-market.md
```diff
@@ -0,0 +1,143 @@
...
+cost to the user. If the ETH price of the gas (gas times current L1 baseefee) is greater than the
+requested guaranteed gas times the L2 gas price, no L1 gas is burnt.
+
+## Limiting Guaranteed Gas
+
+The total amount of guaranteed gas that can be bought in a single L1 block must be limited to
+prevent a denial of service attack against L2 as well as allow the total amount of guaranteed gas
+to be below the L2 block gas limit.
+
+We set a guaranteed gas limit of 2,500,000 gas per L1 block. This corresponds to the L2 gas target
```
trianglesphere, line 46: TODO: Update @trianglesphere

specs/guaranteed-gas-market.md
```diff
@@ -0,0 +1,143 @@
...
+To reduce [Priority Gas Auctions](./glossary.md#priority-gas-auction) and accurately price gas, we
+implement a 1559 style fee market on L1 with the following pseudocode. We also use this opporunity
+to place a hard limit on the amount of guaranteed gas that is provided.
+
+```python
+# Pseudocode to update the L2 Deposit Basefee and cap the amount of guaranteed gas
+# bought in a block. Calling code must handle the gas burn and validity checks on
+# the ability of the account to afford this gas.
+BASE_FEE_MAX_CHANGE_DENOMINATOR = 8
+ELASTICITY_MULTIPLIER = 2
```
trianglesphere: TODO: Update @trianglesphere

specs/deposits.md
```diff
@@ -144,7 +144,9 @@ follows:
 - `context.value` set to `sendValue`
 
 No gas is bought on L2 and no refund is provided. The gas used for the deposit is subtracted from
-the gas pool on L2.
+the gas pool on L2. Gas usage exactly matches other transaction types (including intrinsic gas).
```
maurelian, line 147: is 'gas pool' a commonly used term? I think 'block gas limit' is more intuitive.

trianglesphere:
Closing in favor of https://github.com/ethereum-optimism/optimism/pull/2679
I've moved over relevant comments.

PR #404, opened 2022-05-03
contracts: fix build

tynes:
**Description**
Fixes a build issue


trianglesphere:
I've confirmed that this works with the following steps
- Clean install
- Git submodule init and update
- forge update `packages/contracts/lib/forge-std` (internal forge dependency)

trianglesphere (approved):
I'm still seeing a local diff (probably around which solc I'm using over the forge version). All the contract builds are working through (tested off a fresh install).

PR #406, opened 2022-05-03
fix: ci build bindings

tynes:
**Description**
Build the bindings in CI and fail if there is a diff.

This helps to prevent the bindings from getting out of date as well as solve the problem of different outputs of abigen being used between different members of the team. This uses the latest geth code to build abigen


.github/workflows/go-ci.yml
```diff
@@ -30,3 +30,49 @@ jobs:
...
+      - uses: actions/checkout@v2
+        with:
+          submodules: recursive
+
+      - name: Install Go
+        uses: actions/setup-go@v2
+        with:
+          go-version: 1.18
+
+      - name: Install abigen
```
trianglesphere, line 47: try `go get` or `go install` here?

protolambda (approved):
LGTM. Maybe squash the commits before merge

PR #412, opened 2022-05-04
New Deposit Event

trianglesphere:
**Description**
This is the contracts and rollup node work to support the new fields for deposit gas metering.
This adds the new fields to the contracts and uses a new version of L2 geth that has the fields
in the deposit transaction (but has not implemented the new metering).

**Testing**

Fuzzed `FuzzUnmarshallLogEvent` for about 3 hours locally without any founds bugs.
It's running around 8,000/sec and found ~200 interesting cases.

**Metadata**
- Fixes  ENG-2188


trianglesphere:
Closing because we have moved away from the additional gas mechanism for the deposit fee version.

PR #416, opened 2022-05-05
Fix L2 Output Timestamps

trianglesphere:
**Description**
There was an off by one error in the L2 Output Oracle contract in the timestamp to
block number conversion. In addition, the L2 Output Submitter (op proposer) did
not recognize that there was a mismatch between the timestamp/block number in
the header and the timestamp/block number from the contract.

This bug causes problems when trying to do withdrawals.

**Metadata**
- Fixes ENG-2128


l2os/drivers/l2output/driver.go
```diff
@@ -206,6 +206,15 @@ func (d *Driver) CraftTx(
 		return nil, fmt.Errorf("error resolving checkpoint block: %v", err)
 	}
 
+	l2Header, err := d.cfg.L2Client.HeaderByNumber(ctx, nextCheckpointBlock)
+	if err != nil {
+		return nil, fmt.Errorf("error resolving checkpoint block: %v", err)
+	}
```
protolambda, line 212: Why not change the `outputRootAtBlock` RPC in the rollup-node to serve the L2 block header. That way it's always consistent. The l2 output root submitter shouldn't even need to directly talk to a L2 geth node I think

trianglesphere: That wouldn't solve the original issue nor a race condition on the proof. Original is that the `timestamp -> block number` conversion was wrong. The race condition is that we get by block number but don't check the proof against the header - the only way to do this is by checking the state root.

protolambda: FWIW, the server of the output root already does get a header and verifies the proof against the state root: https://github.com/ethereum-optimism/optimistic-specs/blob/5189b9a26831f0a3e51587e3ddb9aba65195fdaf/opnode/node/api.go#L73
It would be nice to serve the header in the same RPC response, and then the timestamp / block number can be used without extra calls with a separate RPC client to L2 geth.

packages/integration-tests/test/000_withdrawals.spec.ts
```diff
@@ -141,7 +141,7 @@ describe('Withdrawals', () => {
       )
     })
 
-    it('should verify the withdrawal on L1', async function () {
```
tynes, line 144: I don't think we should merge this until it is passing

trianglesphere:
This is ready for a re-review. The test is failing because the sequencer is reorging on the L1 Info transaction between when it sequences it and when it verifies it

opnode/test/system_test.go
```diff
@@ -155,6 +155,15 @@ func TestL2OutputSubmitter(t *testing.T) {
 	initialSroTimestamp, err := l2OutputOracle.LatestBlockTimestamp(&bind.CallOpts{})
 	require.Nil(t, err)
 
+	// Wait until the second output submission from L2. The output submitter submits outputs from the
+	// unsafe portion of the chain which gets reorged on startup. The sequencer has an out of date view
+	// when it creates it's first block and uses and old L1 Origin. It then does not submit a batch
+	// for that block and subsequently reorgs to match what the verifier derives when running the
+	// reconcillation process.
```
tynes, line 162: I'd really like to know why this is happening

trianglesphere: I'm assuming that it starts block production prior to having a full up to date view of the L1 chain.

PR #419, opened 2022-05-06
StateViz: Visualize rollup state changes

Inphi:
Log rollup state changes so they can be later inspected using `stateviz`. `stateviz` is a webserver that serves HTML visualizing rollup state snapshots. Each rollup driver engine can be viewed side by side on a table where each row represents a rollup state change event.

`stateviz` is added to the devnet deployment and accessible at http://localhost:9090 upon running `make devnet-up`. The webpage can be manually refreshed every couple seconds to get the latest snapshots logged by the opnode.

image

**Metadata**
* Fixes ENG-2156


opnode/cmd/stateviz/assets/index.html
```diff
@@ -0,0 +1,27 @@
...
+
+</head>
+
+<body>
+    <div class="container-fluid">
+        <div id="logs" class="row">
+        </div>
+    </div>
+
+    <script src="https://code.jquery.com/jquery-3.6.0.min.js"></script>
```
tynes, line 18: Would it be easy to add the integrity tag for the other scripts as well?

opnode/cmd/stateviz/main.go
```diff
@@ -0,0 +1,248 @@
...
+
+func makeGzipHandler(fn http.HandlerFunc) http.HandlerFunc {
+	return func(w http.ResponseWriter, r *http.Request) {
+		if !strings.Contains(r.Header.Get("Accept-Encoding"), "gzip") {
+			fn(w, r)
+			return
+		}
+		w.Header().Set("Content-Encoding", "gzip")
+		gz := gzip.NewWriter(w)
+		defer gz.Close()
```
tynes, line 192: Does `close` get called too early here?

mslipper: I don't think so - it gets called when the handler function returns.

opnode/rollup/driver/driver.go
```diff
@@ -66,15 +66,15 @@ type Network interface {
...
+func NewDriver(cfg rollup.Config, l2 *l2.Source, l1 *l1.Source, network Network, log log.Logger, snapshotLog log.Logger, sequencer bool) *Driver {
 	output := &outputImpl{
 		Config: cfg,
 		dl:     l1,
 		l2:     l2,
 		log:    log,
 	}
 	return &Driver{
-		s: NewState(log, cfg, l1, l2, output, network, sequencer),
+		s: NewState(log, cfg, l1, l2, output, network, snapshotLog sequencer),
```
trianglesphere, line 77:
```suggestion
		s: NewState(log, cfg, l1, l2, output, network, snapshotLog, sequencer),
```

opnode/service.go
```diff
@@ -80,3 +80,19 @@ func NewLogConfig(ctx *cli.Context) (node.LogConfig, error) {
 	}
 	return cfg, nil
 }
+
+func NewSnapshotLogger(ctx *cli.Context) (log.Logger, error) {
```
trianglesphere, line 84: missing/wrong import? It should be pulling from go-ethereum's logger.

opnode/rollup/driver/state.go
```diff
@@ -469,3 +476,19 @@ func (s *state) loop() {
 		}
 	}
 }
+
+func (s *state) snapshot(event string) {
+	l1HeadJSON, _ := json.Marshal(s.l1Head)
```
trianglesphere, line 481: How do you feel about making a copy and then marshaling / sending in a different thread?

opnode/service.go
```diff
@@ -80,3 +80,19 @@ func NewLogConfig(ctx *cli.Context) (node.LogConfig, error) {
...
+	snapshotFile := ctx.GlobalString(flags.SnapshotLog.Name)
+	handler := log.StreamHandler(os.Stdout, log.JSONFormat())
```
trianglesphere, line 86: Can we not default to stdout and just no-op the logger if we don't get a file?

trianglesphere (commented):
I left some scattered comments prior to this review.
One thing I'd like here is a good way to turn off this feature if it is not configured.
The other is that there are some random go issues that need to be fixed before CI is good.

opnode/rollup/driver/state.go
mslipper: I'd rather not complicate things for now. If marshal performance becomes an issue, then we can process these asynchronously.

opnode/service.go
mslipper: Done.

trianglesphere (commented):
Integration test failing at the build stage:
```
Run make devnet-up
[4](https://github.com/ethereum-optimism/optimistic-specs/runs/6389257496?check_suite_focus=true#step:6:4)
Regenerating L1 genesis.
[5](https://github.com/ethereum-optimism/optimistic-specs/runs/6389257496?check_suite_focus=true#step:6:5)
Bringing up L1...
[6](https://github.com/ethereum-optimism/optimistic-specs/runs/6389257496?check_suite_focus=true#step:6:6)
yaml.parser.ParserError: while parsing a block collection
[7](https://github.com/ethereum-optimism/optimistic-specs/runs/6389257496?check_suite_focus=true#step:6:7)
  in "./docker-compose.yml", line 57, column 7
[8](https://github.com/ethereum-optimism/optimistic-specs/runs/6389257496?check_suite_focus=true#step:6:8)
expected <block end>, but found '<block sequence start>'
[9](https://github.com/ethereum-optimism/optimistic-specs/runs/6389257496?check_suite_focus=true#step:6:9)
  in "./docker-compose.yml", line 58, column 9
[10](https://github.com/ethereum-optimism/optimistic-specs/runs/6389257496?check_suite_focus=true#step:6:10)
make: *** [Makefile:38: devnet-up] Error 1
```

There are itest/docker changes, so if it's related they need to be fixed. If they're not related, I'm happy with merging this.

trianglesphere (approved):
lgtm

Issue #398, opened 2022-05-02
Withdrawer `receive` function


tynes:
**Is your feature request related to a problem? Please describe.**
I'd like to be able to initiate a withdrawal on L2 as easily as possible.

**Describe the solution you'd like**
The `withdrawer` contract should have a `receive` function that you can send value directly to and have it initialize a withdrawal. This would simplify tooling to withdraw from L2 as it could just send eth to a contract to have it be withdrawn. This function could wrap existing functionality and just expose a nice api. This functionality exists on L1 for the deposits feed, so it makes sense to have it work both ways.

PR #421, opened 2022-05-06
refactor: new messengers

smartcontracts:
Implements the new messenger flow. This ensures that the messengers and bridges work in a backwards compatible way. The previous iteration of the messengers didn't take into account deposits that were in flight during the upgrade. This pull request has some major architectural changes so that the bridges now go through the messengers again. This allows for replayability in the bridges as well as gives the bridges replay protection.

A main feature of this PR is that the messenger and bridges now extend a base messenger or bridge contract to make the functionality basically the same on both L1 and L2. The previous iteration of the contracts made a distinction between the L1 and L2 sides of things. The previous APIs are persisted, but now there is a universal API that can be used on the bridges/messengers no matter which domain you are in.

Hardhat deploy scripts are added to the `ops` setup.

The bindings for the `l2os` and `opnode` are regenerated.

Deprecates:
- https://github.com/ethereum-optimism/optimistic-specs/pull/411
- https://github.com/ethereum-optimism/optimistic-specs/pull/396

Unblocks:
- https://github.com/ethereum-optimism/optimistic-specs/pull/423

Closes:
- https://github.com/ethereum-optimism/optimistic-specs/issues/398


packages/contracts/contracts/L2/L2StandardBridge.sol
```diff
@@ -0,0 +1,186 @@
...
+     */
+    function _initiateWithdrawal(
+        address _l2Token,
+        address _from,
+        address _to,
+        uint256 _amount,
+        uint32 _minGasLimit,
+        bytes calldata _data
+    ) internal {
+        address l1Token = OptimismMintableERC20(_l2Token).l1Token();
```
norswap, line 178: Could use `remoteToken()` to avoid legacy interface.

packages/contracts/contracts/L2/L2ToL1MessagePasser.sol
```diff
@@ -62,21 +72,20 @@ contract Withdrawer {
...
         address _target,
         uint256 _gasLimit,
-        bytes calldata _data
-    ) external payable {
-        address from = msg.sender;
-
-        bytes32 withdrawalHash = WithdrawalVerifier._deriveWithdrawalHash(
+        bytes memory _data
+    ) public payable {
+        bytes32 withdrawalHash = WithdrawalVerifier.withdrawalHash(
```
tynes, line 77: Add todo: handle upgradablility

packages/contracts/contracts/L1/OptimismPortal.sol
```diff
@@ -35,6 +119,132 @@ contract OptimismPortal is DepositFeed, WithdrawalsRelay {
...
+        // Prevent reentrency
+        require(_target != address(this), "Cannot send message to self.");
+
+        // Get the output root.
+        L2OutputOracle.OutputProposal memory proposal = L2_ORACLE.getL2Output(_l2Timestamp);
+
+        // Ensure that enough time has passed since the proposal was submitted
+        // before allowing a withdrawal. A fault proof should be submitted
+        // before this check is allowed to pass.
+        require(
```
tynes, line 188: Revert the `<`

packages/contracts/contracts/L1/OptimismPortal.sol
```diff
@@ -1,32 +1,116 @@
...
+
+    /**********
+     * Events *
+     **********/
+
+    /**
+     * @notice Emitted when a Transaction is deposited from L1 to L2. The parameters of this
+     * event are read by the rollup node and used to derive deposit transactions on L2.
+     */
+    event TransactionDeposited(
```
tynes, line 51: Double check that this event is correct

packages/contracts/contracts/L1/L2OutputOracle.sol
```diff
@@ -1,25 +1,33 @@
...
     /**********
      * Events *
      **********/
 
     /// @notice Emitted when an output is appended.
-    event l2OutputAppended(bytes32 indexed _l2Output, uint256 indexed _l2timestamp);
+    event l2OutputAppended(
+        bytes32 indexed _l2Output,
+        uint256 indexed _l1Timestamp,
+        uint256 indexed _l2timestamp
```
trianglesphere, line 22: Should also use block number here. This is a larger refactor though.

packages/contracts/contracts/L1/OptimismPortal.sol
```diff
@@ -35,6 +119,132 @@ contract OptimismPortal is DepositFeed, WithdrawalsRelay {
...
+     * @param _l2Timestamp L2 timestamp of the outputRoot.
+     * @param _outputRootProof Inclusion proof of the withdrawer contracts storage root.
+     * @param _withdrawalProof Inclusion proof for the given withdrawal in the withdrawer contract.
+     */
+    function finalizeWithdrawalTransaction(
+        uint256 _nonce,
+        address _sender,
+        address _target,
+        uint256 _value,
+        uint256 _gasLimit,
```
trianglesphere, line 173: Is this new code? This is the gas limit that should be limited to 64 bites

packages/contracts/contracts/L1/OptimismPortal.sol
trianglesphere: very important

packages/contracts/contracts/L2/L2ToL1MessagePasser.sol
```diff
@@ -62,21 +72,20 @@ contract Withdrawer {
     function initiateWithdrawal(
         address _target,
         uint256 _gasLimit,
```
trianglesphere, line 74: too large gas limit.

packages/contracts/contracts/libraries/Lib_CrossDomainHashing.sol
```diff
@@ -0,0 +1,249 @@
...
+     */
+    function L2TransactionHash(
+        bytes32 _l1BlockHash,
+        uint256 _logIndex,
+        address _from,
+        address _to,
+        bool _isCreate,
+        uint256 _mint,
+        uint256 _value,
+        uint256 _gas,
```
trianglesphere, line 43: too large? Maybe not actually?

packages/contracts/contracts/libraries/Lib_CrossDomainHashing.sol
```diff
@@ -0,0 +1,249 @@
...
+     */
+    function L2Transaction(
+        bytes32 _l1BlockHash,
+        uint256 _logIndex,
+        address _from,
+        address _to,
+        bool _isCreate,
+        uint256 _mint,
+        uint256 _value,
+        uint256 _gas,
```
trianglesphere, line 87: 256 bit gas

packages/contracts/contracts/libraries/Lib_CrossDomainHashing.sol
```diff
@@ -0,0 +1,249 @@
...
+    /**
+     * @notice Encodes the cross domain message based on the version that
+     * is encoded in the nonce
+     */
+    function getVersionedEncoding(
+        uint256 _nonce,
+        address _sender,
+        address _target,
+        uint256 _value,
+        uint256 _gasLimit,
```
trianglesphere, line 155: ditto

packages/contracts/contracts/libraries/Lib_CrossDomainHashing.sol
```diff
@@ -0,0 +1,249 @@
...
+
+    /**
+     * @notice Compute the cross domain hash based on the versioned nonce
+     */
+    function getVersionedHash(
+        uint256 _nonce,
+        address _sender,
+        address _target,
+        uint256 _value,
+        uint256 _gasLimit,
```
trianglesphere, line 176: ditto

packages/contracts/contracts/libraries/Lib_CrossDomainHashing.sol
```diff
@@ -0,0 +1,249 @@
...
+
+    /**
+     * @notice Compute the V1 cross domain serialization
+     */
+    function getEncodingV1(
+        uint256 _nonce,
+        address _sender,
+        address _target,
+        uint256 _value,
+        uint256 _gasLimit,
```
trianglesphere, line 209: ditto

packages/contracts/contracts/libraries/Lib_CrossDomainHashing.sol
```diff
@@ -0,0 +1,249 @@
...
+
+    /**
+     * @notice Compute the V1 hash of a cross domain message
+     */
+    function getHashV1(
+        uint256 _nonce,
+        address _sender,
+        address _target,
+        uint256 _value,
+        uint256 _gasLimit,
```
trianglesphere, line 244: ditto

packages/contracts/contracts/libraries/Lib_WithdrawalVerifier.sol
```diff
@@ -28,13 +31,13 @@ library WithdrawalVerifier {
      * @param _gasLimit Gas to be forwarded to the target.
      * @param _data Data to send to the target.
      */
-    function _deriveWithdrawalHash(
+    function withdrawalHash(
         uint256 _nonce,
         address _sender,
         address _target,
         uint256 _value,
         uint256 _gasLimit,
```
trianglesphere, line 39: 256 bit gas

packages/contracts/contracts/test/CommonTest.t.sol
```diff
@@ -3,14 +3,306 @@ pragma solidity 0.8.10;
...
+    L1CrossDomainMessenger L1Messenger;
+    L2CrossDomainMessenger L2Messenger;
+    L2ToL1MessagePasser messagePasser;
+
+    event SentMessage(
+        address indexed target,
+        address sender,
+        bytes message,
+        uint256 messageNonce,
+        uint256 gasLimit
```
trianglesphere, line 92: ditto

packages/contracts/contracts/test/CommonTest.t.sol
```diff
@@ -3,14 +3,306 @@ pragma solidity 0.8.10;
...
+    );
+
+    event WithdrawalInitiated(
+        uint256 indexed nonce,
+        address indexed sender,
+        address indexed target,
+        uint256 value,
+        uint256 gasLimit,
```
trianglesphere, line 100: ditto

packages/contracts/contracts/universal/CrossDomainMessenger.sol
```diff
@@ -0,0 +1,244 @@
...
+            _sender,
+            _target,
+            _value,
+            _minGasLimit,
+            _message
+        );
+
+        if (_isSystemMessageSender()) {
+            // Should never happen.
+            require(msg.value == _value, "Mismatched message value.");
```
trianglesphere, line 172: I have concerns about this when we are doing stuff on L2 with this. Specifically the L1 value is set to the L2 mint, and then a user specified (to the deposit tx at least) amount is set to the value field. If these fields do not line up exactly, we have have stuck messages. In particular, we expect one version of the Optimism Portal Deposit function to take eth to pay for L2 gas which would change the amount of value emitted.

trianglesphere (approved):
I just flagged everywhere I was a gasLimit that was 256 bits, but didn't do as much in terms of determining if it was valid or not.

I also flagged where I had the concern around msg.value and _value on L2. I don't think this is fully relevant yet, but will be in the future.

Otherwise, this looks good to me.

packages/contracts/contracts/universal/StandardBridge.sol
```diff
@@ -0,0 +1,306 @@
...
+     */
+    function _isOptimismMintable(address _localToken, address _remoteToken)
+        internal
+        view
+        returns (bool)
+    {
+        return ((ERC165Checker.supportsInterface(_localToken, 0x1d1d8b63) &&
+            _remoteToken == OptimismMintableERC20(_localToken).l1Token()) ||
+            (ERC165Checker.supportsInterface(_localToken, 0x0bc32271) &&
+                _remoteToken == OptimismMintableERC20(_localToken).remoteToken()));
```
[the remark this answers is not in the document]
tynes: The call to `l1Token` here is required to ensure that existing cross domain tokens still work

packages/contracts/contracts/L1/L1CrossDomainMessenger.sol
```diff
@@ -0,0 +1,67 @@
...
+     * Internal Functions *
+     **********************/
+
+    /**
+     * @notice Ensure that the L1CrossDomainMessenger can only be called
+     * by the OptimismPortal and the L2 sender is the L2CrossDomainMessenger.
+     */
+    function _isSystemMessageSender() internal view override returns (bool) {
+        return msg.sender == address(portal) && portal.l2Sender() == otherMessenger;
+    }
```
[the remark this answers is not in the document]
tynes: The `OptimismPortal` does make calls on L1 for a withdrawal, the flow through the L2 messenger would be encoding a call to the L1 messenger `relayMessage`

packages/contracts/contracts/L2/L2StandardBridge.sol
smartcontracts: Needs to be l1Token() or will break legacy contracts

packages/contracts/contracts/L1/L2OutputOracle.sol
tynes: We need consensus before making that large of a change, should put together a doc @trianglesphere  and schedule discussion for next office hours @K-Ho

packages/contracts/contracts/L1/OptimismPortal.sol
tynes: This is new code, good call

packages/contracts/contracts/L1/L1CrossDomainMessenger.sol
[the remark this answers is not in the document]
norswap: Gotcha, I get it now 👍

packages/contracts/contracts/L2/L2ToL1MessagePasser.sol
tynes: This is safe, downstream there is a `require(gasleft() > _gasLimit + buffer)` 

I lean towards not changing any of the gas limits here as the only one that needs to be a uint64 is the one used for deposits. Unless there is strong preference for it to be a uint64 everywhere just for sanity reasons, which I would understand too

trianglesphere: I think that it could be fine. I just flagged each instance without too much thought. It's just that any gas limit that serializes to go / actual execution needs to fit in 64 bits (and probably way fewer) otherwise the deposit gets stuck

tynes: Thoughts @smartcontracts?

packages/contracts/contracts/L1/OptimismPortal.sol
tynes: This is fixed in https://github.com/ethereum-optimism/optimistic-specs/pull/429

PR #423, opened 2022-05-09
opnode: Withdrawal E2E test in go

trianglesphere:
**Description**

This does the following:
- Adds withdrawal utilities (to opnode/withdrawals)
- Adds an end to end test in go of withdrawals
- Adds the L2 withdrawer contract
- Updates to a newer version of reference-go-ethereum

**Metadata**
- Fixes ENG-2202


opnode/contracts/Makefile
```diff
@@ -59,3 +82,19 @@ bindings-optimism-portal:
 		--pkg deposit \
 		--type OptimismPortal \
 		--out ./deposit/deposit_feed_raw.go
+
+
```
tynes, line 86: Why exactly is this commented out?

tynes:
Some of this stuff is going to change pending https://github.com/ethereum-optimism/optimistic-specs/pull/421 due to it being updated to support backwards compatible withdrawals

opnode/contracts/Makefile
trianglesphere: Because the L2 Output Oracle Contract is built in the `l2os` package. I can remove these fully.

opnode/contracts/Makefile
```diff
@@ -40,11 +49,29 @@ bindings-l1-block-info:
 		--type L1Block \
 		--out ./l1block/l1_block_info_raw.go
 
+bindings-L2ToL1MessagePasser:
```
tynes, line 52: There are some mismatched names in files here, `withdrawer` -> `L2ToL1MessagePasser`

opnode/withdrawals/utils.go
```diff
@@ -0,0 +1,267 @@
...
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/core/types"
+	"github.com/ethereum/go-ethereum/crypto"
+	"github.com/ethereum/go-ethereum/ethclient"
+	"github.com/ethereum/go-ethereum/ethclient/gethclient"
+	"github.com/ethereum/go-ethereum/rlp"
+	"github.com/ethereum/go-ethereum/rpc"
+)
+
+var L2WithdrawalAddr = common.HexToAddress("0x4200000000000000000000000000000000000016")
```
tynes, line 24: This is no longer correct, the predeploy is at `address(42...0)`

tynes:
This could use a rebase

opnode/contracts/Makefile
trianglesphere: TBH I'm ok with the go names being out of date, I've got an issue to update them (ENG-2195)

opnode/withdrawals/utils.go
trianglesphere: No zeroes?

trianglesphere: Also, this has been removed and I'm using the address in the `predeploys` package. FYI this address is still found in the typesecript helpers

trianglesphere:
> This could use a rebase

Will do once #416 is merged

opnode/withdrawals/utils.go
trianglesphere: This is working in the tests because we init the code at `predeploy.WithdrawalsAddress` and then reference it at that address later.

opnode/test/system_test.go
```diff
@@ -641,3 +645,196 @@ func TestL1InfoContract(t *testing.T) {
...
+	header, err = l1Client.HeaderByNumber(ctx, receipt.BlockNumber)
+	require.Nil(t, err)
+
+	ctx, cancel = context.WithTimeout(context.Background(), 1*time.Second)
+	defer cancel()
+	endBalance, err = l1Client.BalanceAt(ctx, fromAddr, nil)
+	require.Nil(t, err)
+
+	// Ensure that withdrawal - gas fees are added to the L1 balance
+	// Fun fact, the fee is greater than the withdrawal amount
```
protolambda, line 835: Fun fact :sweat_smile: 
We're withdrawing only 500 gwei though, so it makes sense

opnode/withdrawals/proof.go
```diff
@@ -0,0 +1,100 @@
...
+
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/core/types"
+	"github.com/ethereum/go-ethereum/crypto"
+	"github.com/ethereum/go-ethereum/ethclient/gethclient"
+	"github.com/ethereum/go-ethereum/rlp"
+	"github.com/ethereum/go-ethereum/trie"
+)
+
+type proofDB struct {
```
protolambda, line 16: For the output root construction we've some proof code already. But without the storage proof part, just the account proof. There it types the rlp nodes as `hexutil.Bytes` and maintains a map of `common.Hash` to nodes, instead of using strings for keys. Here it nicely uses the account type + some of the geth rpc client types though. In both cases it does verify correctly, and does verify all the things in the response (i.e. the account details in the response are checked against the rlp node in the proof). Either way, some opportunity for a later PR to take the best of both and reduce code.

opnode/withdrawals/utils.go
```diff
@@ -0,0 +1,266 @@
...
+
+// WaitForFinalizationPeriod waits until the timestamp has been submitted to the L2 Output Oracle on L1 and
+// then waits for the finalization period to be up.
+// This functions polls and can block for a very long time if used on mainnet.
+// This returns the timestamp to use for the proof generation.
+func WaitForFinalizationPeriod(ctx context.Context, client *ethclient.Client, portalAddr common.Address, timestamp uint64) (uint64, error) {
+	opts := &bind.CallOpts{Context: ctx}
+	timestampBig := new(big.Int).SetUint64(timestamp)
+
+	portal, err := deposit.NewOptimismPortalCaller(portalAddr, client)
```
protolambda, line 33: Same applies to `deposit` as to `withdrawer` rename I think? Renames in a later PR are fine

opnode/test/system_test.go
```diff
@@ -641,3 +645,196 @@ func TestL1InfoContract(t *testing.T) {
...
+
+	opts.Value = nil
+	tx, err = portal.FinalizeWithdrawalTransaction(
+		opts,
+		params.Nonce,
+		params.Sender,
+		params.Target,
+		params.Value,
+		params.GasLimit,
+		params.Data,
```
protolambda, line 811: Might want to test a `params.Data` that's not empty, but integration tests can pick that up too.

opnode/test/system_test.go
```diff
@@ -641,3 +645,196 @@ func TestL1InfoContract(t *testing.T) {
...
+	if x.Cmp(gasFeeCap) > 0 {
+		x = gasFeeCap
+	}
+	return x.Mul(x, new(big.Int).SetUint64(gasUsed))
+}
+
+// TestWithdrawals checks that a deposit and then withdrawal execution succeeds. It verifies the
+// balance changes on L1 and L2 and has to include gas fees in the balance checks.
+// It does not check that the withdrawal can be executed prior to the end of the finality period.
+func TestWithdrawals(t *testing.T) {
```
protolambda, line 662: Nice test, long function though. With the monorepo change the e2e tests get their own go module (so we can separate any testing dependencies, and pull in all the other modules), maybe after that we can split the `system_test.go` in themed files (general, deposits, withdrawals, p2p, etc.) or maybe even packages.

protolambda (approved):
LGTM! Some things we may want to do in future PRs, nothing blocking though.

tynes (approved):

opnode/test/system_test.go
trianglesphere: Yea, there's a lot of boilerplate that's hard to avoid.

opnode/test/system_test.go
trianglesphere: Good call - that'd be part of doing generic contract calls (rather than simple transfer) via the withdrawal mechanism

opnode/test/system_test.go
trianglesphere: When it's expressed like that, it makes sense. I always forget how may decimals there actually are.

PR #428, opened 2022-05-11
contracts: add backup logic for deposits

smartcontracts:
**Description**
Introduces backup logic for deposits that prevents users from
accidentally making bad deposits. Deposits that don't complete
successfully will be returned back to the layer where the deposits were
made.


tynes:
You will need to run `forge snapshot` and then commit in the diff to the `.gas-snapshot`

packages/contracts/contracts/universal/StandardBridge.sol
```diff
@@ -208,14 +223,45 @@ abstract contract StandardBridge {
         uint256 _amount,
         bytes calldata _data
     ) public onlyOtherBridge {
```
tynes, line 225: Previously, not all revert cases were being caught. With the `try/catch`, it guarantees that all revert cases are caught. The downside is that you can only `try/catch` on public functions, so it needs to be restricted with an `onlySelf` modifier and have the additional check that `localToken` isn't calling itself. I think that this is overall more robust and will guarantee any reverting conditions do not result in users getting their tokens/ETH stuck

packages/contracts/contracts/universal/StandardBridge.sol
```diff
@@ -289,24 +368,19 @@ abstract contract StandardBridge {
         emit ERC20BridgeInitiated(_localToken, _remoteToken, _from, _to, _amount, _data);
     }
```
tynes, line 370: Do you mind bringing back the devdoc?

tynes (approved):

PR #430, opened 2022-05-12
Add state surgery package

mslipper:
Copies OVM ETH into state. See README.md for more info on how this works.

Fixes:

- ENG-2138


state-surgery/migrate.go
```diff
@@ -0,0 +1,269 @@
...
+	totalOVM := new(big.Int)
+
+	// Iterate over each account in the state.
+	it := trie.NewIterator(tr.NodeIterator(nil))
+	for it.Next() {
+		// It's up to use to decode trie data.
+		var data types.StateAccount
+		if err := rlp.DecodeBytes(it.Value, &data); err != nil {
+			panic(err)
+		}
```
protolambda, line 139: I believe some accounts may have a balance in the OVM eth contract, but are otherwise completely unused (zero nonce etc.) and don't show up in the account trie. So they would be missed during this iteration, and not get migrated over their balance, losing funds.

mslipper: Got it. The application does a check to make sure that the migrated balance matches the total supply of OVM ETH, so if any of these exist I should be able to catch them. If these addresses do exist I'll need to do some archaeology to dig up the actual addresses since we can't easily map between OVM ETH storage slots and addresses.

mslipper: I'm running this with the check active against mainnet right now; should have an answer in about an hour or so.

maurelian:
This should be closed and moved to the monorepo if it hasn't already right?

trianglesphere:
Closing in favor of https://github.com/ethereum-optimism/optimism/pull/2655

PR #431, opened 2022-05-12
Add OVM_L1BlockNumber predeploy

mslipper:
Meta:

- Fixes ENG-2139


tynes:
I have a preference for leaving the `L2` folder flat, in its current form the other contracts that are in there are also predeploys but not in `L2/predeploys`

packages/contracts/contracts/L2/predeploys/OVM_L1BlockNumber.sol
```diff
@@ -0,0 +1,17 @@
+// SPDX-License-Identifier: MIT
+pragma solidity 0.8.10;
+
+import { iOVM_L1BlockNumber } from "./iOVM_L1BlockNumber.sol";
+import { L1Block } from "../L1Block.sol";
+
```
tynes, line 6: Technically the current implementation of this contract will return the L1 block number with just a call to the contract, you don't need to abide by the interface.
I think its unlikely anybody is using the contract in this way, but we may want to have a fallback function that does the same thing as `getL1BlockNumber`, see https://github.com/ethereum-optimism/optimism/blob/2e0b633004b14edd14d9c01fd5119b58680a7dfa/packages/contracts/tasks/take-dump.ts#L109

packages/contracts/contracts/L2/predeploys/OVM_L1BlockNumber.sol
```diff
@@ -0,0 +1,17 @@
...
+contract OVM_L1BlockNumber is iOVM_L1BlockNumber {
+    address public l1BlockAddress;
```
tynes, line 8: Perhaps make this internal? Will be exposing more public API that will become maintenance burden over time

packages/contracts/contracts/L2/predeploys/OVM_L1BlockNumber.sol
```diff
@@ -0,0 +1,17 @@
...
+
+    constructor(address _l1BlockAddress) {
```
tynes, line 10: You should get rid of the constructor because that will require execution at deploy time, we want to be able to easily set the bytecode into the state when building the genesis file. Replace the constructor with an `initialize` function that `require`s the `l1BlockAddress == address(0)` then sets it so that it can be only called once

packages/contracts/contracts/L2/predeploys/iOVM_L1BlockNumber.sol
```diff
@@ -0,0 +1,13 @@
+// SPDX-License-Identifier: MIT
+pragma solidity 0.8.10;
+
+/**
+ * @title iOVM_L1BlockNumber
+ */
+interface iOVM_L1BlockNumber {
```
tynes, line 7: @smartcontracts Do we want to standardize on not having interfaces? You mentioned that, I do know that its convenient to import them only when the solidity pragma is loose, but it is very easy for devs to generate their own interfaces with `cast interface <address verified on etherscan or path to artifact>`

packages/contracts/contracts/test/OVM_L1BlockNumber.t.sol
```diff
@@ -0,0 +1,25 @@
+//SPDX-License-Identifier: MIT
+pragma solidity 0.8.10;
+
+import { DSTest } from "forge-std/Test.sol";
+import { Vm } from "forge-std/Vm.sol";
+import { L1Block } from "../L2/L1Block.sol";
+import { OVM_L1BlockNumber } from "../L2/predeploys/OVM_L1BlockNumber.sol";
+
```
tynes, line 8: If you do this:

```solidity
import { Test } from "forge-std/Test.sol";
```

And extend using `Test` instead of `DSTest`, then you don't need to worry about `Vm`. `DSTest` is the most legacy thing made by makerdao, `Test` extends `DSTest` with some nice to haves

packages/contracts/contracts/L2/predeploys/OVM_L1BlockNumber.sol
mslipper: How would that fallback function work? `fallback` can only return `bytes memory`, which is a different API from the one currently in place.

tynes: We would have to use some assembly to do some type coercion. I think its super unlikely that anybody depends on that API but if we want to be pedantic then we may have to support it. The other problem is now that its written in solidity, it'll use a lot more gas than previously - there will be a jumptable for functions, checks on `msg.value == 0` that solidity does automatically. Unlikely that anybody is relying on that extremely low cost of the raw bytecode, but still mentioning another difference. It will be impossible to get around some increases gas costs since we are going from a simple "push a value from the evm context to the stack" to a "call another contract and load the return value into memory" model.

Its extremely unlikely anybody is using the fallback based API

PR #433, opened 2022-05-12
opnode: p2p RPC, fix static-peers

protolambda:
Implement:
- peerstore dump RPC method (with filter for only getting connected peers)
- RPC methods to protect/ban/dial etc. if we ever need to test or intervene a running node
- Fix & test static-peering
- Add logging for connections
- Go client bindings for P2P RPC namespace (`opp2p`)
- Refactor p2p to bundle the p2p components into a p2p node, to not bloat the rollup-node main struct as much, and better test the p2p stack in isolation

### DIFF
--- a/opnode/flags/p2p_flags.go
+++ b/opnode/flags/p2p_flags.go
@@ -147,7 +147,7 @@ var (
 		Usage:    "User-agent string to share via LibP2P identify. If empty it defaults to 'optimism-VERSIONHERE'.",
 		Hidden:   true,
 		Required: false,
-		Value:    "",
+		Value:    "optimism",
 		EnvVar:   p2pEnv("AGENT"),
 	}
 	TimeoutNegotiation = cli.DurationFlag{
--- a/opnode/node/server.go
+++ b/opnode/node/server.go
@@ -7,6 +7,8 @@ import (
 	"net"
 	"net/http"
 
+	"github.com/ethereum-optimism/optimistic-specs/opnode/p2p"
+
 	"github.com/ethereum-optimism/optimistic-specs/opnode/l2"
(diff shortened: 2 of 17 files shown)


opnode/node/server.go
```diff
@@ -32,23 +34,32 @@ func newRPCServer(ctx context.Context, rpcCfg *RPCConfig, rollupCfg *rollup.Conf
...
 }
 
-func (s *rpcServer) Start() error {
-	apis := []rpc.API{{
-		Namespace:     "optimism",
-		Service:       s.api,
-		Public:        true,
+func (s *rpcServer) EnableP2P(backend *p2p.APIBackend) {
+	s.apis = append(s.apis, rpc.API{
+		Namespace:     "",
```
tynes, line 52: I feel like using the `optimism` namespace should be fine here

protolambda: Keeping the p2p server in a "opp2p" namespace for now, since it may not always be active (if p2p is disabled the RPC backend isn't started)

mslipper (approved):

opnode/flags/p2p_flags.go
```diff
@@ -147,7 +147,7 @@ var (
 		Usage:    "User-agent string to share via LibP2P identify. If empty it defaults to 'optimism-VERSIONHERE'.",
```
trianglesphere, line 147: Leave todo to add version later? Could also cut `-VERSIONHERE`

trianglesphere (approved):
One tiny nit, otherwise LGTM

opnode/flags/p2p_flags.go
protolambda: removed the `-VERSIONHERE` part. Think we might want to use the new `runtime/debug` package to grab a version, or continue the current makefile + var override thing.

PR #436, opened 2022-05-12
specs: deposits pay gas on L1

tynes:
**Description**
I didn't notice anywhere in the specs where it is clearly mentioned that deposits are not charged for gas on L2. Correct me if I am wrong, but we are not going with the additional gas scheme so now the only way that deposits are metered is based on the L1 eip1159 curve in the deposit contract


trianglesphere:
The addition to the spec is correct.
There is also a very similar addition in https://github.com/ethereum-optimism/optimistic-specs/pull/401

norswap (approved):

mslipper (approved):

trianglesphere:
Closing in favor of #401

PR #439, opened 2022-05-13
P2P discovery

protolambda:
Implements:
- ENR entry (chain ID and version)
- Background process to do random DHT walk, filter nodes, and add to the peerstore
- Background process to select random, not connected, peers and connect to them
- Discovery test: `A` is the bootnode, `B` and `C` find each other via `A`

### DIFF
--- a/op-node/p2p/discovery.go
+++ b/op-node/p2p/discovery.go
@@ -1,14 +1,30 @@
 package p2p
 
 import (
+	"bytes"
+	"context"
+	secureRand "crypto/rand"
+	"encoding/binary"
+	"fmt"
+	"io"
+	"math/rand"
 	"net"
+	"time"
+
+	"github.com/ethereum-optimism/optimistic-specs/opnode/rollup"
+	"github.com/ethereum/go-ethereum/p2p/enr"
+	"github.com/ethereum/go-ethereum/rlp"
+	"github.com/libp2p/go-libp2p-core/crypto"
+	"github.com/libp2p/go-libp2p-core/network"
+	"github.com/libp2p/go-libp2p-core/peer"
+	ma "github.com/multiformats/go-multiaddr"
 
 	"github.com/ethereum/go-ethereum/log"
 	"github.com/ethereum/go-ethereum/p2p/discover"
 	"github.com/ethereum/go-ethereum/p2p/enode"
 )
 
-func (conf *Config) Discovery(log log.Logger) (*enode.LocalNode, *discover.UDPv5, error) {
+func (conf *Config) Discovery(log log.Logger, rollupCfg *rollup.Config) (*enode.LocalNode, *discover.UDPv5, error) {
 	if conf.NoDiscovery {
 		return nil, nil, nil
 	}
@@ -19,6 +35,11 @@ func (conf *Config) Discovery(log log.Logger) (*enode.LocalNode, *discover.UDPv5
 	if conf.AdvertiseUDPPort != 0 {
 		localNode.SetFallbackUDP(int(conf.AdvertiseUDPPort))
 	}
+	dat := OptimismENRData{
+		chainID: rollupCfg.L2ChainID.Uint64(),
+		version: 0,
+	}
+	localNode.Set(&dat)
 
 	udpAddr := &net.UDPAddr{
 		IP:   conf.ListenIP,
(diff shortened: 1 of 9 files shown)


specs/rollup-node-p2p.md
```diff
@@ -86,10 +86,10 @@ The Ethereum Node Record (ENR) for an Optimism rollup node must contain the foll
 - A UDP port (`udp` field) representing the local discv5 listening port.
 - An Optimism (`optimism` field) L2 network identifier
 
-The `optimism` value is encoded as the concatenation of:
+The `optimism` value is encoded as a single RLP `bytes` value, the concatenation of:
 
-- chain ID (`varint`)
-- fork ID (`varint`)
+- chain ID (`unsigned varint`)
```
tynes, line 91: Why are unsigned values used here?

protolambda: Because the chain ID and version are never negative.

op-node/p2p/node.go
```diff
@@ -4,6 +4,9 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"strconv"
+
+	ma "github.com/multiformats/go-multiaddr"
```
trianglesphere, line 9: Given we use this import only once, it seems little overkill to shorten it

op-node/p2p/discovery.go
```diff
@@ -43,8 +82,295 @@ func (conf *Config) Discovery(log log.Logger) (*enode.LocalNode, *discover.UDPv5
...
+	filter := FilterEnodes(log, cfg)
+	// We pull nodes from discv5 DHT in random order to find new peers.
+	// Eventually we'll find a peer record that matches our filter.
+	randomNodeIter := n.dv5Udp.RandomNodes()
+
+	randomNodeIter = enode.Filter(randomNodeIter, filter)
+	defer randomNodeIter.Close()
+
+	// We pull from the DHT in a slow/fast interval, depending on the need to find more peers
+	discoverTicker := time.NewTicker(time.Second * 5)
```
trianglesphere, line 200: Should be constants to start (IDK if we want to parameterize slow vs faster later one)

op-node/p2p/discovery.go
```diff
@@ -43,8 +82,295 @@ func (conf *Config) Discovery(log log.Logger) (*enode.LocalNode, *discover.UDPv5
...
+			if err != nil {
+				log.Debug("failed connection attempt", "peer", id, "err", err)
+			}
+		}
+	}
+
+	// stops all the workers when we are done
+	defer close(connAttempts)
+	// start workers to try connect to peers
+	for i := 0; i < 4; i++ {
```
trianglesphere, line 240: More magic numbers

op-node/p2p/discovery.go
```diff
@@ -43,8 +82,295 @@ func (conf *Config) Discovery(log log.Logger) (*enode.LocalNode, *discover.UDPv5
...
+		go connectWorker(ctx)
+	}
+
+	// buffer 10 nodes
+	randomNodesCh := make(chan *enode.Node, 3)
```
trianglesphere, line 245: Size + magic number don't match up

op-node/p2p/discovery.go
```diff
@@ -43,8 +82,295 @@ func (conf *Config) Discovery(log log.Logger) (*enode.LocalNode, *discover.UDPv5
...
+				return
+			}
+		}
+	}
+	// Walk the DHT in parallel, the discv5 interface does not use channels for the iteration
+	go bufferNodes()
+
+	// Kick off by trying the nodes we have in our table (previous nodes from last run and/or bootnodes)
+	go func() {
+		<-time.After(time.Second * 3)
```
trianglesphere, line 273: magic number

op-node/p2p/discovery.go
```diff
@@ -43,8 +82,295 @@ func (conf *Config) Discovery(log log.Logger) (*enode.LocalNode, *discover.UDPv5
...
+			}
+			// We add the addresses to the peerstore, and update the address TTL to 24 hours.
+			//After that we stop using the address, assuming it may not be valid anymore (until we rediscover the node)
+			pstore.AddAddrs(info.ID, info.Addrs, time.Hour*24)
+			_ = pstore.AddPubKey(info.ID, (*crypto.Secp256k1PublicKey)(found.Pubkey()))
+			// Tag the peer, we'd rather have the connection manager prune away old peers,
+			// or peers on different chains, or anyone we have not seen via discovery.
+			// There is no tag score decay yet, so just set it to 42.
+			n.ConnectionManager().TagPeer(info.ID, fmt.Sprintf("optimism-%d-%d", dat.chainID, dat.version), 42)
+			log.Debug("discovered peer", "peer", info.ID, "nodeID", found.ID(), "addr", info.Addrs[0])
```
trianglesphere, line 312: can `len(info.Addrs)` ever equal zero?

op-node/p2p/discovery.go
```diff
@@ -43,8 +82,295 @@ func (conf *Config) Discovery(log log.Logger) (*enode.LocalNode, *discover.UDPv5
...
+			if uint(len(connected)) < connectGoal {
+				// Start looking for more peers more actively again
+				faster()
+
+				peersWithAddrs := n.Host().Peerstore().PeersWithAddrs()
+				if err := shufflePeers(peersWithAddrs); err != nil {
+					continue
+				}
+
+				existing := make(map[peer.ID]struct{})
```
trianglesphere, line 328: `map[T]bool` also works nicely as a set

trianglesphere (commented):
There's a couple more magic numbers in the discovery function than I'd like.
I do see how the API is a pain to work with. `go test ./p2p -race` doesn't show anything which is good. The `DiscoveryProcess` is probably worth coming back to to clean up a bit more - there's a lot of moving parts + internal communication. It's long running, so this is slightly less of a concern - but I think the shutdown process is a little hairy.

trianglesphere:
Concurrency diagram

image

op-node/p2p/discovery.go
protolambda: `map[T]bool` is fine, but generally I like `struct{}` better. Takes minimal memory, and iterating a set cannot go wrong, unlike `existing[key] = false` that adds an element to the set where `if existing[key]` is false but iteration catches it.

op-node/p2p/discovery.go
protolambda: No, `enrToAddrInfo` always returns an addr-info with 1 address derived from the ENR.

protolambda:
- Fixed imports / magic numbers
- Opened an issue for possible discovery improvements: #445

I will prioritize testnet + scaling first, but then plan to revisit peer-scoring, and can then address the discovery process once we've a better scoring / peer-tracking system in place.

trianglesphere (approved):
lgtm

PR #414, opened 2022-05-05
Adopt go.work, rename modules to prep for monorepo

protolambda:
This PR:
- renames the modules to their final destination name in the monorepo
- splits the packages into modules: `op-node`, `op-proposer`, `op-batcher`, `op-e2e`
- moves `ops` to `ops-bedrock`, to prepare for monorepo
- Many Makefile and Dockerfile updates to make the above work
- updates CI (including getting coverage for all the separate modules)

PR #441, opened 2022-05-13
Adopt go.work, rename modules to prep for monorepo

mslipper:
Replaces #414 after following @protolambda's instructions.


Makefile
```diff
@@ -35,39 +43,34 @@ clean:
 .PHONY: clean
 
 devnet-up:
-	@bash ./ops/devnet-up.sh
+	@bash ./ops-bedrock/devnet-up.sh
```
tynes, line 46: If we are going to be putting this on a branch of the monorepo, should this be `ops` still?

mslipper: I think we should separate them for a bit. This lets us incrementally merge things together.

go.work
```diff
@@ -0,0 +1,13 @@
+go 1.18
+
+use (
+	./op-batcher
+	./op-e2e
+	./op-node
+	./op-proposer
+)
+
+replace github.com/ethereum/go-ethereum v1.10.16 => github.com/ethereum-optimism/reference-optimistic-geth v0.0.0-20220512193241-32ef0bba48ce
```
tynes, line 10: I notice that this is replaced on `op-e2e/go.mod` as well - I'm assuming that the replace there will override this replace, I feel like it would make sense to have a single replace that applies to the entire repo to ensure that things are all on the same version of `go-ethereum`

trianglesphere:
Was seeing withdrawals test failing - seems like it was not using an up to date version of our geth client

go.work
trianglesphere: The replacement does two things: set the version, and deal with the fact that our reference geth declares itself to be a different module from how it is imported as. As such, I think you might need the replace, but I'm not fully sure.

smartcontracts:
Maybe worth renaming the branch and the PR. For branch, maybe `release/bedrock`.

smartcontracts (commented):
Seems good to me. I don't want to give the explicit approval, but didn't see anything off.

go.work
mslipper: We need both replacements. The one in `go.work` sets the replacement globally for dev purposes. The one in `op-e2e` lets us import the real reference geth package in CI.

trianglesphere (approved):
LGTM
One thing that we'll need to figure out is handling the replace statement across all the modules.

protolambda:
@trianglesphere The way it's supposed to work with `go.work` is that each `go.mod` references the other modules in the workspace by version (and we can version them independently), thus essentially we are able to freeze what each module builds with independently. But since it's not in the `main` branch, and not even in the correct repo yet, we need all those `replace` statements for Go to resolve where the module is coming from. Once the module can be resolved we can get rid of the `replace` statements. The `go.work` will then still bundle them together for cross-module editing, and in dockerfiles we make use of a slimmed down `go.work` with the modules involved in the docker build. This is one of the reasons why I dislike the feature-branch: it will force us to keep those replace statements around for longer, as the modules can't be resolved nicely.

protolambda (approved):
LGTM

protolambda:
And regarding the replace statements:
- The reference geth statements will need to stay; each module needs to be build-able independently,  same as current single module system. But we get to keep the git history separate and import paths the same, big win long term for maintainability and small diff
- The other replace statements, to modules on local path, can eventually be removed once we switch them to proper versioning + resolving via main branch in monorepo, as in my above comment.

PR #443, opened 2022-05-13
For CI: Remove replace statements from go.mod files

trianglesphere:
**Description**
A clear and concise description of the features you're adding in this pull request.


trianglesphere:
Does not work when building modules individually.

PR #449, opened 2022-05-26
Use ticker instead of time.After in select loop

trianglesphere:
**Description**

Repeatedly using time.After in a select loop leaks resources as
resources are allocated for each timer, but not released until the
timer fires. The timer may not fire due to other return paths. By
using tickers, this resource leak is no longer able to happen.

I verified that this problem no longer exists with the following
semgrep command.
semgrep --config="r/dgryski.semgrep-go.timeafter.leaky-time-after"

**Metadata**
- Fixes ENG-2155


trianglesphere:
closing to move to monorepo

PR #450, opened 2022-05-26
Use net.JoinHostPort instead of fmt.Sprintf

trianglesphere:
**Description**
fmt.Sprintf is discouraged from being used to create network addresses.

**Metadata**
- Fixes #[Link to Issue]


trianglesphere:
Closing to move to monorepo.

PR #453, opened 2022-06-02
contracts: Update withdrawer references to L2ToL1MessagePasser

maurelian:
**Description:**

This is a fix to a bullet point in OZ-N-05:
- The Lib_WithdrawalVerifier contract refers to an obsolete withdrawer predeploy
contract rather than the L2ToL1MessagePasser contract.

Fixing that error turned into a larger commit which I broke out here. Since these contracts
are not yet live, I think we should do everything we can now to avoid getting stuck with
misleading names

I've made this change in two commits. One that modifies the solidity, another which modifies the makefile and generated bindings in the op-bindings package, because TBH I'm not sure what the consequences of that are. If it's OK to do so, these changes should probably be squashed.


trianglesphere:
FYI If bindings are changed, you need to update usage of them

maurelian:
> FYI If bindings are changed, you need to update usage of them

right of course. Moving to draft.

